PR #16, opened 2018-01-26
[kanister] Add types of artifacts

julio-lopez:
* Add kanister artifact types: CloudObject and Untyped
* Rerun `make codegen`
* Update tutorial to reference typed artifacts
* Update examples blueprints to reference typed artifacts


julio-lopez:
@tdmanv Only the commits with new changes show up, 6b256fe in this case.
The history is not as clean as it could be because of the additional merge commit created when merging into master. But, this will do for now.

PR #29, opened 2018-02-06
Omit helm specific labels (#2285)

tdmanv:
## Overview
Label propagation is not enforced, so we cannot rely on it. This patch handles the specific case where helm charts only propagate their app label as we've seen with stable/mysql.

## Followup
We'll deep to come up with a more robust way of going from Deployment to ReplicaSets. One approach would be to remove label filtering alltogether.

## Test Plan
Unit tests pass and manually tested on the stable/mysql chart. This exposes a gap in our testing infra where we only test with constructed specs. we should be testing with externally defined specs/charts.


julio-lopez:
LG +1 👍

PR #69, opened 2018-04-14
Location CRD, 1.10go, kubetask changes and profiles

depohmel:


ntolia:
Need to push a new PR

PR #96, opened 2018-06-18
Add DeleteData Kanister function

tdmanv:


tdmanv:
resubmitting with unit test fix

PR #137, opened 2018-11-02
Fix kanister-operator helm release (#4328)

SupriyaKasten:
* Fix kanister-operator helm release


SupriyaKasten:
@pavannd1 please try to merge after it gets approved to check if you can.

PR #148, opened 2019-01-03
GCP location and credentials for Kanister,   Prefer the get script to install kanister tools, Bump tool script to 0.16.0,  Skip flaky scale test

SupriyaKasten:


SupriyaKasten:
Resyncing!

PR #202, opened 2019-08-13
Fix the bug where ListSnapshot failing to query deleted fs

### LINKED ISSUE #201: EFS SnapshotsList returns error if originating filesystem is deleted
The problem happens in case:
- A snapshot is taken from filesystem.
- Filesystem is deleted
- SnapshotsList returns an error saying that the filesystem cannot be found.

hakanmemisoglu:
Fixes the problem where EFS ListSnapshots fails to query already deleted filesystem.

* Do not fail when VolumeGet operation in ListSnapshots returns an error.
* Do not fail when volume argument in snapshotFromRecoveryPointByVault is nil.

Fixes #201

### DIFF
--- a/pkg/blockstorage/awsefs/awsefs.go
+++ b/pkg/blockstorage/awsefs/awsefs.go
@@ -453,10 +453,9 @@ func (e *efs) snapshotsFromRecoveryPoints(ctx context.Context, rps []*backup.Rec
 		if err != nil {
 			return nil, errors.Wrap(err, "Failed to get volume ID from recovery point ARN")
 		}
-		vol, err := e.VolumeGet(ctx, volID, "")
-		if err != nil {
-			return nil, errors.Wrap(err, "Failed to get EFS volume")
-		}
+		// VolumeGet might return error since originating filesystem might have
+		// been deleted.
+		vol, _ := e.VolumeGet(ctx, volID, "")
 		snap, err := snapshotFromRecoveryPointByVault(rp, vol, tags, e.region)
 		if err != nil {
 			return nil, errors.Wrap(err, "Failed to get snapshot from the vault")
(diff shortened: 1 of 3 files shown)


pkg/blockstorage/awsefs/awsefs.go
```diff
@@ -453,10 +453,9 @@ func (e *efs) snapshotsFromRecoveryPoints(ctx context.Context, rps []*backup.Rec
 		if err != nil {
 			return nil, errors.Wrap(err, "Failed to get volume ID from recovery point ARN")
 		}
-		vol, err := e.VolumeGet(ctx, volID, "")
-		if err != nil {
-			return nil, errors.Wrap(err, "Failed to get EFS volume")
-		}
+		// VolumeGet might return error since originating filesystem might have
+		// been deleted.
+		vol, _ := e.VolumeGet(ctx, volID, "")
```
tdmanv: is it possible to check the specific NotFound error?

pkg/blockstorage/awsefs/conversion.go
```diff
@@ -90,8 +94,9 @@ func snapshotFromRecoveryPointByVault(rp *backup.RecoveryPointByBackupVault, vol
 	if rp.RecoveryPointArn == nil {
 		return nil, errors.New("Recovery point has no RecoveryPointArn")
 	}
-	if volume == nil {
-		return nil, errors.New("Nil volume as argument")
+	encrypted := false
+	if volume != nil {
+		encrypted = volume.Encrypted
 	}
```
pavannd1: What happens when the volume is `nil` since we are removing this check?

hakanmemisoglu: `blockstorage.Snasphot`'s `Volume` field will be `nil`, `Encrypted` field will be `false`. 
Do you see any panic situation?

pavannd1: I don't see any usage of `blockstorage.Snapshot.Volume` for EFS but want to make sure it's okay to set it to `nil`

PR #215, opened 2019-08-19
Add AssumeRole support for EFS provider

hakanmemisoglu:
## Change Overview

Adds AssumeRole support for EFS provider.

## Pull request type

- [X] Feature

### DIFF
--- a/pkg/blockstorage/awsebs/awsebs.go
+++ b/pkg/blockstorage/awsebs/awsebs.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -21,6 +20,7 @@ import (
 	"github.com/kanisterio/kanister/pkg/blockstorage"
 	ktags "github.com/kanisterio/kanister/pkg/blockstorage/tags"
 	"github.com/kanisterio/kanister/pkg/blockstorage/zone"
+	awsconfig "github.com/kanisterio/kanister/pkg/config/aws"
 )
 
 var _ blockstorage.Provider = (*ebsStorage)(nil)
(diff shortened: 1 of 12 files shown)


pkg/blockstorage/awsefs/awsefs.go
```diff
@@ -61,13 +62,23 @@ func NewEFSProvider(config map[string]string) (blockstorage.Provider, error) {
 		return nil, errors.New("Account ID is empty")
 	}
 	accountID := *user.Account
-	efsCli := awsefs.New(s, aws.NewConfig().WithRegion(region))
-	backupCli := backup.New(s, aws.NewConfig().WithRegion(region))
+	creds := awsConfig.Credentials
```
vkamra: @hakanmemisoglu - this makes sense to me i.e. adding the ability to accept an AWS role as an input and using that if specified - this is the direction we want to head in for all AWS operations.

@julio-lopez - second pair of eyes please? Some context on why we're adding this. Instead of creating a "test user" that has the credentials required - we've created a "customer permissions" role - that we will allow all users to assume. The EFS tests are the first one to use that functionality

pkg/blockstorage/awsefs/awsefs.go
```diff
@@ -61,13 +62,23 @@ func NewEFSProvider(config map[string]string) (blockstorage.Provider, error) {
...
 		EFS:       efsCli,
 		Backup:    backupCli,
 		region:    region,
-		accountID: accountID}, nil
+		accountID: accountID,
+	}, nil
+}
+
+func roleARN(accountID, roleName string) string {
+	return fmt.Sprintf("arn:aws:iam::%s:role/%s", accountID, roleName)
```
vkamra: I think we should accept "roleArn" as an input instead of roleName. It is valid to specify a role that is in a different account

tdmanv:
This change looks good and makes sense to me.
I have a high level question though - @hakanmemisoglu do you have a sense for how this parameter will be specified from the callers? What is the end mechanism to configure this?

hakanmemisoglu:
Blockstorage structs are mainly used by the structs in function package. These function structs are responsible for setting up the config. 

I think, as a next step, we should add EFS support in these functions. Config should be set up exactly how EBS is set up. As an additional step, we can also check Role information. If we find one, role argument should be passed to the provider.

Right now, configuration building is done by checking credentials in profile and the info from PVC (in EBS case). We can add define additional key-pair for AssumeRole ARN in credentials. Or  we can add operation parameters argument to the whole process; so PVC info, secret credentials and operation parameters together will be sufficient to run Kanister ops (functions).

pkg/blockstorage/awsebs/awsebs.go
```diff
@@ -66,21 +69,22 @@ func NewProvider(config map[string]string) (blockstorage.Provider, error) {
 }
 
 // GetConfig returns a configuration to establish AWS connection and the connected region name.
-func GetConfig(config map[string]string) (*aws.Config, string, error) {
+func GetConfig(config map[string]string) (*aws.Config, string, string, error) {
```
julio-lopez: It is not immediately obvious what the first and second return strings mean, even the first one alone for that matter. Given that the return types are not self documenting, let's use named return values here.
Also, please update the function comment to reflect the returned arguments.

pkg/blockstorage/awsefs/awsefs.go
julio-lopez: It seems that we should encapsulate the support for AWS roles as part of creating either the config or the session for AWS ops. This could be used for block and potentially object store operations as well.
For now, it may be enough to factor out the role-related functionality into a separate function, and then we can figure out how to generalize it. Thoughts?

PR #226, opened 2019-08-22
Update ScaleWorkload to provide replicas as string & int value

SupriyaKasten:
## Change Overview

ScaleWorkload only allowed static int values for `replicas` parameter. With these changes, we can provide string values to `replicas` in addition to int, kanister will take care of conversion.

## Pull request type

- [ ] Work in Progress
- [ ] Trival/Minor
- [ ] Bugfix
- [x] Feature
- [ ] Documentation

## Test Plan

- [ ] Manual
- [x] Unit test
- [ ] E2E

### DIFF
--- a/pkg/function/scale_test.go
+++ b/pkg/function/scale_test.go
@@ -101,7 +101,7 @@ func newScaleBlueprint(kind string) *crv1alpha1.Blueprint {
 						Name: "testScale",
 						Func: "ScaleWorkload",
 						Args: map[string]interface{}{
-							ScaleWorkloadReplicas: 2,
+							ScaleWorkloadReplicas: "2",
 						},
 					},
 				},
@@ -234,7 +234,7 @@ func (s *ScaleSuite) TestGetArgs(c *C) {
 		{
 			tp: param.TemplateParams{},
 			args: map[string]interface{}{
-				ScaleWorkloadReplicas:     2,
+				ScaleWorkloadReplicas:     "2",
 				ScaleWorkloadNamespaceArg: "foo",
 				ScaleWorkloadNameArg:      "app",
 				ScaleWorkloadKindArg:      param.StatefulSetKind,
@@ -269,7 +269,7 @@ func (s *ScaleSuite) TestGetArgs(c *C) {
 				},
 			},
 			args: map[string]interface{}{
-				ScaleWorkloadReplicas: 2,
+				ScaleWorkloadReplicas: "2",
 			},
 			wantKind:      param.DeploymentKind,
 			wantName:      "app",
--- a/pkg/function/scale_workload.go
+++ b/pkg/function/scale_workload.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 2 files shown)


pkg/function/scale_workload.go
```diff
@@ -59,11 +60,19 @@ func (*scaleWorkloadFunc) RequiredArgs() []string {
...
 
 func getArgs(tp param.TemplateParams, args map[string]interface{}) (namespace, kind, name string, replicas int32, err error) {
-	err = Arg(args, ScaleWorkloadReplicas, &replicas)
+	var rep interface{}
+	err = Arg(args, ScaleWorkloadReplicas, &rep)
 	if err != nil {
 		return namespace, kind, name, replicas, err
 	}
-
+	if val, ok := rep.(int); ok {
```
tdmanv: please use a type switch here: https://tour.golang.org/methods/16

pkg/function/scale_workload.go
```diff
@@ -59,11 +60,19 @@ func (*scaleWorkloadFunc) RequiredArgs() []string {
...
+		replicas = int32(val)
+	} else if val, ok := rep.(string); ok {
+		strToInt := intstr.Parse(val)
```
tdmanv: let's just use https://golang.org/pkg/strconv/#Atoi

pkg/function/scale_workload.go
```diff
@@ -59,11 +60,19 @@ func (*scaleWorkloadFunc) RequiredArgs() []string {
...
+		replicas = strToInt.IntVal
+	} else {
+		return namespace, kind, name, replicas, errors.Wrapf(err, "Failed to decode arg `%s`", ScaleWorkloadReplicas)
```
tdmanv: typically I'd use a naked return with named return values. 
```suggestion
		err = errors.Errorf("Invalid arg type %T for Arg %s ", rep, ScaleWorkloadReplicas)
		return
```

pkg/function/scale_test.go
```diff
@@ -269,7 +269,7 @@ func (s *ScaleSuite) TestGetArgs(c *C) {
 				},
 			},
 			args: map[string]interface{}{
-				ScaleWorkloadReplicas: 2,
+				ScaleWorkloadReplicas: "2",
```
tdmanv: Let's leave one of these an int to test both codepaths

SupriyaKasten: @tdmanv we do have two of these as int.

pkg/function/scale_workload.go
```diff
@@ -59,11 +60,27 @@ func (*scaleWorkloadFunc) RequiredArgs() []string {
...
 
 func getArgs(tp param.TemplateParams, args map[string]interface{}) (namespace, kind, name string, replicas int32, err error) {
-	err = Arg(args, ScaleWorkloadReplicas, &replicas)
+	var rep interface{}
+	err = Arg(args, ScaleWorkloadReplicas, &rep)
 	if err != nil {
 		return namespace, kind, name, replicas, err
 	}
 
+	var val int
```
pavannd1: [minor] Should this be a function? Okay to be done as an enhancement later

PR #227, opened 2019-08-23
Do not create build artifacts in GOOS directory

vkamra:
## Change Overview

Mapping `"$(PWD)/bin/$(ARCH)/$$(go env GOOS)_$(ARCH)` to -> `/go/bin` inside the
build container would result in artifacts being placed under `bin/amd64/darwin_amd64`
when building on OS X - which is not correct since these are not OS X executables

This also broke `make deploy` and `make release-container` commands

## Pull request type

- [ ] Work in Progress
- [ ] Trival/Minor
- [x] Bugfix
- [ ] Feature
- [ ] Documentation

## Test Plan

- [x] Manual
- [ ] Unit test
- [ ] E2E

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,6 @@ shell: build-dirs
 		-v "${HOME}/.kube:/root/.kube"                          \
 		-v "$(PWD):/go/src/$(PKG)"                              \
 		-v "$(PWD)/bin/$(ARCH):/go/bin"                         \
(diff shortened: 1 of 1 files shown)


vkamra:
Hold off on merging - I want to make sure I understand why this was there to start with.

vkamra:
> Hold off on merging - I want to make sure I understand why this was there to start with.

Should be good to go.

Issue #231, opened 2019-08-23
Continuous logs from spawned commands


tdmanv:
KubeTask and KubeExec currently only print logs after completion and for failures don't print anything. This makes it difficult to determine what happened during execution.

tdmanv:
`KubeTask` and several other Functions use `PodRunner`: https://github.com/kanisterio/kanister/blob/d1228b3b25825d842c400bc3cfca45201c977011/pkg/kube/pod_runner.go#L27. 

I think it makes sense to consolidate log handling into this package.

We use PodRunner in the following Functions: https://github.com/kanisterio/kanister/search?l=Go&q=NewPodRunner&type=

Many of these Functions do the following:

1. Wait for the pod to be ready
2. Wait for the to complete
3. Collect logs
4. Parse output

This functionality can be improved and pushed into PodRunner.

PR #264, opened 2019-09-06
Add log reader helper

tdmanv:
## Change Overview

* Adds a helper to get logs w/ the `follow=true`

## Pull request type

- [ ] Work in Progress
- [ ] Trival/Minor
- [ ] Bugfix
- [x] Feature
- [ ] Documentation

## Issues

- https://github.com/kanisterio/kanister/issues/231

## Test Plan

```/go/src/github.com/kanisterio/kanister # go test -v ./pkg/kube/ -check.v -check.f LogReaderSuite
=== RUN   Test
PASS: log_reader_test.go:41: LogReaderSuite.TestLogReader       0.000s
OK: 1 passed
--- PASS: Test (0.00s)
PASS
ok      github.com/kanisterio/kanister/pkg/kube 0.008s
```

- [ ] Manual
- [x] Unit test
- [ ] E2E

### DIFF
--- a/pkg/kube/log_reader.go
+++ b/pkg/kube/log_reader.go
@@ -1,0 +1,41 @@
+package kube
+
+import (
+	"io"
+
+	"k8s.io/client-go/rest"
+)
+
+var _ io.ReadCloser = (*logReader)(nil)
+
+func newLogReader(rw rest.ResponseWrapper) *logReader {
+	return &logReader{rw: rw}
+}
+
+// logReader defers a call to ResponseWrapper.Stream() until Read is called.
+// This is to help handle kubernetes behavior where calls to
+// Pod.GetLogs(...).Stream() will hang until at least one byte is returned.
+// kubectl logs handles this in
+// https://github.com/kubernetes/kubernetes/pull/67573/files#diff-12d472fe036bbe778e84dffc71564eb1R355
+type logReader struct {
+	rw  rest.ResponseWrapper
+	err error
+	rc  io.ReadCloser
+}
+
+func (lr *logReader) Read(p []byte) (n int, err error) {
+	if lr.rc == nil {
+		lr.rc, lr.err = lr.rw.Stream()
+	}
+	if lr.err != nil {
+		return 0, lr.err
+	}
+	return lr.rc.Read(p)
+}
+
+func (lr *logReader) Close() error {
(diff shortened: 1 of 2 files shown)


pkg/kube/log_reader.go
```diff
@@ -0,0 +1,41 @@
...
+// https://github.com/kubernetes/kubernetes/pull/67573/files#diff-12d472fe036bbe778e84dffc71564eb1R355
+type logReader struct {
+	rw  rest.ResponseWrapper
+	err error
+	rc  io.ReadCloser
+}
+
+func (lr *logReader) Read(p []byte) (n int, err error) {
+	if lr.rc == nil {
+		lr.rc, lr.err = lr.rw.Stream()
```
julio-lopez: A potential downside here is that the connection is not pre-established. But that may be OK for the use case.

pkg/kube/log_reader.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// logReader defers a call to ResponseWrapper.Stream() until Read is called.
+// This is to help handle kubernetes behavior where calls to
+// Pod.GetLogs(...).Stream() will hang until at least one byte is returned.
+// kubectl logs handles this in
+// https://github.com/kubernetes/kubernetes/pull/67573/files#diff-12d472fe036bbe778e84dffc71564eb1R355
+type logReader struct {
+	rw  rest.ResponseWrapper
+	err error
```
julio-lopez: What is `err` used for?

julio-lopez: NM, saw it

tdmanv: this could be used in the Read method. Let me push that down

pkg/kube/log_reader.go
julio-lopez: Also, this has an "interesting" behavior.
If the first read fails, I may be able to call `Read()` again, and have it succeed later, which is surprising.
An error from `rc.Read()` is not cached, but I'm guessing the expectation is that if `rc.Read()` fails once, then it will fail from then on.

tdmanv: I agree it would be less surprising to return an error if it was cached. I'll change the check to:
```
if lr.rc == nil || lr.err != nil {
```

PR #272, opened 2019-09-10
Initial structure of log repo

SupriyaKasten:
## Change Overview

Make logs more descriptive and easy to use & read

This task will be divided in small PRs and merged
This PR is the initial structure of the log repo

## Pull request type

- [ ] Work in Progress
- [ ] Trival/Minor
- [ ] Bugfix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -5,35 +5,37 @@ import (
 	"fmt"
 
 	"github.com/sirupsen/logrus"
-
-	"github.com/pkg/field"
 )
 
-type logLevel string
+// Level describes the current log level.
+type Level uint32
 
 const (
-	logTypeInfo  logLevel = "Info"
-	logTypeError logLevel = "Error"
-	logTypeDebug logLevel = "Debug"
+	// DebugLevel log level.
+	DebugLevel Level = Level(logrus.DebugLevel)
+	// InfoLevel log level.
+	InfoLevel Level = Level(logrus.InfoLevel)
+	// ErrorLevel log level.
+	ErrorLevel Level = Level(logrus.ErrorLevel)
 )
 
 type logger struct {
-	level logLevel
+	level Level
 	entry *logrus.Entry
 	ctx   context
 	err   error
 }
 
 func Info() Printer {
-	return &logger{level: logTypeInfo}
+	return &logger{level: InfoLevel}
 }
 
 func Error() Printer {
-	return &logger{level: logTypeError}
+	return &logger{level: ErrorLevel}
 }
 
 func Debug() Printer {
-	return &logger{level: logTypeDebug}
+	return &logger{level: DebugLevel}
 }
 
 // Most commonly used logging function
@@ -47,11 +49,11 @@ func WithContext(ctx) {
 
 func (l *logger) Print(msg string) {
 	switch l.level {
-	case logTypeInfo:
+	case InfoLevel:
 		logrus.Info(msg...)
 	case logTypeError:
-		logrus.Error(msg...)
-	case logTypeDebug:
+		ErrorLevel.Error(msg...)
+	case DebugLevel:
 		logrus.Debug(msg...)
 	}
 }
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -5,74 +5,47 @@ import (
 	"fmt"
 
 	"github.com/sirupsen/logrus"
-)
 
-// Field type TBD.
-type F interface{}
+	"github.com/pkg/field"
+)
 
 type logLevel string
 
 const (
-	logTypeInfo        logLevel = "Info"
-	logTypeError       logLevel = "Error"
-	logTypeDebug       logLevel = "Debug"
-	logTypeWithFields  logLevel = "WithFields"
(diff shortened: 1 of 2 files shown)


pkg/log/log.go
```diff
@@ -0,0 +1,67 @@
+package log
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/pkg/field"
```
SupriyaKasten: Removing this import

pkg/log/log.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type logLevel string
+
+const (
+	logTypeInfo  logLevel = "Info"
```
tdmanv: Rather than use string literals, please use the constants from logrus

pkg/log/log.go
tdmanv: is this used in this package yet?

pkg/log/log.go
```diff
@@ -0,0 +1,92 @@
...
+// Field type TBD.
+type F interface{}
+
+type logLevel string
+
+const (
+	logTypeInfo        logLevel = "Info"
+	logTypeError       logLevel = "Error"
+	logTypeDebug       logLevel = "Debug"
+	logTypeWithFields  logLevel = "WithFields"
```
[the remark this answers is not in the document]
SupriyaKasten: @tdmanv  This is from the old commit. New commit has this removed.

pkg/log/log.go
```diff
@@ -0,0 +1,69 @@
...
+func WithContext(ctx) {
+	Info().WithContext(ctx)
+}
+
+func (l *logger) Print(msg string) {
+	switch l.level {
+	case InfoLevel:
+		logrus.Info(msg...)
+	case logTypeError:
+		ErrorLevel.Error(msg...)
```
tdmanv: Why not
```suggestion
		logrus.Error(msg...)
```

pkg/log/printer.go
```diff
@@ -0,0 +1,11 @@
+package log
+
+import (
+	"context"
+)
+
+type Printer interface {
```
julio-lopez: At this point, it makes sense to rename this interface as `Logger`, since now it has more than just the `Print` function
```suggestion
type Logger interface {
```

pkg/log/printer.go
```diff
@@ -0,0 +1,11 @@
...
+	Print(msg string)
+	WithContext(ctx context.Context) Printer
```
julio-lopez:
```suggestion
	WithContext(ctx context.Context) Logger
```

pkg/log/printer.go
```diff
@@ -0,0 +1,11 @@
...
+	WithError(err error) Printer
```
julio-lopez:
```suggestion
	WithError(err error) Logger
```

pkg/log/log.go
```diff
@@ -0,0 +1,68 @@
...
+
+func Error() Printer {
+	return &logger{level: ErrorLevel}
+}
+
+func Debug() Printer {
+	return &logger{level: DebugLevel}
+}
+
+// Most commonly used logging function
```
julio-lopez: Need to fix the comment.

And add comments for the functions above. This should have been caught by the linter.
```suggestion
// Print adds `msg` to the log at `InfoLevel`. It is a wrapper for `Info().Print(msg)`, since this is the most common use case.
```

pkg/log/log.go
```diff
@@ -0,0 +1,68 @@
...
+	// DebugLevel log level.
+	DebugLevel Level = Level(logrus.DebugLevel)
+	// InfoLevel log level.
+	InfoLevel Level = Level(logrus.InfoLevel)
+	// ErrorLevel log level.
+	ErrorLevel Level = Level(logrus.ErrorLevel)
+)
+
+type logger struct {
+	level Level
```
julio-lopez: We can revise the implementation later. For now, this interface helps modifying / adding the call points.

SupriyaKasten:
@julio-lopez  I did not commit your suggestions directly since I wanted to make some more changes. Pushed a PR with your suggestions, Ready for another review!

pkg/log/log.go
```diff
@@ -47,22 +47,24 @@ func WithContext(ctx context.Context) {
 }
 
 func (l *logger) Print(msg string) {
+	var message interface{}
```
julio-lopez: Is this needed?

SupriyaKasten: it was failing before with: `pkg/log/log.go:52:14: cannot use msg (type string) as type []interface {} in argument to logrus.Info`

julio-lopez: OK.

SupriyaKasten: Oh I think I know where I went wrong! pushing an updated PR.

SupriyaKasten: Sorry for the confusion @julio-lopez, I was calling `logrus.Info(msg...)` instead if `logrus.Info(msg)`. Fixing it now

Issue #273, opened 2019-09-10
Structured fields support for logging


julio-lopez:
In order to facilitate debugging and troubleshooting, it is desirable to include contextual information in errors and log messages in a structured manner. In particular, it is desirable to preserve semantic information about the values of request parameters or other variables in a way that makes it easy to later extract the original semantic meaning without having to perform log parsing and inference. At a high level, imagine if log messages where structured records (JSON documents) instead of simply arbitrary strings.

As a first step, it is desirable to have a simple mechanism and API that allows capturing the semantic boundaries of different fields and data types pushed through logging, or attached to errors for that matter.

One potential approach to address this issue is to include both in logs and errors, a collection (think list) of `(key, value)` tuples or fields. It would be desirable to be able to attach these fields to `context.Context` and extract those at logging time.

julio-lopez:
Basic functionality in #276 and #279. Extensions will be added based on the use cases.

PR #279, opened 2019-09-10
An implementation for field.Field

julio-lopez:
## Change Overview

Linked-list based implementation.
The approach attempts to prevent/limit the mutability of the fields outside the `field` package.

Also, in the package the contents of `linkedField struct` are not modified after the struct has been created.

## Pull request type

- [x] Work in Progress
- [x] Feature

## Issues

- #273 

## Test Plan

- [x] Unit test

### DIFF
--- a/pkg/field/linked_field.go
+++ b/pkg/field/linked_field.go
@@ -1,0 +1,83 @@
+package field
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Fields are exposed as an interface instead of directly as pointer to this
+// struct. That enforces the immutability of fields outside this package. Once
+// a field is created, its contents cannot be changed or overwritten outside
+// this package by simply assigning one field to another, via
+// f1 = f2 or *fp1 = *fp2.
+type field struct {
+	key   string
+	value interface{}
+}
+
+func (f field) Key() string {
+	return f.key
+}
+
+func (f field) Value() interface{} {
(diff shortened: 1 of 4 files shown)


julio-lopez:
Minimum number of exported functions.

We may need to add other functions/helpers down the road to support other use cases for logging and errors.

pkg/field/linked_field.go
```diff
@@ -0,0 +1,83 @@
...
+// Fields are exposed as an interface instead of directly as pointer to this
+// struct. That enforces the immutability of fields outside this package. Once
+// a field is created, its contents cannot be changed or overwritten outside
+// this package by simply assigning one field to another, via
+// f1 = f2 or *fp1 = *fp2.
+type field struct {
+	key   string
+	value interface{}
+}
+
```
tdmanv: [minor]
```suggestion
var _ fmt.Stringer = field{}
var _ Field = field{}

```

julio-lopez: This one makes sense:
`var` _ fmt.Stringer = field{}`

However, `var _ Field = field{}` doesn't.

See [interface checks](https://golang.org/doc/effective_go.html#blank_implements)

> Don't do this for every type that satisfies an interface, though. By convention, such declarations are only used when there are no static conversions already present in the code, which is a rare event.

PR #280, opened 2019-09-10
Make empty config an error in EBS client constructor

hakanmemisoglu:
## Change Overview

This PR fixes the panic problem where the argument config is nil.

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [X] Bugfix
- [ ] Feature
- [ ] Documentation

## Test Plan

- [X] Manual
- [ ] Unit test
- [ ] E2E

### DIFF
--- a/pkg/blockstorage/awsebs/awsebs.go
+++ b/pkg/blockstorage/awsebs/awsebs.go
@@ -23,7 +23,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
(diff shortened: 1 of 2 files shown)


pkg/blockstorage/awsebs/awsebs.go
```diff
@@ -76,6 +76,9 @@ func NewProvider(config map[string]string) (blockstorage.Provider, error) {
 
 // newEC2Client returns ec2 client struct.
 func newEC2Client(awsRegion string, config *aws.Config, role string) (*EC2, error) {
+	if config == nil {
+		config = aws.NewConfig()
+	}
```
pavannd1: Will the default config have valid `Credentials` set? Is it going to affect the flow in any way?

tdmanv: Should we instead return an error if the config is nil and make the caller get the config?

hakanmemisoglu:
@tdmanv @pavannd1 Updated the PR.

PR #291, opened 2019-09-13
Add pod and container name to log context

SupriyaKasten:
## Change Overview

Add pod name and container name to context in addition to actionset name and phase name 

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/pkg/function/restore_data.go
+++ b/pkg/function/restore_data.go
@@ -23,7 +23,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	kanister "github.com/kanisterio/kanister/pkg"
-	"github.com/kanisterio/kanister/pkg/field"
 	"github.com/kanisterio/kanister/pkg/format"
 	"github.com/kanisterio/kanister/pkg/kube"
 	"github.com/kanisterio/kanister/pkg/param"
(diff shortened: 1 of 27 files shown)


pkg/function/restore_data.go
```diff
@@ -176,6 +177,10 @@ func (*restoreDataFunc) Exec(ctx context.Context, tp param.TemplateParams, args
 	}
 	// Validate and get optional arguments
 	restorePath, encryptionKey, pod, vols, backupTag, backupID, err := validateAndGetOptArgs(args)
+
+	if pod != "" {
```
tdmanv: doesn't this Function also use pod runner?

pkg/function/restore_data_all.go
```diff
@@ -128,6 +129,7 @@ func (*restoreDataAllFunc) Exec(ctx context.Context, tp param.TemplateParams, ar
 	output := make(map[string]interface{})
 	for _, pod := range pods {
 		go func(pod string) {
+			ctx = field.Context(ctx, field.PodNameKey, pod)
```
tdmanv: same here

pkg/function/restore_data.go
SupriyaKasten: It does, but the pod name set here is different than the one that is generated by pod runner.

tdmanv: I would either drop this or rename the key since it's getting overwritten anyway

julio-lopez: Use a different key in order to tell apart the 2 pod names?

PR #292, opened 2019-09-16
Extract AWS credentials from secret

hakanmemisoglu:
## Change Overview

Extract AWS credentials from AWS typed secret.

* Validate AWS credential secret.
* Extract required `awsAccessKeyID` and `awsSecretAccessKey` fields from the secret.
* Extract optional `awsSessionToken` from the field.

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [X] Feature
- [ ] Documentation

## Test Plan

- [ ] Manual
- [X] Unit test
- [ ] E2E


pkg/secrets/aws.go
```diff
@@ -0,0 +1,48 @@
...
+// ExtractAWSCredentials extracts AWS credential values from the given secret.
+//
+// Extracted values from the secrets are:
+// - AWS_ACCESS_KEY_ID (required)
+// - AWS_SECRET_ACCESS_KEY (required)
+// - AWS_SESSION_TOKEN (optional)
+//
+// If the type of the secret is not "secret.kanister.io/aws", it returns an error.
+// If the required types are not avaialable in the secrets, it returns an errror.
+func ExtractAWSCredentials(secret *v1.Secret) (*credentials.Value, error) {
```
tdmanv: Let's split Extract from Validate.

[minor] we can also validate that there are not extra fields in data.

pkg/secrets/aws.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	// AWSSecretType represent the secret type for AWS credentials.
+	AWSSecretType string = "secrets.kanister.io/aws"
+
+	// AWSAccessKeyID is the key for AWS access key ID.
+	AWSAccessKeyID string = "AWS_ACCESS_KEY_ID"
```
tdmanv: typically the keys names are lowercased.

pkg/secrets/aws_test.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	. "gopkg.in/check.v1"
+	v1 "k8s.io/api/core/v1"
+)
+
+// Hook up gocheck into the "go test" runner.
+func Test(t *testing.T) { TestingT(t) }
```
tdmanv: [minor] usually we register the tests in a file with the same name as the package.

pkg/secrets/aws.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	// AWSSecretType represent the secret type for AWS credentials.
+	AWSSecretType string = "secrets.kanister.io/aws"
+
+	// AWSAccessKeyID is the key for AWS access key ID.
+	AWSAccessKeyID string = "awsAccessKeyID"
```
julio-lopez: Why not use the same values that the AWS SDK and CLI use? That's what mot people would expect these (environment) variables to be named, and thus to be the corresponding keys in the secret

pkg/secrets/aws.go
julio-lopez: @tdmanv See my comment.

pkg/secrets/aws.go
hakanmemisoglu: @tdmanv suggested we are using camel case with secret data fields in Kanister.

tdmanv: I think I said lowercase. The field names I've seen elsewhere in our code are:
```
    access_key_id: XXX
    secret_access_key: XXX
```

pkg/secrets/aws.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	// AWSSecretType represents the secret type for AWS credentials.
+	AWSSecretType string = "secrets.kanister.io/aws"
```
tdmanv: [minor] do these need to be exported?

hakanmemisoglu: Not needed right now. However I think some Kanister package sure can use this value to validate their own input.

pkg/secrets/aws.go
julio-lopez: @tdmanv lower case is _"surprising"_, what's the rationale for it?
With the given key names, it would not be possible to _directly_ [export all key-value pairs in a secret as environment variables](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#configure-all-key-value-pairs-in-a-secret-as-container-environment-variables).

pkg/secrets/aws.go
```diff
@@ -0,0 +1,72 @@
...
+		return errors.New("awsAccessKeyID is a required field")
+	}
+	if _, ok := secret.Data[AWSSecretAccessKey]; !ok {
+		return errors.New("awsSecretAccessKey is a required field")
+	}
+	count := 2
+	if _, ok := secret.Data[AWSSessionToken]; ok {
+		count++
+	}
+	if len(secret.Data) > count {
```
julio-lopez: This check seems too restrictive in the sense that it should be possible to easily extend the usage of these secrets by adding other fields without having to go and explicitly modify code that uses these secrets. In particular, this is not _forward_ compatible. That is, in the future we decide to include other fields, suppose the role or some other metadata instead of the token, (a) it would require validation changes to allow the new definition and (b) older versions of the code would break when encountering the new types of secrets.

PR #293, opened 2019-09-17
Add definition to log.Print() func

SupriyaKasten:
## Change Overview

Function definition for log.Print() func

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [x] Feature
- [ ] Documentation

## Test Plan

- [ ] Manual
- [ ] Unit test
- [ ] E2E

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -48,40 +48,32 @@ func WithContext(ctx context.Context) {
 	Info().WithContext(ctx)
 }
 
+func WithError(err error) {
+	Error().WithError(err)
+}
+
 func (l *logger) Print(msg string) {
 	logFields := make(logrus.Fields)
 	if l.ctx != nil {
-		ctxFields := field.FromContext(l.ctx)
-		for _, f := range ctxFields {
+		ctxFields := field.FromContext(l.ctx).Fields()
+		for _, cf := range ctxFields {
 			logFields[cf.Key()] = cf.Value()
 		}
 	}
-	if len(logFields) > 0 {
-		if l.entry != nil {
-			l.entry = l.entry.WithFields(logFields)
-		} else {
-			l.entry = logrus.WithFields(logFields)
-		}
+
+	if l.entry != nil {
+		l.entry = l.entry.WithFields(logFields)
+	} else {
+		l.entry = logrus.WithFields(logFields)
 	}
+
 	switch l.level {
 	case InfoLevel:
-		if l.entry != nil {
-			l.entry.Info(msg)
-		} else {
-			logrus.Info(msg)
-		}
+		l.entry.Info(msg)
 	case ErrorLevel:
-		if l.entry != nil {
-			l.entry.Error(msg)
-		} else {
-			logrus.Error(msg)
-		}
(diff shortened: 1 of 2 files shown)


pkg/log/log.go
```diff
@@ -46,23 +48,62 @@ func WithContext(ctx context.Context) {
 	Info().WithContext(ctx)
 }
 
-func (l *logger) Print(msg string) {
+func (l *logger) Print(msg string, fields field.Fields) {
+	logFields := make(logrus.Fields)
+	if l.ctx != nil {
+		ctxFields := field.FromContext(l.ctx)
```
tdmanv: does this handle nil? Can we make it handle nil?

pkg/log/log.go
```diff
@@ -46,23 +48,62 @@ func WithContext(ctx context.Context) {
 	Info().WithContext(ctx)
 }
 
-func (l *logger) Print(msg string) {
+func (l *logger) Print(msg string, fields field.Fields) {
```
tdmanv: Please decouple accepting fields as a param and extracting the fields from the context.

julio-lopez: For now, let's have the `Print` signature be:
```suggestion
func (l *logger) Print(msg string) {
```
The `fields` argument is not immediately needed, right? They will be extracted from the context, no?
We can easily extend the `Print` function later to accept other arguments. However, we have not gone through the exercise of evaluating the tradeoffs of different alternatives.
We decided to use `WithContext()` and `WithError()` for now as a mean to achieve the immediate desired functionality.

pkg/log/log.go
```diff
@@ -46,23 +48,50 @@ func WithContext(ctx context.Context) {
 	Info().WithContext(ctx)
 }
 
+func WithError(err error) {
+	Error().WithError(err)
+}
+
 func (l *logger) Print(msg string) {
+	logFields := make(logrus.Fields)
+	if ctxFields := field.FromContext(l.ctx); ctxFields != nil {
```
tdmanv:
```suggestion
	if ctxFields := field.FromContext(l.ctx); len(ctxFields) != 0 {
```

SupriyaKasten: @tdmanv ctxFields is of type field.Fieldser interface. I think `len(ctxFields)` will fail.

tdmanv: you're right! please ignore this comment

pkg/log/log.go
```diff
@@ -46,23 +48,50 @@ func WithContext(ctx context.Context) {
...
+		l.entry.Error(msg)
 	case DebugLevel:
-		logrus.Debug(msg)
+		l.entry.Debug(msg)
 	}
 }
 
 func (l *logger) WithContext(ctx context.Context) Logger {
 	l.ctx = ctx
+	if l.entry != nil {
```
julio-lopez: Later we can figure out an idiomatic way of avoiding having to always check whether entry is `nil` or not.
And in particular, it is desirable to make "logger" immutable. That's mostly implementation and would be compatible with the API currently exposed.

PR #294, opened 2019-09-17
Add a secret union field for credentials 

hakanmemisoglu:
## Change Overview

We need a new credentials type to accept optional AWS session token.

* Adds new credential type to accept generic credentials.

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [X] Feature
- [ ] Documentation


pkg/apis/cr/v1alpha1/types.go
```diff
@@ -249,12 +249,14 @@ type CredentialType string
 
 const (
 	CredentialTypeKeyPair CredentialType = "keyPair"
+	CredentialTypeAws     CredentialType = "aws"
```
tdmanv: Rather than include the type `Aws` here, let's make this more general purpose since the type will already be captured by the secret.

hakanmemisoglu:
@hakanmemisoglu Converted it to `TypedSecret`. I am open for suggestions for the name.

pkg/apis/cr/v1alpha1/types.go
```diff
@@ -248,13 +248,15 @@ type Location struct {
 type CredentialType string
 
 const (
-	CredentialTypeKeyPair CredentialType = "keyPair"
+	CredentialTypeKeyPair     CredentialType = "keyPair"
+	CredentialTypeTypedSecret CredentialType = "typedSecret"
```
pavannd1: Will it be possible to use the `secret` field inside `KeyPair`?

hakanmemisoglu: Don't we want a good separation between these two?
Do you think add a new union member struct is bad for us?

pavannd1: Just thinking of ways to make it generic. Will the secret contain just the role or even the AWS key and ID?

hakanmemisoglu: Got it. It will have only secret field. The reason I am trying to make it separate is that the consumer code should understand which type credentials they are by looking at the type. Then it should decode the data in the secret reference with a know logic. For example in this case, if it is typed secret, then fetch the secret with its reference. Then check the type of the secret. If it is a AWS typed secret, then fetch the well-known keys from its data section.

pavannd1: Okay, thanks for the clarification.

pkg/apis/cr/v1alpha1/types.go
```diff
@@ -249,12 +249,14 @@ type CredentialType string
...
+	CredentialTypeSecret  CredentialType = "secret"
 )
 
 // Credential
 type Credential struct {
-	Type    CredentialType `json:"type"`
-	KeyPair *KeyPair       `json:"keyPair"`
+	Type    CredentialType   `json:"type"`
+	KeyPair *KeyPair         `json:"keyPair"`
+	Secret  *ObjectReference `json:"aws"`
```
julio-lopez: Shouldn't the JSON name be `secret` ?

hakanmemisoglu: Right, let me update.

PR #298, opened 2019-09-17
Handle nil context value in field.FromContext() func

SupriyaKasten:
## Change Overview

Handle nil context value in field.FromContext() func

## Pull request type

- [ ] Work in Progress
- [x] Trivial/Minor
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/pkg/field/context.go
+++ b/pkg/field/context.go
@@ -10,8 +10,10 @@ const ctxKey = contextKey(43)
 
 // FromContext returns the fields present in ctx if any
 func FromContext(ctx context.Context) Fields {
-	if s, ok := ctx.Value(ctxKey).(Fields); ok {
-		return s
+	if ctx != nil {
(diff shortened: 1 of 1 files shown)


pkg/field/context.go
```diff
@@ -10,8 +10,10 @@ const ctxKey = contextKey(43)
 
 // FromContext returns the fields present in ctx if any
 func FromContext(ctx context.Context) Fields {
-	if s, ok := ctx.Value(ctxKey).(Fields); ok {
-		return s
+	if ctx != nil {
```
julio-lopez: OTOH, this should be handled by the caller. For example, `context.WithDeadline()` does not accept a `nil` context.
However, this was not explicitly documented in `FromContext` from the beginning, and overall this approach is more robust.

PR #300, opened 2019-09-18
Add secret type for Param.Credentials

hakanmemisoglu:
## Change Overview

Adds a union member for param.Credentials to accommodate generic secret types.

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/pkg/param/param.go
+++ b/pkg/param/param.go
@@ -91,26 +91,20 @@ type CredentialType string
 
 const (
 	CredentialTypeKeyPair CredentialType = "keyPair"
-	CredentialTypeAWS     CredentialType = "aws"
+	CredentialTypeSecret  CredentialType = "secret"
 )
 
 // Credential resolves the storage
 type Credential struct {
 	Type    CredentialType
 	KeyPair *KeyPair
-	Aws     *AWSSecret
+	Secret  *v1.Secret
 }
 
 // KeyPair is a credential that contains two strings: an ID and a secret.
 type KeyPair struct {
 	ID     string
 	Secret string
-}
-
-type AWSSecret struct {
-	AccessKeyID  string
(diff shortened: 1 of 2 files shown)


pkg/param/param.go
```diff
@@ -90,12 +91,14 @@ type CredentialType string
 
 const (
 	CredentialTypeKeyPair CredentialType = "keyPair"
+	CredentialTypeAWS     CredentialType = "aws"
```
pavannd1: Should we make this `CredentialTypeSecret` too? In case we want something similar for Azure.
@SupriyaKasten

hakanmemisoglu: My thinking was, at this point, we should materialize the type information. For @SupriyaKasten's case, we can also add `CredentialTypeAzure`.

pkg/param/param.go
```diff
@@ -237,6 +242,32 @@ func fetchKeyPairCredential(ctx context.Context, cli kubernetes.Interface, c *cr
...
 
+func fetchSecretCredential(ctx context.Context, cli kubernetes.Interface, sr *crv1alpha1.ObjectReference) (*Credential, error) {
+	if sr == nil {
+		return nil, errors.New("Secret reference cannot be nil")
+	}
+	s, err := cli.CoreV1().Secrets(sr.Namespace).Get(sr.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to fetch the secret")
+	}
+	switch string(s.Type) {
```
tdmanv: I would push this logic into the secrets package and expose it through `ValidateCredential` or just validate

PR #305, opened 2019-09-19
Support creating S3 args from Secret type Credential

hakanmemisoglu:
## Change Overview

Support creating Restic S3 args from Secret type Credentials.

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/pkg/restic/restic.go
+++ b/pkg/restic/restic.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 
 	crv1alpha1 "github.com/kanisterio/kanister/pkg/apis/cr/v1alpha1"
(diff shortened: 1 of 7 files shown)


pkg/restic/restic.go
```diff
@@ -138,7 +140,7 @@ func resticArgs(profile *param.Profile, repository, encryptionKey string) []stri
 	var cmd []string
 	switch profile.Location.Type {
 	case crv1alpha1.LocationTypeS3Compliant:
-		cmd = resticS3Args(profile, repository)
+		cmd, _ = resticS3Args(profile, repository)
```
hakanmemisoglu: Alternatives:
1) Check error and log it.
2) Check error and propagate it to the caller methods (Need to change function signatures in this package).

tdmanv: let's propagate.

hakanmemisoglu:
@tdmanv 
Updated the test with 2 cases.
Changed the functions in other packages accordingly.

PR #316, opened 2019-09-23
Fix common logging

julio-lopez:
## Change Overview

Send log messages to a common logger instead of creating a new logger for every message.

In the future, this will allow implementing additional functionality such as:
* Setting the log level (at the package level) for the entire process
* Setting logging destinations and options.

## Pull request type

- [x] Bugfix
- [x] Feature

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -22,7 +22,6 @@ const (
 
 type logger struct {
 	level Level
-	entry *logrus.Entry
 	ctx   context.Context
 	err   error
 }
@@ -33,21 +32,18 @@ var log = logrus.New()
 func Info() Logger {
 	return &logger{
 		level: InfoLevel,
-		entry: logrus.NewEntry(logrus.New()),
 	}
 }
 
 func Error() Logger {
 	return &logger{
 		level: ErrorLevel,
-		entry: logrus.NewEntry(logrus.New()),
 	}
 }
 
 func Debug() Logger {
 	return &logger{
 		level: DebugLevel,
-		entry: logrus.NewEntry(logrus.New()),
 	}
 }
 
@@ -85,12 +81,10 @@ func (l *logger) Print(msg string) {
 
 func (l *logger) WithContext(ctx context.Context) Logger {
 	l.ctx = ctx
-	l.entry = l.entry.WithContext(ctx)
 	return l
 }
 
 func (l *logger) WithError(err error) Logger {
 	l.err = err
-	l.entry = l.entry.WithError(err)
 	return l
 }
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -52,12 +52,12 @@ func Print(msg string) {
 	Info().Print(msg)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/log/log.go
```diff
@@ -69,27 +67,16 @@ func (l *logger) Print(msg string) {
...
 
 func (l *logger) WithContext(ctx context.Context) Logger {
 	l.ctx = ctx
-	l.entry = l.entry.WithContext(ctx)
 	return l
 }
 
 func (l *logger) WithError(err error) Logger {
 	l.err = err
-	l.entry = l.entry.WithError(err)
```
SupriyaKasten: @julio-lopez how are we planning to propagate the error from here?

julio-lopez: Good catch. Let me fix it and add tests.

julio-lopez: @SupriyaKasten Addressed the issue and started adding a minimal set of tests. This particular issue is not covered by the current set of tests. I'll add more. Feel free to block the PR until enough test coverage is included.

SupriyaKasten: Thanks @julio-lopez, will block it, so that mergify does not merges it automatically before it is ready.

julio-lopez: @SupriyaKasten Expanded the tests, it should cover the basic functionality that is currently implemented. PTAL and LMK if I missed anything. Thanks.

julio-lopez:
@SupriyaKasten Expanded test coverage. PTAL.

pkg/log/log.go
```diff
@@ -69,27 +67,19 @@ func (l *logger) Print(msg string) {
...
+	entry := log.WithFields(logFields)
+	if l.err != nil {
+		entry = entry.WithError(l.err)
 	}
+	entry.Log(logrus.Level(l.level), msg)
 }
 
 func (l *logger) WithContext(ctx context.Context) Logger {
 	l.ctx = ctx
-	l.entry = l.entry.WithContext(ctx)
```
SupriyaKasten: @julio-lopez What do you think about setting the logrus.entry.Context using: `entry =entry.WithContext(l.ctx)` in the func `(l *logger) Print(msg string){}` ? We do extract the context fields though.

SupriyaKasten:
PR looks great! Thanks @julio-lopez  🚀

pkg/log/log.go
julio-lopez: The issues is that it would require keeping the entry around, instead of the context.
We may want to change the approach in the future and set the context in the entry, as you suggest, and then get the fields from the context at a much later stage, in the "hooks" or the "formatters". 
There is no need to set the context in the entry and keep it around, given the current approach where the fields are extracted and added to the entry before calling `Log(level, ...)`, as you point out. The fields need to be explicitly extracted, since `logrus` does not do anything special with the context itself. In particular, there is no context specific information in the log messages sent by default to stderr. This is because there is no general way to discover what values have been set in a context.

julio-lopez:
Thanks @SupriyaKasten

Issue #329, opened 2019-10-01
Sample blueprint to snapshot/restore AWS RDS database instances


vkamra:
We should implement a sample blueprint that leverages the AWS API to snapshot/restore AWS RDS databases.

K8s applications that use a managed service such as RDS will typically create a ConfigMap with the relevant configuration (e.g. see https://github.com/kastenhq/pgtest).

This task tracks implementing a blueprint that uses the AWS snapshot/restore functionality would demonstrate how users can leverage Kanister with such applications.

PrasadG193:
@tdmanv can we close this?

tdmanv:
Thanks @PrasadG193 !

PR #334, opened 2019-10-07
Add example for aws rds postgres integration with Kanister

PrasadG193:
## Change Overview
This PR adds an example to demonstrate Kanister integration with AWS RDS PostgreSQL engine

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [ ] Feature
- [x] Documentation

## Issues

- #329 
## Test Plan

- [x] Manual
- [ ] Unit test
- [ ] E2E


examples/aws-rds/postgresql/README.md
```diff
@@ -0,0 +1,211 @@
...
+    --allocated-storage 20 --db-instance-class db.t2.micro	 \
+    --db-instance-identifier test-postgresql-instance \
+    --engine postgres \
+    --master-username master \
+    --vpc-security-group-id sg-xxxxyyyyzzz \ # Sec group with TCP 5432 inbound rule
+    --master-user-password secret99
+
+aws rds wait db-instance-available --db-instance-identifier=test-postgresql-instance
+```
+
```
vkamra: @PrasadG193 @tdmanv - I'm wondering if we can simplify this top level README. The only requirement for someone to test this out should be:
1) Create a new RDS instance or have instanceID/host info available for an existing instance
2) Create a ConfigMap with this info
3) Create a Blueprint
4) Run backup/restore commands

The pgtest app is useful for testing but complicates this example.

PrasadG193: Won't that confuse users? They may not get the context behind why do they need configmap and which configmap to use/update. 
In current PR we can add a note suggesting instead of pgtest app, they can use their own application

vkamra: Users in this example likely already have an application that is using an RDS DB and pgtest on the top level will be confusing.

What will be most useful for them will be to know how they can use this example blueprint with their DB instance. 

I think the pgtest example is useful but perhaps this detailed README.md is in a sub-folder and the top level is very minimal

PrasadG193: @vkamra @tdmanv I have updated the README as per your suggestions.

PR #339, opened 2019-10-08
VolumeGet implementation for VMware FCD

hakanmemisoglu:
## Change Overview

VolumeGet implementation for FCD.

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/pkg/blockstorage/vmware/conversion.go
+++ b/pkg/blockstorage/vmware/conversion.go
@@ -1,0 +1,33 @@
+package vmware
+
+import (
+	"github.com/kanisterio/kanister/pkg/blockstorage"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func convertFromObjectToVolume(vso *types.VStorageObject) *blockstorage.Volume {
+	return &blockstorage.Volume{
+		Type:         blockstorage.TypeFCD,
+		ID:           vso.Config.Id.Id,
+		CreationTime: blockstorage.TimeStamp(vso.Config.CreateTime),
+		Size:         vso.Config.CapacityInMB / 1024,
+		Az:           "",
+		Iops:         0,
+		Encrypted:    false,
+	}
+}
+
+// ID wraps ID string with vim25.ID struct.
+func vimID(id string) types.ID {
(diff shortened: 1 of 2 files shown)


pkg/blockstorage/vmware/vmware.go
```diff
@@ -72,7 +72,18 @@ func (p *fcdProvider) VolumeDelete(ctx context.Context, volume *blockstorage.Vol
...
-	return nil, errors.New("Not implemented")
+	obj, err := p.gom.Retrieve(ctx, vimID(id))
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to query the disk")
+	}
+	kvs, err := p.gom.RetrieveMetadata(ctx, vimID(id), nil, "")
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to get volume metadata")
+	}
+	tags := convertKeyValueToTags(kvs)
```
tdmanv: you could also have this return kv directly.

pkg/blockstorage/vmware/conversion.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/kanisterio/kanister/pkg/blockstorage"
+)
+
+func convertFromObjectToVolume(vso *types.VStorageObject) *blockstorage.Volume {
+	return &blockstorage.Volume{
+		Type:         blockstorage.TypeFCD,
+		ID:           vso.Config.Id.Id,
+		CreationTime: blockstorage.TimeStamp(vso.Config.CreateTime),
+		Size:         vso.Config.CapacityInMB / 1024,
```
tdmanv: do we need any nil checks? will config be non nil?

tdmanv:
I added the DNM tag incase you wanted to address the feedback first

pkg/blockstorage/vmware/conversion.go
hakanmemisoglu: They are not pointers, they must be safe.

PR #344, opened 2019-10-09
SnapshotCreate and SnapshotGet for FCD

hakanmemisoglu:
## Change Overview

This PR adds SnapshotCreate and SnapshotGet for VMware FCDs.

## Pull request type

- [ ] Work in Progress
- [ ] Trivial/Minor
- [ ] Bugfix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/pkg/blockstorage/vmware/conversion.go
+++ b/pkg/blockstorage/vmware/conversion.go
@@ -3,6 +3,7 @@ package vmware
 import (
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/vmware/govmomi/vim25/types"
 
 	"github.com/kanisterio/kanister/pkg/blockstorage"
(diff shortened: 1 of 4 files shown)


pkg/blockstorage/vmware/conversion.go
```diff
@@ -21,13 +24,39 @@ func convertFromObjectToVolume(vso *types.VStorageObject) *blockstorage.Volume {
 	}
 }
 
+func convertFromObjectToSnapshot(vso *types.VStorageObjectSnapshotInfoVStorageObjectSnapshot, volID string) *blockstorage.Snapshot {
+	return &blockstorage.Snapshot{
```
tdmanv: can vso be nil?

pkg/blockstorage/vmware/vmware.go
```diff
@@ -16,6 +17,11 @@ import (
 
 var _ blockstorage.Provider = (*fcdProvider)(nil)
 
+const (
+	noDescription   = ""
+	defaultWaitTime = 10 * time.Minute
```
tdmanv: is there any guidance on how long this might take? 10 Minutes seems like a low timeout. I'd say let's increase this. Do you know if task.Wait honors context cancellation? if so we can effectively change the timeout in our tests vs prod.

hakanmemisoglu: @tdmanv I don't have definite answers for your questions. 
I am not sure, `task.Wait` honors the context cancellation.
However, in my manual testing, these were fast operations, they were completed in seconds.

tdmanv: 10 minutes works for me for now

PR #348, opened 2019-10-10
Add kopia/kopia:alpine-adf7454 to tools image

julio-lopez:
## Change Overview

Add kopia binary to tools image

## Pull request type

- [x] Feature


julio-lopez:
Should we merge this?

PR #412, opened 2019-11-15
Bugfix: use getConfig() to get correct creds for blockstorage provider

SupriyaKasten:
## Change Overview

Bugfix: use getConfig() to get correct creds for blockstorage provider when calling `provider.SnapshotCreateWaitForCompletion()` func

Without this change we did not check if `role` was set in param.Profile. With this PR, `getConfig()` will take care of providing correct configuration for the blockstorage provider. 

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/function/wait_for_snapshot_completion.go
+++ b/pkg/function/wait_for_snapshot_completion.go
@@ -65,22 +65,16 @@ func waitForSnapshotsCompletion(ctx context.Context, snapshotinfo string, profil
 	if err != nil {
 		return errors.Wrapf(err, "Could not decode JSON data")
 	}
+
 	for _, pvcInfo := range PVCData {
-		config := make(map[string]string)
 		if err = ValidateLocationForBlockstorage(profile, pvcInfo.Type); err != nil {
 			return errors.Wrap(err, "Profile validation failed")
 		}
-		switch pvcInfo.Type {
-		case blockstorage.TypeEBS:
+		config := getConfig(profile, pvcInfo.Type)
+		if pvcInfo.Type == blockstorage.TypeEBS {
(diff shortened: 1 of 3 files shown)


pkg/function/wait_for_snapshot_completion.go
```diff
@@ -65,22 +65,16 @@ func waitForSnapshotsCompletion(ctx context.Context, snapshotinfo string, profil
...
 	}
+
 	for _, pvcInfo := range PVCData {
-		config := make(map[string]string)
 		if err = ValidateLocationForBlockstorage(profile, pvcInfo.Type); err != nil {
 			return errors.Wrap(err, "Profile validation failed")
 		}
-		switch pvcInfo.Type {
-		case blockstorage.TypeEBS:
+		config := getConfig(profile, pvcInfo.Type)
```
pavannd1: `getConfig` is not handling unsupported storage provider. Will something else handle `default` case?

SupriyaKasten: Oh I assumed `ValidateLocationForBlockstorage ` func takes care of that. Let me update the PR.

pavannd1: That also doesn't. It was called `ValidateProfile` before I renamed it. Do you think we should handle the check there?

SupriyaKasten: I think so, since it will fail at validation stage, I just added a commit to this PR for the change.

PR #416, opened 2019-11-18
integration test for mysql app

viveksinghggits:
## Change Overview
After this change the integration test suite includes test for mysql app as well. We are installing single instance `stable/mysql` and then doing the tests to backup and restore the data.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues

- #XXX

## Test Plan
Ran the integration test for mysql and checked if the app is being installed, reset, backed up and then restored successfully.
Test logs are uploaded [here](https://gist.github.com/viveksinghggits/76aa2e9223163c7ead088cfe7639968e) 
- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/app/mysql.go
+++ b/pkg/app/mysql.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	mysqlWaitTimeout = 5 * time.Minute
+	mysqlWaitTimeout = 1 * time.Minute
 )
 
 type MysqlDB struct {
--- a/pkg/app/mysql.go
+++ b/pkg/app/mysql.go
@@ -1,0 +1,210 @@
+// Copyright 2019 The Kanister Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"context"
+	"strconv"
+	"strings"
+	"time"
+
+	crv1alpha1 "github.com/kanisterio/kanister/pkg/apis/cr/v1alpha1"
+	"github.com/kanisterio/kanister/pkg/field"
+	"github.com/kanisterio/kanister/pkg/helm"
+	"github.com/kanisterio/kanister/pkg/kube"
+	"github.com/kanisterio/kanister/pkg/log"
+	"github.com/pkg/errors"
+	"k8s.io/client-go/kubernetes"
+)
+
+type MysqlDB struct {
+	cli       kubernetes.Interface
+	namespace string
+	name      string
+	chart     helm.ChartInfo
+}
+
+func NewMysqlDB(name string) App {
+	return &MysqlDB{
+		name: name,
+		chart: helm.ChartInfo{
+			Release:  name,
+			RepoURL:  helm.StableRepoURL,
+			Chart:    "mysql",
+			RepoName: helm.StableRepoName,
+			Version:  "1.4.0",
+			Values: map[string]string{
+				"mysqlRootPassword":   "mysecretpassword",
+				"persistence.enabled": "false",
+				"imagePullPolicy":     "Always",
+			},
+		},
+	}
+}
+
+func (mdb *MysqlDB) Init(ctx context.Context) error {
+
+	cfg, err := kube.LoadConfig()
+	if err != nil {
+		return err
+	}
+
+	mdb.cli, err = kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (mdb *MysqlDB) Install(ctx context.Context, namespace string) error {
+
+	mdb.namespace = namespace
+
+	cli := helm.NewCliClient("helmv3")
+	log.Print("Adding repo.", field.M{"app": mdb.name})
+	err := cli.AddRepo(ctx, mdb.chart.RepoName, mdb.chart.RepoURL)
+	if err != nil {
+		return errors.Wrapf(err, "Error helm repo for app %s.", mdb.name)
+	}
+
+	log.Print("Installing mysql instance using helm.", field.M{"app": mdb.name})
+	err = cli.Install(ctx, mdb.chart.RepoName+"/"+mdb.chart.Chart, mdb.chart.Version, mdb.chart.Release, mdb.namespace, mdb.chart.Values)
+	if err != nil {
+		return errors.Wrapf(err, "Error intalling application %s through helm.", mdb.name)
+	}
+
+	return nil
+}
+
+func (mdb *MysqlDB) IsReady(ctx context.Context) (bool, error) {
+	log.Print("Waiting for the mysql instance to be ready.", field.M{"app": "mysql"})
+	ctx, cancel := context.WithTimeout(ctx, time.Minute*5)
+	defer cancel()
+	err := kube.WaitOnDeploymentReady(ctx, mdb.cli, mdb.namespace, mdb.name)
+	if err != nil {
+		return false, err
+	}
+
+	log.Print("Application instance is ready.", field.M{"app": mdb.name})
+	return true, nil
+
+}
+
+func (mdb *MysqlDB) Object() crv1alpha1.ObjectReference {
+
+	return crv1alpha1.ObjectReference{
+		Kind:      "deployment",
+		Name:      mdb.name,
+		Namespace: mdb.namespace,
+	}
+}
+
+func (mdb *MysqlDB) Uninstall(ctx context.Context) error {
+	cli := helm.NewCliClient("helmv3")
+
+	err := cli.Uninstall(ctx, mdb.name, mdb.namespace)
+	if err != nil {
+		log.WithError(err).Print("Failed to uninstall app, you will have to uninstall it manually.", field.M{"app": mdb.name})
+		return err
+	}
+	log.Print("Uninstalled application.", field.M{"app": mdb.name})
+
+	return nil
+}
+
+func (mdb *MysqlDB) Ping(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


pkg/app/mysql.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func (mdb *MysqlDB) Count(context.Context) (int, error) {
+	podname, containername, err := mdb.GetPodAndContainerName(mdb.namespace)
+	selectRowsCMD := []string{"sh", "-c", "mysql -u root --password=" + mdb.password + " -e 'use testdb; select count(*) from pets; '"}
+	stdout, stderr, err := kube.Exec(mdb.cli, mdb.namespace, podname, containername, selectRowsCMD, nil)
+	format.Log(podname, containername, stdout)
+	format.Log(podname, containername, stderr)
+	if err != nil {
+		log.Fatalf("Error while selecting rows from table %s:", err.Error())
```
tdmanv: Please return an error here

pkg/app/mysql.go
```diff
@@ -0,0 +1,244 @@
...
+			helmrepoURL:   helmrepoURL,
+			helmChartName: helmChart,
+			helmAppname:   helmAppname,
+		},
+	}
+}
+
+func (mdb *MysqlDB) Init(ctx context.Context) error {
+	var ok bool
+	mdb.region, ok = os.LookupEnv(awsconfig.Region)
```
tdmanv: Do these need to be set in the mysql struct? Can they instead be handled by the profile?

pkg/app/mysql.go
```diff
@@ -0,0 +1,244 @@
...
+
+	return nil
+}
+
+func (mdb *MysqlDB) Ping(context.Context) error {
+
+	podname, containername, err := mdb.GetPodAndContainerName(mdb.namespace)
+
+	// exec into the pod and create the test database, read password from secret
+	loginMysql := []string{"sh", "-c", "mysql -u root --password=" + mdb.password + ""}
```
tdmanv: Is it possible to use database/SQL and the fo driver?

pkg/app/mysql.go
```diff
@@ -0,0 +1,244 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	crv1alpha1 "github.com/kanisterio/kanister/pkg/apis/cr/v1alpha1"
+	awsconfig "github.com/kanisterio/kanister/pkg/config/aws"
+	"github.com/kanisterio/kanister/pkg/format"
+	"github.com/kanisterio/kanister/pkg/kube"
+	"github.com/kanisterio/kanister/pkg/testutil"
+	log "github.com/sirupsen/logrus"
```
tdmanv: Please use our logging library pkg/log

pkg/app/mysql.go
viveksinghggits: Hi @tdmanv ,
I think to use database/SQL pkg and the driver we will have to expose the mysql deployment either using service (NodePort) or ingress I think.
What do you suggest ?

pkg/app/mysql.go
viveksinghggits: Hi @tdmanv 
I agree, and it was duplicated here in mysql struct. These values are closely related to the profile and should be set there only.
I will make the respective changes.

pkg/app/mysql.go
```diff
@@ -0,0 +1,227 @@
...
+			Kind:      "Secret",
+			Name:      "mysql-test",
+			Namespace: mdb.namespace,
+		},
+	}
+}
+
+// GetPodAndContianerName takes namespace as input and returns the pod and container that is running in
+// that namespace for deployment that was created through helm
+func (mdb *MysqlDB) GetPodAndContainerName(deploymentNamespace string) (string, string, error) {
```
PrasadG193: @viveksinghggits We should probably add new helper functions like -
```
StatefulSetContainers(ctx context.Context, kubeCli kubernetes.Interface, namespace string, name string) []string
DeploymentContainers(ctx context.Context, kubeCli kubernetes.Interface, namespace string, name string) []string
```
 in `pkg/kube/workload.go` to return containers from Statefulset or Deployment. Since we will be needing this for other apps also

@tdmanv @SupriyaKasten Any thoughts?

PrasadG193:
@viveksinghggits Please rebase your branch with master. We have moved helm client from `pkg/testutil` to `pkg/helm` (https://github.com/kanisterio/kanister/pull/415). Please make the changes accordingly

pkg/app/mysql.go
```diff
@@ -0,0 +1,227 @@
...
+
+	return crv1alpha1.ObjectReference{
+		Kind:      "deployment",
+		Name:      "mysql-test",
+		Namespace: mdb.namespace,
+	}
+}
+
+func (mdb *MysqlDB) Uninstall(ctx context.Context) error {
+	log.Print("MysqlDB: UnInstalling mysql instance.")
```
PrasadG193: @viveksinghggits let's use the following format for logging accross all the apps 
`log.Info/Debug/Error().Print("msg", field.M{"app": app.name, additional_fields... })`
Let me know what you think.

pkg/app/mysql.go
```diff
@@ -0,0 +1,227 @@
...
+
+	return nil
+}
+
+func (mdb *MysqlDB) Install(ctx context.Context, namespace string) error {
+
+	mdb.namespace = namespace
+
+	cli := testutil.NewHelmClient()
+	log.Print("Adding repo.")
```
PrasadG193: @viveksinghggits let's use the following format for logging accross all the apps 
`log.Info/Debug/Error().Print("msg", field.M{"app": app.name, additional_fields... })`
Let me know what you think.

viveksinghggits:
Hi @PrasadG193 ,
I have rebased the branch and pushed the updated code.

pkg/app/mysql.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		log.WithError(err).Print("Error getting deployment to ping.", field.M{"app": "mysql"})
+	}
+	deploymentLabelSelector := deployment.Spec.Selector.MatchLabels
+	var podLabelSelector []string
+	for key, value := range deploymentLabelSelector {
+		podLabelSelector = append(podLabelSelector, key+"="+value)
+	}
+
+	pods, err := mdb.cli.CoreV1().Pods(mdb.namespace).List(metav1.ListOptions{
```
PrasadG193: Please use `kube.FetchPods()` here (https://github.com/kanisterio/kanister/blob/master/pkg/kube/workload.go#L172)

PrasadG193: Let's raise separate PR for this along with changes mentioned in https://github.com/kanisterio/kanister/pull/416/files#r347747536

pkg/app/mysql.go
viveksinghggits: Hi @tdmanv 
as discussed lets go ahead with execing into the pod to talk to the application.

pkg/app/mysql.go
viveksinghggits: Hi @PrasadG193 ,
I have refactored the code to use the new methods from `kube` package.

pkg/app/mysql.go
```diff
@@ -0,0 +1,244 @@
...
+		password: "mysecretpassword",
+		chart: chartInfo{
+			helmrepoURL:   helmrepoURL,
+			helmChartName: helmChart,
+			helmAppname:   helmAppname,
+		},
+	}
+}
+
+func (mdb *MysqlDB) Init(ctx context.Context) error {
```
PrasadG193: @viveksinghggits We should check if required AWS creds are set in Init(). If not return error to skip the test

pkg/testing/integration_test.go
```diff
@@ -63,6 +63,14 @@ var _ = Suite(&IntegrationSuite{
 	profile:   newSecretProfile("", "", ""),
 })
 
+var _ = Suite(&IntegrationSuite{
+	name:      "mysql",
+	namespace: "mysql",
+	app:       app.NewMysqlDB("https://kubernetes-charts.storage.googleapis.com/", "stable", "mysql"),
+	bp:        app.NewBlueprint("mysql"),
+	profile:   newSecretProfile("infracloud.kanister.io", "", ""),
```
PrasadG193: @tdmanv We probably want to use different bucket for this?

pkg/app/mysql.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	return nil
+}
+
+func (mdb *MysqlDB) Install(ctx context.Context, namespace string) error {
+
+	mdb.namespace = namespace
+
+	cli := helm.NewCliClient(helmVersion)
```
PrasadG193: Please use `helm.V3` instead

pkg/app/mysql.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return crv1alpha1.ObjectReference{
+		Kind:      "deployment",
+		Name:      mdb.name,
+		Namespace: mdb.namespace,
+	}
+}
+
+func (mdb *MysqlDB) Uninstall(ctx context.Context) error {
+	cli := helm.NewCliClient(helmVersion)
```
PrasadG193: same changes here.

pkg/testing/integration_test.go
```diff
@@ -157,7 +166,18 @@ func (s *IntegrationSuite) TestRun(c *C) {
 	testEntries := 3
 	// Add test entries to DB
 	if a, ok := s.app.(app.DatabaseApp); ok {
-		err = a.Ping(ctx)
+		// wait for application to be actually ready
+		timeoutCtx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
PrasadG193: Let's declare timeout value as a constant in headers and use everywhere

pkg/testing/integration_test.go
```diff
@@ -225,7 +250,18 @@ func (s *IntegrationSuite) TestRun(c *C) {
...
+		defer waitCancel()
+		err := poll.Wait(timeoutCtx, func(ctx context.Context) (bool, error) {
+			err := a.Ping(ctx)
+			if err != nil {
+				return false, nil
+			} else {
+				return true, nil
+			}
+		})
+
```
PrasadG193: @viveksinghggits Can we move this snippet to a function to avoid code duplication?

PR #424, opened 2019-11-21
add mongodb app in integration test

viveksinghggits:
## Change Overview

- This change adds mongodb apps integration into our testing framework. There are some intermittent issues, that I will try and fix as soon as possible. 

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan
Ran go test command to test this change.
```
go test -v ./pkg/testing -tags=integration -timeout 30m -count=1
```

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/app/mongodb.go
+++ b/pkg/app/mongodb.go
@@ -17,6 +17,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strconv"
 	"time"
 
@@ -46,9 +47,10 @@ func NewMongoDB(name string) App {
 		name:     name,
 		chart: helm.ChartInfo{
 			Release:  name,
-			RepoUrl:  helm.StableRepoURL,
+			RepoURL:  helm.StableRepoURL,
 			Chart:    "mongodb",
 			RepoName: helm.StableRepoName,
+			Version:  "7.4.6",
 			Values: map[string]string{
 				"replicaSet.enabled":  "true",
 				"image.repository":    "kanisterio/mongodb",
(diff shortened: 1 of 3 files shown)


pkg/app/mongodb.go
```diff
@@ -0,0 +1,220 @@
...
+		Namespace: mongo.namespace,
+	}
+}
+
+func (mongo *MongoDB) Uninstall(ctx context.Context) error {
+	cli := helm.NewCliClient(helm.V3)
+
+	log.Print("Uninstalling application.", field.M{"app": mongo.name})
+	err := cli.Uninstall(ctx, mongo.name, mongo.namespace)
+	if err != nil {
```
PrasadG193: We can just return `errors.Wrapf(err)`here. If error is nil errors.Wrapf() also return nil

pkg/app/mongodb.go
```diff
@@ -0,0 +1,220 @@
...
+		return errors.Wrapf(err, "Error %s checking if the pod is master.", stderr)
+	}
+
+	// extract the ismaster field from the output and make sure
+	// this ismaster field is true
+	re := regexp.MustCompile("\"ismaster\" : ([a-z]*),")
+	match := re.FindStringSubmatch(isMasterStdout)
+
+	isMaster, err := strconv.ParseBool(match[1])
```
PrasadG193: We should ensure `len(match) >= 2` before accessing the element

pkg/app/mongodb.go
```diff
@@ -0,0 +1,220 @@
...
+}
+func (mongo *MongoDB) Reset(ctx context.Context) error {
+	log.Print("Resetting the application.", field.M{"app": mongo.name})
+	// delete all the entries from the restaurants dummy colleciton
+	// we are not deleting the database becasue we are dealing with admin database here
+	// and deletion admin database is prohibited
+	deleteDBCMD := []string{"sh", "-c", fmt.Sprintf("mongo admin --authenticationDatabase admin -u %s -p $MONGODB_ROOT_PASSWORD --quiet --eval \"rs.slaveOk(); db.restaurants.drop()\"", mongo.username)}
+	stdout, stderr, err := mongo.execCommand(ctx, deleteDBCMD)
+	if err != nil {
+		return errors.Wrapf(err, "Error %s, resetting the mongodb application. stdout is %s", stderr, stdout)
```
PrasadG193: No need to check for err here. We can just return `errors.Wrapf`

pkg/app/mongodb.go
viveksinghggits: yeah, I understand, but if there is an error I wanted to print when did we encounter the error. Do you think `errors.Wrapf(err)` is better idea.

pkg/app/mongodb.go
viveksinghggits: So, output of `db.ismaster` is always going to have `ismaster` field in output so we are always going to have `len(match)>= 2`. But yeah, I think we should that.

pkg/app/mongodb.go
PrasadG193: We are anyways returning `errors.Wrapf(err, "Error while uninstalling the application.")` on line no 134. What I meant is we don't need to check `if err != nil` if we are using `errors.Wrapf()` function

pkg/app/mongodb.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/kanisterio/kanister/pkg/helm"
+	"github.com/kanisterio/kanister/pkg/kube"
+	"github.com/kanisterio/kanister/pkg/log"
+)
+
+type MongoDB struct {
+	cli       kubernetes.Interface
+	namespace string
+	username  string
+	password  string
```
PrasadG193: @viveksinghggits are we using this field?

PR #479, opened 2020-01-03
Add aws error fields and their values to log

SupriyaKasten:
## Change Overview

Add aws error fields and their values to log

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Logs
- [ ] :robot: Test

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"os"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -37,6 +38,8 @@ const (
 	LoggingServiceHostEnv = "LOGGING_SVC_SERVICE_HOST"
 	LoggingServicePortEnv = "LOGGING_SVC_SERVICE_PORT_LOGGING"
 )
+
+const errorFieldName = "error"
 
 type logger struct {
 	level Level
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -117,7 +120,24 @@ func (l *logger) Print(msg string, fields ...field.M) {
 
 	entry := log.WithFields(logFields)
 	if l.err != nil {
-		entry = entry.WithError(l.err)
+		switch e := l.err.(type) {
```
ewhamilton: It appears that the effect is to change AWS errors into fields and never set the error in the entry for those.  Is that the intent?

It would be more efficient and I think easier to read if adding the AWS fields was done slightly above here as part of filling in logFields before line 121.  A check of `if e, ok := l.err.(awserr.Error); ok {` checks for nil and the type match (I think).

pkg/log/log.go
```diff
@@ -117,7 +120,24 @@ func (l *logger) Print(msg string, fields ...field.M) {
...
+		case awserr.Error:
+			errFields := make(logrus.Fields)
+			errFields["awsErrorCode"] = e.Code()
+			errFields["awsErrorMessage"] = e.Message()
+			if er, ok := e.(awserr.RequestFailure); ok {
+				errFields["awsRequestStatusCode"] = er.StatusCode()
+				errFields["awsRequestID"] = er.RequestID()
+			}
+			if nextErr := e.OrigErr(); nextErr != nil {
```
ewhamilton: What is this doing with OrigErr and the "error"  Field?

SupriyaKasten: `OrigErr` returns the original error if one was set.  Nil is returned if not set.
Since we are not calling entry.WithError() on the awserr which sets the "error" field to the error passed as parameter, we are setting it. this will assign the field `error` with value `nextErr` in the error log.

entry.WithError(err) internally calls entry.WithField(ErrorKey, err). in the case of awserrs we are extracting the fields and setting them.

pkg/log/log.go
SupriyaKasten: To answer your first question: here in case of awserr, we are doing the work for entry.WithError(). We are extracting the well know fields and assigning them to the log to make it more descriptive and clear. Otherwise, entry.WithError(err) internally calls entry.WithFields(ErrorKey, err).


About moving the aws fields before line 121, i really liked the suggestion, let me test with nil value and push something.

SupriyaKasten:
@ewhamilton please review the changes in the PR that you suggested, when you get time. Thanks!

PR #481, opened 2020-01-03
Add init func to set env variable fields and their values into logs

SupriyaKasten:
## Change Overview

Add init func to set env variable fields and their values into logs

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: log
- [ ] :robot: Test

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -103,6 +125,12 @@ func WithError(err error) Logger {
 
 func (l *logger) Print(msg string, fields ...field.M) {
 	logFields := make(logrus.Fields)
+
+	envFields := envVarFields.Fields()
```
ewhamilton: Suggest  combining line 129 and line 130 into ```
for _, f := range envVarFields.Fields() {
```

PR #482, opened 2020-01-06
Add stack Trace to error logs

SupriyaKasten:
## Change Overview

Add stackTrace to error logs

## Pull request type

- [x] :sunflower: Log

Sample output:
```
time="2020-01-08T17:17:04.815055486Z" level=info msg="Failed to execute phase: v1alpha1.Phase{Name:\"myPhase-0\", State:\"pending\", Output:map[string]interface {}(nil)}:" ActionSet=test-actionset-4t89d Phase=myPhase-0 error="Kanister function failed" stackTrace="\ngithub.com/kanisterio/kanister/pkg/testutil.failFunc\n\t/go/src/github.com/kanisterio/kanister/pkg/testutil/func.go:45\ngithub.com/kanisterio/kanister/pkg/testutil.(*mockKanisterFunc).Exec\n\t/go/src/github.com/kanisterio/kanister/pkg/testutil/func.go:106\ngithub.com/kanisterio/kanister/pkg.(*Phase).Exec\n\t/go/src/github.com/kanisterio/kanister/pkg/phase.go:72\ngithub.com/kanisterio/kanister/pkg/controller.(*Controller).runAction.func1\n\t/go/src/github.com/kanisterio/kanister/pkg/controller/controller.go:406\ngopkg.in/tomb%2ev2.(*Tomb).run\n\t/go/pkg/mod/gopkg.in/tomb.v2@v2.0.0-20161208151619-d5d1b5820637/tomb.go:163\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1337"
```

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -6,9 +6,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"time"
 
-	"github.com/luci/go-render/render"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -42,6 +40,8 @@ const (
 	LoggingServicePortEnv = "LOGGING_SVC_SERVICE_PORT_LOGGING"
 )
 
+const stackTrace = "stackTrace"
+
 type logger struct {
 	level Level
 	ctx   context.Context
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -70,6 +74,16 @@ func SetOutput(sink OutputSink) error {
 	}
 }
 
+func addStackTraceField(e *logrus.Entry, err error) *logrus.Entry {
+	if e != nil && err != nil {
+		split := strings.SplitAfter(fmt.Sprintf("%+v", err), err.Error())
```
ewhamilton: This method of logging the stack trace presumes that the error was created with `github.com/pkg/errors.New(()` or wrapped with `github.com/pkg/errors.Wrap()`.  This will not be true for kio/errors.  That's likely OK for now.  

Given that dependence though, a better way to check for the stacktrace is to use a typecast to errors.stackTracer interface (Apparently that works even if the interface type is not exported: https://godoc.org/github.com/pkg/errors).  Otherwise use a local definition of an interace and check for presence of StackTrace() method.  It's still OK to use the fmt.Sprintf("%+v") since the errors docs document it.

SupriyaKasten: So, the reason why I took this approach was because I tested it with both error created by kio/errors and github.com/pkg/errors. 
For kio/error:  fmt.Sprintf("%+v", err) EQUALS err.Error() and there is no stack trace. So len(split) will always be 1 for this case.
For github.com/pkg/errors: there is no guarantee that we get a stackTrace for each error, so  len(split) > 1  only when we have a stackTrace.

PR #520, opened 2020-01-21
Include MySQL DeploymentConfig app in integration test suite

viveksinghggits:
## Change Overview
This PR adds MySQL that is deployed on OpenShift cluster using DeploymentConfig resource into our integration test suite.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

Since we cant deploy OpenShift on Containers we can test this app locally only as of now. Make the changes in the Makefile to not run in docker image, and include `MysqlDBDepConfig` in `SHORT_APPS` in `integration-test.sh`, after that just run 

```
make integration-test
```

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/app/mysql-deploymentconfig.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Error installing app on openshift cluster.", field.M{"app": mdep.name, "out": out})
+	}
+
+	err = mdep.createMySQLToolsPod(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "Error creating mysql tools pod")
+	}
+
+	err = mdep.createMySQLSecret(ctx)
```
PrasadG193: `return mdep.createMySQLSecret(ctx)`

pkg/app/mysql-deploymentconfig.go
```diff
@@ -0,0 +1,275 @@
...
+							Name:  "MYSQL_ROOT_PASSWORD",
+							Value: mdep.envVar["MYSQL_ROOT_PASSWORD"],
+						},
+					},
+				},
+			},
+		},
+	}
+
+	_, err := mdep.cli.CoreV1().Pods(mdep.namespace).Create(mysqlToolPod)
```
PrasadG193: `return err`

pkg/app/mysql-deploymentconfig.go
```diff
@@ -0,0 +1,275 @@
...
+			Name:      mdep.name,
+			Namespace: mdep.namespace,
+		},
+		Data: map[string][]byte{
+			"mysql-root-password": []byte(mdep.envVar["MYSQL_ROOT_PASSWORD"]),
+		},
+	}
+
+	_, err := mdep.cli.CoreV1().Secrets(mdep.namespace).Create(mysqlSecret)
+	if err != nil {
```
PrasadG193: No need to check if err is nil. Same at other places

viveksinghggits: I wanted to wrap the error detail with the actual message, do you think removing that is better option?

PR #543, opened 2020-01-29
Handle Namespace/ServiceAccount defaults in PodRunner

tdmanv:
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod_test.go
+++ b/pkg/kube/pod_test.go
@@ -23,7 +23,7 @@ import (
 	"time"
 
 	. "gopkg.in/check.v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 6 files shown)


pkg/kube/pod_test.go
```diff
@@ -62,17 +62,76 @@ func (s *PodSuite) TearDownSuite(c *C) {
...
+			Image:              "kanisterio/kanister-tools:0.28.0",
+			Command:            []string{"sh", "-c", "tail -f /dev/null"},
+			ServiceAccountName: "dummy-sa",
+		},
+		{
+			Namespace:    cns,
+			GenerateName: "test-",
+			Image:        "kanisterio/kanister-tools:0.28.0",
+			Command:      []string{"sh", "-c", "tail -f /dev/null"},
+		},
```
PrasadG193: Should add one more testcase, where - `Namespace: cns` and `ServiceAccountName: "non-default"`.

PrasadG193:
Overall looks good. Please fix CI failure and suggested test case.

viveksinghggits:
After increasing the timeout the test passes, we can use command 
```
make test
```
to run the tests and the logs can be found [here](https://gist.github.com/viveksinghggits/a59ad8d5e85a18ae5f772bb649720d17).

pkg/kube/pod_test.go
```diff
@@ -62,17 +77,80 @@ func (s *PodSuite) TearDownSuite(c *C) {
...
-	})
-	c.Assert(err, IsNil)
-	c.Assert(WaitForPodReady(ctx, s.cli, s.namespace, pod.Name), IsNil)
-	c.Assert(DeletePod(context.Background(), s.cli, pod), IsNil)
+
+	podOptions := []*PodOptions{
+		{
+			Namespace:    s.namespace,
+			GenerateName: "test-",
+			Image:        "kanisterio/kanister-tools:0.28.0",
```
tdmanv: please use constants for this image. We should have it defined somewhere

viveksinghggits: there is one in package function's util.go but 
1. that is not exported and 
2. using that is going to introduce import cycle
so I am planning to make constant in the same source file.

PR #574, opened 2020-02-21
K10 3548 Handle unresponsive nodes when choosing zones on restore

bathina2:
## Change Overview

Adding checks on available nodes to make sure they are ready and are schedulable before selecting a zone.

## Pull request type

Please check the type of change your PR introduces:
- [x] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test

## Issues

- #K10-3548, K10-3515

https://kasten.atlassian.net/browse/K10-3548
https://kasten.atlassian.net/browse/K10-3515

## Test Plan

- [ x] :muscle: Manual
- [ x] :zap: Unit test
- [ ] :green_heart: E2E

PR #580, opened 2020-02-24
K10 3548

bathina2:
## Change Overview

Sorry if this PR is long. It got bigger than I anticipated. Good things is a majority is UT's. There are 2 main parts to it.

1. Validating if the nodes returned by kubernetes are A) Schedulable and B) In the 'Ready' State.

2. Picking the correct zone to create a Volume in.  For this I followed the following approach-
a) Get the zones of a given region from the provider
b) Get the zones available to a cluster from the node objects derived from kubernetes.
c) Sanitize the kubernetes zone names to match those from the provider. If there wasn't an exact match,  I used the levenshtein algorithm found here https://godoc.org/github.com/agnivade/levenshtein to get the best match.
d) If the source zone is in the sanitized zones we return it. However if we can't find the source zone, we return the list of all sanitized zones and pick the first one.
e) If the sanitized zone list is empty we validate and use the source zones.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #574 

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,6 +172,9 @@ test: build-dirs
 integration-test: build-dirs
 	@$(MAKE) run CMD='-c "./build/integration-test.sh short"'
 
+openshift-test:
+	@/bin/bash ./build/integration-test.sh openshift
+
 golint:
 	@$(MAKE) run CMD='-c "./build/golint.sh"'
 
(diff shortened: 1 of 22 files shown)


bathina2:
Breaking this up into smaller PRs and closing

PR #581, opened 2020-02-25
Adding checks to fetch ready nodes from kubernetes

bathina2:
## Change Overview

Checking node status and schedulable fields before selecting nodes for export.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/blockstorage/zone/zone.go
```diff
@@ -28,6 +28,8 @@ import (
 	"github.com/kanisterio/kanister/pkg/kube"
 	kubevolume "github.com/kanisterio/kanister/pkg/kube/volume"
 	"github.com/kanisterio/kanister/pkg/log"
+	v1 "k8s.io/api/core/v1"
```
vkamra: Could skip `v1` alias here

vkamra: 3rd party imports should be in the second import group. Could you move these above please?

pkg/blockstorage/zone/zone.go
```diff
@@ -192,13 +194,17 @@ func NodeZonesAndRegion(ctx context.Context, cli kubernetes.Interface) (map[stri
...
 	if err != nil {
 		return nil, "", errors.Wrap(err, nodeZonesErr)
 	}
 	zoneSet := make(map[string]struct{})
 	regionSet := make(map[string]struct{})
 	for _, n := range ns.Items {
+		// For kubernetes 1.17 onwards failureDomain annotations are being deprecated
+		// and will need to use topology.kubernetes.io/zone=us-east-1c and
+		// topology.kubernetes.io/region=us-east-1
+		// https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesioregion
```
vkamra: @bathina2 - can you follow up with a PR that does uses the new annotations? The code can look at either for now to maintain compatibility with older versions

pkg/blockstorage/zone/zone.go
```diff
@@ -222,6 +228,65 @@ func NodeZonesAndRegion(ctx context.Context, cli kubernetes.Interface) (map[stri
...
+		"spec.unschedulable": "false",
+	}.AsSelector().String()})
+	if err != nil {
+		return nil, err
+	}
+	Filter(ns, func(node v1.Node) bool {
+		return IsNodeSchedulable(&node)
+	})
+	if len(ns.Items) == 0 {
+		return nil, errors.New("There are currently no ready, schedulable nodes in the cluster")
```
vkamra: It would be useful to know which nodes were filtered out and why if we are hitting this case (maybe just stats - similar to what is logged in K8s events for pods e.g. 3 found, 2 were unschedulable, 1 was not ready)

pkg/blockstorage/zone/zone.go
```diff
@@ -222,6 +222,61 @@ func NodeZonesAndRegion(ctx context.Context, cli kubernetes.Interface) (map[stri
...
+func GetReadySchedulableNodes(cli kubernetes.Interface) (*v1.NodeList, error) {
+	ns, err := cli.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	total := len(ns.Items)
+	var unschedulable, notReady int
+	var l []v1.Node
+	for _, node := range ns.Items {
+		if !IsNodeReady(&node) {
```
tdmanv: [nit] prefer boolean switch statements for if-else-else

pkg/blockstorage/zone/zone.go
```diff
@@ -222,6 +222,61 @@ func NodeZonesAndRegion(ctx context.Context, cli kubernetes.Interface) (map[stri
 	return zoneSet, region[0], nil
 }
 
+// GetReadySchedulableNodes addresses the common use case of getting nodes you can do work on.
+// 1) Needs to be schedulable.
+// 2) Needs to be ready.
+// Derived from "k8s.io/kubernetes/test/e2e/framework/node"
+// TODO: check for taints as well
+func GetReadySchedulableNodes(cli kubernetes.Interface) (*v1.NodeList, error) {
```
tdmanv: why not return a `[]v1.Node` directly?

PR #584, opened 2020-02-25
Changing implementation of FromSourceRegionZone

bathina2:
## Change Overview

Current implementation will use the new levenshtein methods to calculate best match for zone name

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #K10-3548

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/blockstorage/awsebs/awsebs.go
```diff
@@ -26,15 +26,16 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/jpillora/backoff"
-	"github.com/kanisterio/kanister/pkg/poll"
 	"github.com/pkg/errors"
```
DeepikaDixit: [nit] fix imports

pkg/blockstorage/awsebs/awsebs.go
```diff
@@ -407,13 +408,13 @@ func (s *ebsStorage) VolumeCreateFromSnapshot(ctx context.Context, snapshot bloc
 	if snapshot.Volume.VolumeType == "" || snapshot.Volume.Az == "" || snapshot.Volume.Tags == nil {
 		return nil, errors.Errorf("Required volume fields not available, volumeType: %s, Az: %s, VolumeTags: %v", snapshot.Volume.VolumeType, snapshot.Volume.Az, snapshot.Volume.Tags)
 	}
-
-	zones, err := zone.FromSourceRegionZone(ctx, s, snapshot.Region, snapshot.Volume.Az)
+	cli, _ := kube.NewClient()
```
DeepikaDixit: Is the `_` needed? If there's error, we shouldn't ignore it

pkg/blockstorage/gcepd/gcepd.go
```diff
@@ -362,7 +363,8 @@ func (s *gpdStorage) VolumeCreateFromSnapshot(ctx context.Context, snapshot bloc
 		return nil, errors.Wrapf(err, "Could not validate zones: %s", snapshot.Volume.Az)
 	}
 	zones = splitZones(snapshot.Volume.Az)
-	zones, err = zone.FromSourceRegionZone(ctx, s, snapshot.Region, zones...)
+	cli, _ := kube.NewClient()
```
DeepikaDixit: Same as before

pkg/blockstorage/awsebs/awsebs.go
bathina2: I assumed we wanted to group all the kanister imports together so I moved it to the list below?

DeepikaDixit: Oh sorry. I misread it. The imports are fine

pkg/blockstorage/awsebs/awsebs.go
bathina2: The error is not needed. If the CLI is nil we use the provider specified zones directly.

pkg/blockstorage/gcepd/gcepd.go
bathina2: same as above

pkg/blockstorage/awsebs/awsebs.go
```diff
@@ -407,13 +408,13 @@ func (s *ebsStorage) VolumeCreateFromSnapshot(ctx context.Context, snapshot bloc
...
+	zones, err := zone.FromSourceRegionZone(ctx, s, cli, snapshot.Region, snapshot.Volume.Az)
 	if err != nil {
 		return nil, err
 	}
 	if len(zones) != 1 {
-		return nil, errors.Errorf("Length of zone slice should be 1, got %d", len(zones))
+		log.Debug().Print("Length of zone slice is greater than 1", field.M{"numberOfZones": len(zones)})
```
tdmanv: What motivates this change in behavior?

pkg/blockstorage/awsebs/awsebs.go
tdmanv: Do we always expect to be able to get the CLI? I would hoist this one level higher when we construct ebsStorage and log an error if it's not used. 

This would even let us mock the cli in unit tests by setting the the field in the struct.

pkg/blockstorage/gcepd/gcepd.go
tdmanv: same as earlier

pkg/blockstorage/zone/zone.go
```diff
@@ -39,153 +36,89 @@ type (
...
-					log.WithError(err).Print("Ignoring error getting Zone from KnownNodeZones.")
-				}
-			}
-		}
+	validZoneNames, err := m.FromRegion(ctx, sourceRegion)
+	if err != nil || len(validZoneNames) == 0 {
+		return nil, errors.Wrapf(err, "No provider zones for region (%s)", sourceRegion)
 	}
+	if !isNil(kubeCli) {
+		getAvailableZones(ctx, newZones, kubeCli, validZoneNames, sourceZones, sourceRegion)
```
tdmanv: This modifies `newZones`, right? I would just return it, otherwise this takes extra effort to read.

I would also push down handling of a nil kubeCli and return an error if something goes wrong.

pkg/blockstorage/zone/zone.go
```diff
@@ -39,153 +36,89 @@ type (
...
+	}
+	// If Kubernetes provided zones are invalid use valid sourceZones
 	if len(newZones) == 0 {
+		log.Info().Print("Validating source zones")
```
tdmanv: Please reduce this logging level. I would probably delete this, but it's your call.

pkg/blockstorage/zone/zone.go
```diff
@@ -39,153 +36,89 @@ type (
...
-	if validZones, err := m.FromRegion(ctx, region); err == nil {
-		for _, z := range validZones {
-			if zone == z {
-				return true
-			}
-		}
-	}
-	return false
+func isNil(i interface{}) bool {
+	return i == nil || reflect.ValueOf(i).IsNil()
```
tdmanv: reflection can be expensive. Why use this over a typed nil comparison?

pkg/blockstorage/awsebs/awsebs.go
bathina2: The scenario where, if the original zone wasn't found in the cluster we would return a list of all available zones in the cluster.

pkg/blockstorage/zone/zone.go
bathina2: So this is taking into account that if kubecli is nil we would use the zones from the provider to compare the source zone against. If you'd rather we fail out at this point thats fine too

pkg/blockstorage/zone/zone.go
bathina2: Will remove comment

pkg/blockstorage/zone/zone.go
```diff
@@ -277,10 +210,11 @@ func isNodeReady(node *v1.Node) bool {
 	return false
 }
 
-func sanitizeAvailableZones(availableZones map[string]struct{}, validZoneNames []string) map[string]struct{} {
+// SanitizeAvailableZones validates and updates a map of zones against a list of valid zone names
+func SanitizeAvailableZones(availableZones map[string]struct{}, validZoneNames []string) map[string]struct{} {
```
tdmanv: does this need to be exported?

bathina2: There were existing tests in other packages (awsebs, gcepd etc.) that were doing some zone validation. I exported this to be able to maintain those tests

pkg/blockstorage/zone/zone.go
bathina2: Sure, seemed easier but will switch over to nil comparison

pkg/blockstorage/azure/azuredisk.go
```diff
@@ -375,12 +375,12 @@ func getRegionAndZoneID(ctx context.Context, s *adStorage, sourceRegion, volAz s
 		return region, "", nil
 	}
 
-	zones, err := zone.FromSourceRegionZone(ctx, s, sourceRegion, volAz)
+	zones, err := zone.FromSourceRegionZone(ctx, s, cli, sourceRegion, volAz)
 	if err != nil {
 		return "", "", err
 	}
 	if len(zones) != 1 {
```
tdmanv: Do we need to check if `len(zones) == 0` now that we log and continue below

bathina2: I return an error if the len(zones)==0.

tdmanv: gotcha.

pkg/blockstorage/gcepd/gcepd.go
```diff
@@ -358,11 +359,15 @@ func (s *gpdStorage) VolumeCreateFromSnapshot(ctx context.Context, snapshot bloc
 	var zones []string
 	var region string
 	// Validate Zones
-	if _, err = getRegionFromZones(snapshot.Volume.Az); err != nil {
+	if region, err = getRegionFromZones(snapshot.Volume.Az); err != nil {
 		return nil, errors.Wrapf(err, "Could not validate zones: %s", snapshot.Volume.Az)
 	}
 	zones = splitZones(snapshot.Volume.Az)
-	zones, err = zone.FromSourceRegionZone(ctx, s, snapshot.Region, zones...)
+	cli, err := kube.NewClient()
```
tdmanv: We should hoist this to the gpdStorage struct

pkg/blockstorage/awsebs/awsebs.go
tdmanv: Is the order of that list deterministic? Does that mean we would end up using the same node since below we use zones[0]?

pkg/blockstorage/zone/zone.go
```diff
@@ -83,20 +86,45 @@ func getAvailableZones(ctx context.Context, kubeCli kubernetes.Interface, validZ
...
+		for _, zone := range sourceZones {
+			if z := consistentZone(zone, sanitizedAvailableZones); z != "" {
+				newZones[z] = struct{}{}
+			}
 		}
 	}
 	return newZones
 }
 
+// consistentZone will return the same zone given a source and a list of zones.
```
tdmanv: Please also add some comments on why we need to do this, otherwise the next change to this implementation may also    revert this functionality.

PR #589, opened 2020-02-27
Extend ResourceRequirement with LabelSelector

tylerje:
## Change Overview

Allow label filtering in ResourceMatcher by adding LabelSelector to ResourceRequirement and Resource in filter.go.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #(internal, no issue in repo on this)

## Test Plan

`go test -v ./pkg/filter -check.f=FilterSuite --check.vv`

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/filter/filter.go
```diff
@@ -108,18 +112,36 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
 	v1.LocalObjectReference `json:",inline,omitempty"`
+	// Provides the Group, Version, and Resource values (GVR)
 	ResourceTypeRequirement `json:",inline,omitempty"`
+	// Specifies a set of label requirements to be used as filters for matches
+	metav1.LabelSelector `json:",inline,omitempty"`
 }
 
 // Matches returns true if the specified resource name/GVR matches the requirement
-func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource) bool {
+func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource, obj runtime.Object) bool {
```
tdmanv: I know we said to use runtime.Object, but I wonder if metav1.Object would be easier to use since it has a `GetLabels()` method

vkamra: Should the input here just be `labels map[string]string`? The caller can extract the labels and pass those in

tylerje: @vkamra that may be a better approach given that we can create a LabelSelector and then Selector and pass the map to the Selector's Matches method. This would make the use of the filter pkg simpler.

tylerje: Have moved to map as @vkamra suggested. This may require a change in how clients create a ResourceMatcher in order to get labels to be used. I also need to add tests to cover the changes before making this ready for review.

pkg/filter/filter.go
```diff
@@ -108,24 +119,40 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
+	// Specifies a set of label requirements to be used as filters for matches
+	metav1.LabelSelector `json:",inline,omitempty"`
 }
 
 // Matches returns true if the specified resource name/GVR matches the requirement
-func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource) bool {
+func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource, resourceLabels map[string]string) bool {
 	// If the requirement does not specify a resource name - only check the
 	// ResourceTypeRequirement i.e. GVR match
+	isMatch := false
```
ewhamilton: I think you'll find this to be more readable and simpler if we notice that the old code needlessly uses `r.LocalObjectReference.Name` and `r.Name`-- they're the same field.  Also `matches(r.Name, name)` already handles the case of `r.Name == ""`, so the original code is more complex than needed.

Could turn into something like `switch{
case !matches(r.Name, name):
    return false
case !matches(r.ResourceTypeRequiremnt.Matches(gvr):
   return false
}
check the labels...
`
Or could unroll and use matches on r.Group, r.Version, and r.Resource too...

tylerje: Good suggestion. Have taken the simple way out to simplify for now.

pkg/filter/filter.go
```diff
@@ -108,24 +119,35 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
 
 // Matches returns true if the specified resource name/GVR matches the requirement
-func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource) bool {
+func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource, resourceLabels map[string]string) bool {
 	// If the requirement does not specify a resource name - only check the
 	// ResourceTypeRequirement i.e. GVR match
-	if r.LocalObjectReference.Name == "" {
-		return r.ResourceTypeRequirement.Matches(gvr)
+	isMatch := matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)
+	if isMatch && len(resourceLabels) != 0 && (len(r.MatchExpressions) != 0 || len(r.MatchLabels) != 0) {
```
aaron-kasten: nit: I think this is the code @ewhamilton was discussing previously.  This can be simplified and the nesting bumped up by negating the tests. ie. `if !isMatch { return false }` etc...

kernel.org/process/coding-style.rst: "... if you need more than 3 levels of indentation, you’re screwed anyway, and should fix your program."

pkg/filter/filter.go
```diff
@@ -108,24 +119,35 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
+		lsSel := &metav1.LabelSelector{MatchLabels: r.MatchLabels, MatchExpressions: r.MatchExpressions}
+		if sel, err := metav1.LabelSelectorAsSelector(lsSel); err == nil {
+			lsSet := labels.Set(resourceLabels)
```
aaron-kasten: shadowing here ... I recommend you integrate the call `labels.Set(resourceLabels)` into `Matches()` which removes the shadow and leaves `lsSel` in outer scope intact.

pkg/filter/filter_test.go
```diff
@@ -375,10 +376,14 @@ func (s *FilterSuite) TestJoin(c *C) {
 func (s *FilterSuite) TestResourceIncludeExclude(c *C) {
 	ssTypeRequirement := ResourceTypeRequirement{Group: "apps", Resource: "statefulsets"}
 	pvcTypeRequirement := ResourceTypeRequirement{Version: "v1", Resource: "persistentvolumeclaims"}
-	ss1 := Resource{Name: "ss1", GVR: schema.GroupVersionResource{Group: "apps", Resource: "statefulsets"}}
-	ss2 := Resource{Name: "specificname", GVR: schema.GroupVersionResource{Group: "apps", Resource: "statefulsets"}}
-	pvc1 := Resource{Name: "pvc1", GVR: schema.GroupVersionResource{Version: "v1", Resource: "persistentvolumeclaims"}}
-	pvc2 := Resource{Name: "specificname", GVR: schema.GroupVersionResource{Version: "v1", Resource: "persistentvolumeclaims"}}
+	ss1 := Resource{Name: "ss1", GVR: schema.GroupVersionResource{Group: "apps", Resource: "statefulsets"},
```
aaron-kasten: My unit-test don't work for this.  No action yet as Shippable will check.

tylerje: I'm running
`go test -v ./pkg/filter -check.f=FilterSuite --check.vv`
and tests pass.

pkg/filter/filter.go
tylerje: Great suggestions. Pushed updates.

pkg/filter/filter.go
```diff
@@ -16,15 +16,19 @@ package filter
...
 )
 
+// ResourceTypeRequirement contains group, version and resource values
 type ResourceTypeRequirement struct {
 	Group    string `json:"group,omitempty"`
 	Version  string `json:"version,omitempty"`
 	Resource string `json:"resource,omitempty"`
 }
 
+// Matches returns true if group, version and resource values match or are empty
```
ewhamilton: Thanks for the comments.

pkg/filter/filter.go
```diff
@@ -108,24 +119,34 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
 type ResourceRequirement struct {
+	// Provides the Name of the resource object
 	v1.LocalObjectReference `json:",inline,omitempty"`
+	// Provides the Group, Version, and Resource values (GVR)
 	ResourceTypeRequirement `json:",inline,omitempty"`
+	// Specifies a set of label requirements to be used as filters for matches
+	metav1.LabelSelector `json:",inline,omitempty"`
 }
 
 // Matches returns true if the specified resource name/GVR matches the requirement
```
ewhamilton:
```suggestion
// Matches returns true if the specified resource name/GVR/labels matches the requirement
```

pkg/filter/filter.go
```diff
@@ -108,24 +119,34 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
-func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource) bool {
+func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource, resourceLabels map[string]string) bool {
 	// If the requirement does not specify a resource name - only check the
 	// ResourceTypeRequirement i.e. GVR match
-	if r.LocalObjectReference.Name == "" {
-		return r.ResourceTypeRequirement.Matches(gvr)
+	matchFound := matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)
```
ewhamilton: As I think @aaron-kasten mentioned (if not here, maybe elsewhere), this would be cleaner using short retrurns and not using a `matchFound`.  Also I think in case of error we don't want to match, so:
```
if !matches(r.Name, name) || !r.ResourceTypeRequirement.Matches(gvr) {
    return false
}
if len(r.MatchExpressions) == 0 && len(r.MatchLabels) == 0 {
    return true
}
sel, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: r.MatchLabels, MatchExpressions: r.MatchExpressions})
if err != nil {
    // malformed requirement labels
    return false
}
return sel.Matches(labels.Set(resourceLabels))
```

tylerje: I've committed these changes with two minor differences. See comments in code for the explanations.

pkg/filter/filter.go
```diff
@@ -108,24 +119,41 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
 	// If the requirement does not specify a resource name - only check the
 	// ResourceTypeRequirement i.e. GVR match
-	if r.LocalObjectReference.Name == "" {
-		return r.ResourceTypeRequirement.Matches(gvr)
+	if !(matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)) {
+		return false
 	}
-	return matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)
+	if len(resourceLabels) == 0 || (len(r.MatchExpressions) == 0 && len(r.MatchLabels) == 0) {
+		// If there are no resourceLabels to be matched, the LabelSelector is ignored.
```
ewhamilton: Thanks for the comments and clearly stated rationale.  I don't think this is the policy that we want.  Consider the use case of "match everything with label foo=bar".

pkg/filter/filter.go
```diff
@@ -108,24 +119,41 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
+		// If there is no LabelSelector defined, the specified resourceLabels are ignored.
+		return true
+	}
+	sel, err := metav1.LabelSelectorAsSelector(&r.LabelSelector)
+	if err != nil {
+		// A match was found on Name or GVR but labels could not be evaluated.
```
ewhamilton: Arguably we should have validated the label when policy or action was validated (let's make sure to to so), but if we get here we have a policy question of whether to ignore the label and treat as match or fail to match.  My sense is that we shouldn't match (better to do nothing than do the wrong thing), but if there's precedent elsewhere I'd opt for consistency.  Do you know of any analogues?  What's your reasoning?

pkg/filter/filter.go
tylerje: This would occur if the calling code failed to pass in the resource labels. Would it be better to return an error if LabelSelector is specified and no labels are supplied? Or do we want to assume that no match in that case occurs? I think we have the logic working. Now we need to decide what the policy would be. I think you're right but I want to think through each case and gather input. @vkamra @tdmanv ??

pkg/filter/filter.go
tylerje: Reasoning is just a gut feel. I'm more inclined to return the error rather than burying it and making an assumption either way.

pkg/filter/filter.go
aaron-kasten: @tylerje I'm in agreement ... time to write up what the inputs and outputs would be of various scenarios.  Overall you are on the right track.

pkg/filter/filter_test.go
```diff
@@ -453,6 +458,18 @@ func (s *FilterSuite) TestResourceIncludeExclude(c *C) {
 			include:   []Resource{ss2, pvc2},
 			exclude:   []Resource{ss1, pvc1},
 		},
+		{
```
aaron-kasten: Need more unit tests here to demonstrate usage and corner cases.  For example, test to filter only on label.

tylerje: Added two tests which required a modification to the Matches logic.

pkg/filter/filter.go
tylerje: Two tests added. This required a change to the logic.

pkg/filter/filter.go
tylerje: Additional tests cleared up and changes made that align with this comment.

tylerje:
@aaron-kasten that is a good idea. I'll get that done.

pkg/filter/filter.go
```diff
@@ -108,24 +119,44 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
+	// If ResourceRequirement Name/GVR value specified, exact match on parameter is required.
+	if !(matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)) {
+		return false
 	}
-	return matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)
+	if len(resourceLabels) == 0 && len(r.MatchExpressions) == 0 && len(r.MatchLabels) == 0 {
+		// If there are no resourceLabels and no LabelSelector we return match of GVR from above.
+		return true
+	}
+	if len(resourceLabels) == 0 && (len(r.MatchExpressions) != 0 || len(r.MatchLabels) != 0) {
```
ewhamilton: This does not handle the case where the requirement is an exclusion label.  Such as `r.MatchExpressions{{key: foo,  operator: NotIn, values: [bar]}}`.

pkg/filter/filter.go
```diff
@@ -108,24 +119,44 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
-		return r.ResourceTypeRequirement.Matches(gvr)
+// Matches returns true if the specified resource name/GVR/labels matches the requirement
+func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource, resourceLabels map[string]string) bool {
+	// If ResourceRequirement Name/GVR value of empty string matches on any specified in parameters.
+	// If ResourceRequirement Name/GVR value specified, exact match on parameter is required.
+	if !(matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)) {
+		return false
 	}
-	return matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)
+	if len(resourceLabels) == 0 && len(r.MatchExpressions) == 0 && len(r.MatchLabels) == 0 {
```
ewhamilton: @vkamra, @tdmanv We need a policy clarification/decision here please.  If there is no LabelSelector in the requirement, should we / can we return a true match regardless what the labels are on the object.

@tylerje Another way of asking this is, what will an empty LabelSelector return from sel.Matches below?

vkamra: @ewhamilton - I think that makes sense. If the user didn't specify labels as a requirement - the `resourceLabels` should be ignored.

pkg/filter/filter.go
tylerje: This is true. Do we have to assume that if the caller does not supply resourceLabels, the MatcheExpressions should be ignored? If a resource has no labels and the user wants to include labels NotIn [list], then we would want to return a match true given that the resource has no labels "in" that list. Let me experiment and see what I can come up with.

tylerje: Removed that check and added test. You're right. It's not needed.

pkg/filter/filter.go
```diff
@@ -108,24 +119,40 @@ func JoinResourceTypeMatchers(ms ...ResourceTypeMatcher) ResourceTypeMatcher {
...
-func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource) bool {
-	// If the requirement does not specify a resource name - only check the
-	// ResourceTypeRequirement i.e. GVR match
-	if r.LocalObjectReference.Name == "" {
-		return r.ResourceTypeRequirement.Matches(gvr)
+// Matches returns true if the specified resource name/GVR/labels matches the requirement
+func (r ResourceRequirement) Matches(name string, gvr schema.GroupVersionResource, resourceLabels map[string]string) bool {
+	// If ResourceRequirement Name/GVR value of empty string matches on any specified in parameters.
+	// If ResourceRequirement Name/GVR value specified, exact match on parameter is required.
+	if !(matches(r.Name, name) && r.ResourceTypeRequirement.Matches(gvr)) {
```
ewhamilton:
```suggestion
	if !(matches(r.Name, name) || !r.ResourceTypeRequirement.Matches(gvr)) {
```
Please add a test case too-- in this case for name being same and GVR different.

pkg/filter/filter.go
ewhamilton: @tyler In the interest of clarity and conceptual alignment, we can remove the `len(resourceLabels) == 0 && ` term from line 137 and update the comment on line 138.

This makes it clearer that:
* If there is no required Name (r.Name == ""), any name (including none) matches
* If there is no required Group (r.Group == ""), any group (including none) matches
* If there is no required Version (r.Version == ""), any version (including none) matches
* If there is no required Resource (r.Version == ""), any resource (including none) matches
* If there is no required LabelSelector (r.LabelSelector == LabelSelector{}), any label (including none) matches

pkg/filter/filter.go
tylerje: Good catch. You mentioned this before but in my refactoring it got lost. Test and change added.

PR #598, opened 2020-03-03
Support to have OpenShift apps ephemeral or persistent

viveksinghggits:
## Change Overview
In current implementation if we deployed OpenShift apps as part of test, the apps were persistent always this change adds the support to have epehmeral apps as well.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan
Run 
```
make openshift-test
```
to test the things, make sure you have setup a minishift cluster using command 
```
make start-minishift vm-driver=virtualbox
```
Logs of this test can be found [here](https://gist.github.com/viveksinghggits/9817f35c0d78f5b4dd706ba15d04e353)

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/app/mongodb-deploymentconfig.go
+++ b/pkg/app/mongodb-deploymentconfig.go
@@ -38,27 +38,27 @@ const (
 )
 
 type MongoDBDepConfig struct {
-	cli        kubernetes.Interface
-	osCli      osversioned.Interface
-	name       string
-	namespace  string
-	dbTemplate string
-	user       string
-	label      string
-	osClient   openshift.OSClient
-	params     map[string]string
+	cli         kubernetes.Interface
+	osCli       osversioned.Interface
+	name        string
+	namespace   string
+	user        string
+	label       string
+	osClient    openshift.OSClient
+	params      map[string]string
+	storageType storage
 }
 
 func NewMongoDBDepConfig(name string) App {
 	return &MongoDBDepConfig{
-		name:       name,
-		dbTemplate: getOpenShiftDBTemplate(mongoDepConfigName, ephemeralStorage),
-		user:       "admin",
+		name: name,
+		user: "admin",
 		params: map[string]string{
 			"MONGODB_ADMIN_PASSWORD": "secretpassword",
 		},
-		label:    getLabelOfApp(mongoDepConfigName),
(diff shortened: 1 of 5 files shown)


pkg/app/mysql-deploymentconfig.go
```diff
@@ -55,7 +55,7 @@ type MysqlDepConfig struct {
 func NewMysqlDepConfig(name string) App {
 	return &MysqlDepConfig{
 		name:       name,
-		dbTemplate: getOpenShiftDBTemplate(mysqlDepConfigName),
+		dbTemplate: getOpenShiftDBTemplate(mysqlDepConfigName, ephemeralStorage),
```
PrasadG193: We should be to choose persistent storage type also. We can add a field in `MysqlDepConfig` and refer it

pkg/app/mongodb-deploymentconfig.go
```diff
@@ -34,31 +34,31 @@ import (
...
+	name        string
+	namespace   string
+	user        string
+	label       string
+	osClient    openshift.OSClient
+	params      map[string]string
+	storageType storage
 }
 
 func NewMongoDBDepConfig(name string) App {
```
PrasadG193: Should we accept storagetype as a argument to the function? It will allow use to switch storage types if we import and use this in another project.

pkg/app/utils.go
```diff
@@ -21,9 +21,13 @@ import (
 )
 
 const (
-	dbTemplateURI = "https://raw.githubusercontent.com/openshift/origin/master/examples/db-templates/%s-persistent-template.json"
+	dbTemplateURI = "https://raw.githubusercontent.com/openshift/origin/v3.11.0/examples/db-templates/%s-%s-template.json"
+	//persistentStorage storage = "persistent"
```
PrasadG193: should uncomment `persistentStorage`?

viveksinghggits: Hey @PrasadG193,  we should not uncomment `persistentStorage` because we are  not using it anywhere as of now.

PR #602, opened 2020-03-04
Upgrade kanister build image to v0.0.8

SupriyaKasten:
## Change Overview

Upgrade kansiter build image to v0.0.8

## Pull request type

Please check the type of change your PR introduces:
- [x] :hamster: Trivial/Minor

PR #597, opened 2020-03-03
Update build image binaries to the latest version

PrasadG193:
## Change Overview

Bump build image binaries to the latest version:
kind: v0.7.0
kubectl: v1.17 .0
golangci-lint: v1.23.7
goreleaser: v0.127.0

## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [x] :rainbow: Refactoring (no functional changes, no api changes)
- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

PR #605, opened 2020-03-05
Revert "Upgrade kanister build image to v0.0.8"

PrasadG193:
Reverts kanisterio/kanister#602

@SupriyaKasten the image has to be built from PR - https://github.com/kanisterio/kanister/pull/597
I am assuming you have built the image from `master`
```
$ docker run --rm kanisterio/build:v0.0.8 kind version
v0.5.1
```
That's why its still using older dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ IMAGE_NAME := $(BIN)
 
 IMAGE := $(REGISTRY)/$(IMAGE_NAME)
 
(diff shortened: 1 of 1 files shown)


PrasadG193:
Let's not override the existing image, instead push image with newer tag

SupriyaKasten:
@PrasadG193 my bad, I might have mistakenly checked out one of my local branch before running `docker built`. Sorry about that 🤦‍♀

PR #607, opened 2020-03-05
Revert "Revert "Extend ResourceRequirement with LabelSelector""

tylerje:
Reverts kanisterio/kanister#603
In draft until DeepCopyInto can be added to ResourceRequirement in this branch.

### DIFF
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -127,6 +127,13 @@ type ResourceRequirement struct {
 	metav1.LabelSelector `json:",inline,omitempty"`
 }
 
+// DeepCopyInto provides explicit deep copy implementation to avoid
+func (r ResourceRequirement) DeepCopyInto(out *ResourceRequirement) {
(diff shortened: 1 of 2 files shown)


tylerje:
DeepCopyInto support added and tested by copying into K10 vendor code to assure the deepcopy-gen code will behave properly.

pkg/filter/filter.go
```diff
@@ -127,6 +127,13 @@ type ResourceRequirement struct {
 	metav1.LabelSelector `json:",inline,omitempty"`
 }
 
+// DeepCopyInto provides explicit deep copy implementation to avoid
+func (r ResourceRequirement) DeepCopyInto(out *ResourceRequirement) {
```
tdmanv: Please also add a simple unit test that confirms that this does do a DeepCopy

tylerje: @tdmanv are we ready to pull the hold off merging?

PR #613, opened 2020-03-09
Changing implementation of FromSourceRegionZone

bathina2:
Reverts kanisterio/kanister#612

### DIFF
--- a/pkg/blockstorage/zone/zone.go
+++ b/pkg/blockstorage/zone/zone.go
@@ -79,6 +79,8 @@ func getAvailableZones(ctx context.Context, m Mapper, kubeCli kubernetes.Interfa
 		log.Info().Print("Source region and available region mismatch", field.M{"sourceRegion": sourceRegion, "availableRegion": availableRegion})
 	}
 	// TODO: validate availableRegion
+	// if we fail to get zones from the available region,
+	// we will defer to the source region for the zones.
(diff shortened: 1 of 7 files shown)


bathina2:
Added UTs to test cross region migration as well

pkg/blockstorage/zone/zone.go
```diff
@@ -79,12 +78,14 @@ func getAvailableZones(ctx context.Context, kubeCli kubernetes.Interface, validZ
 	if availableRegion != sourceRegion {
 		log.Info().Print("Source region and available region mismatch", field.M{"sourceRegion": sourceRegion, "availableRegion": availableRegion})
 	}
-	if len(availableZones) <= 0 { // Will never occur, NodeZonesAndRegion returns error if empty
-		log.Info().Print("No available zones found", field.M{"availableRegion": availableRegion})
+	// TODO: validate availableRegion
+	validZoneNames, err := m.FromRegion(ctx, availableRegion)
+	if err != nil || len(validZoneNames) == 0 {
```
tdmanv: what happens when `err != nil`?

bathina2: That is when we fail to get any zones from the available region (from k8s). We defer to using the source region.

tdmanv: Makes sense. Please add a comment since we're dropping the error.

PR #623, opened 2020-03-12
Kanister docs update to version 0.27.0

vkamra:


julio-lopez:
if we do frequent releases then this will require a manually created PR every time?

PR #628, opened 2020-03-18
Avoid using operation timeout while executing WaitForPodReady func

SupriyaKasten:
## Change Overview

Avoid using operation timeout while executing WaitForPodReady func

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod.go
+++ b/pkg/kube/pod.go
@@ -205,7 +205,6 @@ func checkPVCAndPVStatus(vol v1.Volume, p *v1.Pod, cli kubernetes.Interface, nam
 		if pv.Status.Phase == v1.VolumeFailed {
 			return errors.Errorf("PV %s associated with PVC %s has status: %s message: %s reason: %s namespace: %s", pvName, pvcName, v1.VolumeFailed, pv.Status.Message, pv.Status.Reason, namespace)
 		}
-
 	} else if pvc.Status.Phase == v1.ClaimLost {
 		return errors.Errorf("PVC %s assoicated with pod %s has status: %s", pvcName, p.Name, v1.ClaimLost)
 	}
--- a/pkg/kube/pod.go
+++ b/pkg/kube/pod.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"io"
 	"io/ioutil"
+	"time"
 
 	json "github.com/json-iterator/go"
 	"github.com/pkg/errors"
@@ -30,6 +31,9 @@ import (
 	"github.com/kanisterio/kanister/pkg/log"
 	"github.com/kanisterio/kanister/pkg/poll"
 )
+
+// podReadyWaitTimeout is the time to wait for pod to be ready
+const podReadyWaitTimeout = 5 * time.Minute // Q: Is 5 minutes sufficient?
 
 // PodOptions specifies options for `CreatePod`
 type PodOptions struct {
@@ -114,7 +118,9 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
 
 // WaitForPodReady waits for a pod to exit the pending state
 func WaitForPodReady(ctx context.Context, cli kubernetes.Interface, namespace, name string) error {
-	err := poll.Wait(ctx, func(ctx context.Context) (bool, error) {
+	timeoutCtx, waitCancel := context.WithTimeout(ctx, podReadyWaitTimeout)
+	defer waitCancel()
+	err := poll.Wait(timeoutCtx, func(ctx context.Context) (bool, error) {
 		p, err := cli.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
 		if err != nil {
 			return false, err
@@ -123,7 +129,7 @@ func WaitForPodReady(ctx context.Context, cli kubernetes.Interface, namespace, n
 		// check if nodes are up and available
 		n := strings.Split(p.NodeName, "/")
 		if n[0] != "" {
-			ns, err := cli.CoreV1().Nodes().Get(n[0], metav1.GetOptions{})
+			node, err := cli.CoreV1().Nodes().Get(n[0], metav1.GetOptions{})
 			if err != nil {
 				return false, errors.Wrapf(err, "Failed to get node %s", n[0])
 			}
@@ -152,19 +158,6 @@ func WaitForPodReady(ctx context.Context, cli kubernetes.Interface, namespace, n
 				} else if pvc.Status.Phase == v1.ClaimLost {
 					return false, errors.Errorf("PVC %s assoicated with pod %s has status: %s", pvcName, name, v1.ClaimLost)
 				}
-			}
-		}
-
-		// check if container is in healthy state
-		for _, container := range p.Status.InitContainerStatuses {
-			if container.State == v1.ContainerStateWaiting {
-				// Might not need to check container state
-			}
-		}
-
-		for _, container = range p.Status.ContainerStatus {
-			if container.State == v1.ContainerStateWaiting {
-				// Might not need to check container state
 			}
 		}
 
--- a/pkg/kube/pod.go
+++ b/pkg/kube/pod.go
@@ -217,7 +217,10 @@ func checkPVCAndPVStatus(vol v1.Volume, p *v1.Pod, cli kubernetes.Interface, nam
 		}
 	}
 
-	if pvc.Status.Phase == v1.ClaimPending {
+	switch pvc.Status.Phase {
+	case v1.ClaimLost:
+		return errors.Errorf("PVC %s assoicated with pod %s has status: %s", pvcName, p.Name, v1.ClaimLost)
+	case v1.ClaimPending:
 		pvName := pvc.Spec.VolumeName
 		if pvName == "" {
 			// wait for timeout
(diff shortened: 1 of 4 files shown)


pkg/kube/pod.go
```diff
@@ -114,16 +119,71 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
 	return errors.Wrapf(err, "Pod did not transition into running state. Namespace:%s, Name:%s", namespace, name)
 }
 
+func checkPVCAndPVStatus(p *v1.Pod, cli kubernetes.Interface, namespace string) error {
+	for _, vol := range p.Spec.Volumes {
+		if vol.VolumeSource.PersistentVolumeClaim != nil {
+			pvcName := vol.VolumeSource.PersistentVolumeClaim.ClaimName
+			pvc, err := cli.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
+			if err != nil {
+				return errors.Wrapf(err, "Failed to get pvc %s", pvcName)
```
pavannd1: [nit] Comparing with other logs. Some have caps and some don't
```suggestion
				return errors.Wrapf(err, "Failed to get PVC %s", pvcName)
```

pkg/kube/pod.go
```diff
@@ -114,16 +119,71 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
+				pvName := pvc.Spec.VolumeName
+				if pvName == "" {
+					return errors.Errorf("PVC %s in namespace %s not bound", pvcName, namespace)
+				}
+				pv, err := cli.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
+				if err != nil {
+					return errors.Wrapf(err, "Failed to get PV %s, namespace: %s", pvName, namespace)
+				}
+				if pv.Status.Phase == v1.VolumeFailed {
+					return errors.Errorf("PV %s associated with pvc %s has status: %s message: %s reason: %s namespace: %s", pvName, pvcName, v1.VolumeFailed, pv.Status.Message, pv.Status.Reason, namespace)
```
pavannd1:
```suggestion
					return errors.Errorf("PV %s associated with PVC %s has status: %s message: %s reason: %s namespace: %s", pvName, pvcName, v1.VolumeFailed, pv.Status.Message, pv.Status.Reason, namespace)
```
nit again!

pkg/kube/utils.go
```diff
@@ -134,3 +134,25 @@ func GetRegionFromLabels(labels map[string]string) string {
...
+
+// IsNodeReady returns true if it's Ready condition is set to true
+// Derived from "k8s.io/kubernetes/test/e2e/framework/node"
+func IsNodeReady(node *v1.Node) bool {
+	for _, cond := range node.Status.Conditions {
+		if cond.Type == v1.NodeReady {
+			if cond.Status == v1.ConditionTrue {
+				return true
+			}
+		}
```
julio-lopez: Any reason for not collapsing these two?
```suggestion
		if cond.Type == v1.NodeReady && cond.Status == v1.ConditionTrue {
			return true
		}
```

pkg/kube/pod.go
```diff
@@ -31,6 +33,9 @@ import (
 	"github.com/kanisterio/kanister/pkg/poll"
 )
 
+// podReadyWaitTimeout is the time to wait for pod to be ready
+const podReadyWaitTimeout = 5 * time.Minute // Q: Is 5 minutes sufficient?
```
julio-lopez: Did you mean to make this a PR question or something to follow up on later?
If the later, then file a ticket, otherwise:

```suggestion
const podReadyWaitTimeout = 5 * time.Minute
```

SupriyaKasten: @julio-lopez yes thanks, I meant to make it a PR question. 

@tdmanv @julio-lopez would love to get your input on this Q.

julio-lopez: I do not know what a good value would be, that seems long enough, but not sure what other implications it may have. @tdmanv should be able to provide better guidance here.

julio-lopez:
At a high level:
👍 LG +1

pkg/kube/pod.go
tdmanv: I also don't have a good sense here, but 5 minutes sounds as good as anything. Let's include 5 minutes in the error we return so we know we what this is set to.

pkg/kube/pod.go
```diff
@@ -114,16 +119,71 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
+	timeoutCtx, waitCancel := context.WithTimeout(ctx, podReadyWaitTimeout)
+	defer waitCancel()
+	err := poll.Wait(timeoutCtx, func(ctx context.Context) (bool, error) {
 		p, err := cli.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
 		if err != nil {
 			return false, err
 		}
+
+		// check if nodes are up and available
+		n := strings.Split(p.Spec.NodeName, "/")
```
tdmanv: Let's put this logic into it's own function - I think that would make this more readable

pkg/kube/pod.go
```diff
@@ -114,16 +119,71 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
+		if err := checkPVCAndPVStatus(p, cli, namespace); err != nil {
+			return false, err
+		}
+
 		return p.Status.Phase != v1.PodPending && p.Status.Phase != "", nil
 	})
 	return errors.Wrapf(err, "Pod did not transition into running state. Namespace:%s, Name:%s", namespace, name)
 }
 
+func checkPVCAndPVStatus(p *v1.Pod, cli kubernetes.Interface, namespace string) error {
```
tdmanv: please describe the intent of this function a bit - if the PVC is not ready, is that an error condition and we should return right away? Or does that mean we should wait for the timeout?  As-is this function mixes approaches. if `PersistentVolumeClaim` is nil for any volume, we return no error, but other empty conditions are treated as errors.

pkg/kube/pod.go
```diff
@@ -114,16 +119,71 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
+	for _, vol := range p.Spec.Volumes {
+		if vol.VolumeSource.PersistentVolumeClaim != nil {
+			pvcName := vol.VolumeSource.PersistentVolumeClaim.ClaimName
+			pvc, err := cli.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
+			if err != nil {
+				return errors.Wrapf(err, "Failed to get PVC %s", pvcName)
+			}
+			if pvc.Status.Phase == v1.ClaimPending {
```
tdmanv: what guarantees do we have if a claim is pending? Do they change if we're using a StatefulSet vs. a Deployment?

pkg/kube/pod.go
SupriyaKasten: So here we are checking if the PVC is stuck in pending state or has status =  `ClaimLost` .
If the PVC is stuck in pending state then we check the status of the PV associate with that PVC. If the PV is lost or has status failed then we return an error since the pvc will not transition into running state and hence the pods will also not transition into running state. otherwise we wait for the timeout.
if the status is `claimlost` then we return an error.

pkg/kube/pod.go
SupriyaKasten: Do you mean guarantee of the pod coming up? or the PVC transitioning into `running` ?

SupriyaKasten: Discussed the changes offline with @tdmanv , pushing a commit for it.

pkg/kube/pod.go
```diff
@@ -114,14 +119,97 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
+		}
+		pv, err := cli.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
+		if err != nil {
+			// wait for timeout
+			return nil
+		}
+		if pv.Status.Phase == v1.VolumeFailed {
+			return errors.Errorf("PV %s associated with PVC %s has status: %s message: %s reason: %s namespace: %s", pvName, pvcName, v1.VolumeFailed, pv.Status.Message, pv.Status.Reason, namespace)
+		}
+
```
tdmanv: Fix golint
```
pkg/kube/pod.go:208: unnecessary trailing newline (whitespace)
	} else if pvc.Status.Phase == v1.ClaimLost {
Makefile:196: recipe for target 'run' failed
```

```suggestion
```

pkg/kube/pod.go
```diff
@@ -114,14 +119,97 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
+	}
+	pvcName := vol.VolumeSource.PersistentVolumeClaim.ClaimName
+	pvc, err := cli.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
+	if err != nil {
+		// Do not return err, wait for timeout, since sometimes in case of statefulsets, they trigger creation of a volume
+		return nil
+	}
+	if pvc.Status.Phase == v1.ClaimPending {
+		pvName := pvc.Spec.VolumeName
+		if pvName == "" {
```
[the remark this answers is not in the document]
SupriyaKasten: Yes, in case of StatefulSet, it is possible that the pvName is empty in PVC spec before the pv is created.

SupriyaKasten:
> Thank you for the logic change, I think that looks more idiomtic.
> 
> In the test plan you've checked the unit test box - please highlight the unit tests that cover these changes. If none exist, please add a kube test that verifies that we correctly handle the provisioning lifecycle of a StatefulSet.

@tdmanv I have made changes to `TestPodWithVolumes()` func in kube/pod_test.go to test waitForPodReady() func.

SupriyaKasten:
@tdmanv  addressed the review feedback, ready for another review

PR #629, opened 2020-03-18
Use dynamic client to create volume snapshots

PrasadG193:
## Change Overview

Use dynamic client to create volume snapshots
Remove external-snapshotter dep from snapshot pkg
New interface Snapshotter to perform snapshot operations
Remove external-snapshotter dep from kube/volume package

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/snapshot/apis/v1alpha1/types.go
+++ b/pkg/kube/snapshot/apis/v1alpha1/types.go
@@ -1,4 +1,6 @@
 /*
+Copyright 2020 The Kanister Authors.
+
 Copyright 2018 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
--- a/pkg/kube/snapshot/apis/v1alpha1/types.go
+++ b/pkg/kube/snapshot/apis/v1alpha1/types.go
@@ -30,6 +30,9 @@ const (
 	VolumeSnapshotResourcePlural = "volumesnapshots"
 	// VolumeSnapshotClassResourcePlural is "volumesnapshotclasses"
 	VolumeSnapshotClassResourcePlural = "volumesnapshotclasses"
+
+	GroupName = "snapshot.storage.k8s.io"
+	Version   = "v1alpha1"
 )
 
 // VolumeSnapshot is a user's request for taking a snapshot. Upon successful creation of the actual
--- a/pkg/kube/snapshot/apis/v1alpha1/types.go
+++ b/pkg/kube/snapshot/apis/v1alpha1/types.go
@@ -1,0 +1,204 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	core_v1 "k8s.io/api/core/v1"
+	storage "k8s.io/api/storage/v1beta1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// VolumeSnapshotContentResourcePlural is "volumesnapshotcontents"
+	VolumeSnapshotContentResourcePlural = "volumesnapshotcontents"
+	// VolumeSnapshotResourcePlural is "volumesnapshots"
+	VolumeSnapshotResourcePlural = "volumesnapshots"
+	// VolumeSnapshotClassResourcePlural is "volumesnapshotclasses"
+	VolumeSnapshotClassResourcePlural = "volumesnapshotclasses"
+)
+
+// VolumeSnapshot is a user's request for taking a snapshot. Upon successful creation of the actual
+// snapshot by the volume provider it is bound to the corresponding VolumeSnapshotContent.
+// Only the VolumeSnapshot object is accessible to the user in the namespace.
+type VolumeSnapshot struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Spec defines the desired characteristics of a snapshot requested by a user.
+	Spec VolumeSnapshotSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
+
+	// Status represents the latest observed state of the snapshot
+	Status VolumeSnapshotStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// VolumeSnapshotList is a list of VolumeSnapshot objects
+type VolumeSnapshotList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+	// Items is the list of VolumeSnapshots
+	Items []VolumeSnapshot `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+// VolumeSnapshotSpec describes the common attributes of a volume snapshot
+type VolumeSnapshotSpec struct {
+	// Source has the information about where the snapshot is created from.
+	// In Alpha version, only PersistentVolumeClaim is supported as the source.
+	// If not specified, user can create VolumeSnapshotContent and bind it with VolumeSnapshot manually.
+	Source *core_v1.TypedLocalObjectReference `json:"source" protobuf:"bytes,1,opt,name=source"`
+
+	// SnapshotContentName binds the VolumeSnapshot object with the VolumeSnapshotContent
+	SnapshotContentName string `json:"snapshotContentName" protobuf:"bytes,2,opt,name=snapshotContentName"`
+
+	// Name of the VolumeSnapshotClass used by the VolumeSnapshot. If not specified, a default snapshot class will
+	// be used if it is available.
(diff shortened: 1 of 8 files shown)


pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+	if us == nil || len(us.Items) == 0 {
+		return "", errors.Errorf("Failed to find any VolumeSnapshotClass in the cluster: %v", err)
+	}
+	for _, vsc := range us.Items {
+		ans := vsc.GetAnnotations()
+		if val, ok := ans[annotationKey]; ok && val == annotationValue {
+			return vsc.GetName(), nil
+		}
+	}
+	return "", errors.Errorf("Failed to find VolumesnapshotClass with %s annotation in the cluster", annotationKey)
```
bathina2: Nit- do you want to mention the both key and value in the error?

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+	// VolSnapContentGVR specifies GVR schema for VolumeSnapshotContents
+	VolSnapContentGVR = schema.GroupVersionResource{Group: v1alpha1.GroupName, Version: v1alpha1.Version, Resource: v1alpha1.VolumeSnapshotContentResourcePlural}
+)
+
+type SnapshotAlpha struct {
+	kubeCli kubernetes.Interface
+	dynCli  dynamic.Interface
+}
+
+func (sna *SnapshotAlpha) GetVolumeSnapshotClass(annotationKey, annotationValue string) (string, error) {
```
bathina2: Nit - its self explanatory buy may wanna add a comment to fit the convention

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+	for _, vsc := range us.Items {
+		ans := vsc.GetAnnotations()
+		if val, ok := ans[annotationKey]; ok && val == annotationValue {
+			return vsc.GetName(), nil
+		}
+	}
+	return "", errors.Errorf("Failed to find VolumesnapshotClass with %s annotation in the cluster", annotationKey)
+}
+
+// Create creates a VolumeSnapshot and returns it or any error happened meanwhile.
```
bathina2: error "that" happened

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+func (sna *SnapshotAlpha) Create(ctx context.Context, name, namespace, volumeName string, snapshotClass *string, waitForReady bool) error {
```
bathina2: consider changing volumeName to pvcName because thats what it is

pkg/kube/snapshot/snapshot.go
```diff
@@ -17,237 +17,75 @@ package snapshot
...
+	//
+	// 'annotationKey' is the annotation key which has to be present on VolumeSnapshotClass.
+	// 'annotationValue' is the value for annotationKey in VolumeSnapshotClass spec.
+	// This returns error if no VolumeSnapshotClass found.
+	GetVolumeSnapshotClass(annotationKey, annotationValue string) (string, error)
+	// Create creates a VolumeSnapshot and returns it or any error happened meanwhile.
+	//
+	// 'name' is the name of the VolumeSnapshot.
+	// 'namespace' is namespace of the PVC. VolumeSnapshot will be crated in the same namespace.
+	// 'volumeName' is the name of the PVC of which we will take snapshot. It must be in the same namespace 'ns'.
```
bathina2: change to pvcName

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/kanisterio/kanister/pkg/kube/snapshot/apis/v1alpha1"
+	"github.com/kanisterio/kanister/pkg/poll"
+)
+
+const (
+	pvcKind = "PersistentVolumeClaim"
```
bathina2: PVCKind

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+			"kind":       VolSnapKind,
+			"metadata": map[string]interface{}{
+				"name":      name,
+				"namespace": namespace,
+			},
+			"spec": map[string]interface{}{
+				"source": map[string]interface{}{
+					"kind":      pvcObjectRef.Kind,
+					"name":      pvcObjectRef.Name,
+					"namespace": pvcObjectRef.Namespace,
```
bathina2: in sna.Create you are not calling createVolumeSnapshot with the pvcObjectRef.Namespace field set. Either set it there or use the namespace passed in the arguments.

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+	err = sna.WaitOnReadyToUse(ctx, name, namespace)
+	if err != nil {
+		return err
+	}
+
+	_, err = sna.Get(ctx, name, namespace)
+	return err
+}
+
+// Get will return the VolumeSnapshot in the namespace 'namespace' with given 'name'.
```
bathina2: nit - in a particular 'namespace' with a given 'name'

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// Delete will delete the VolumeSnapshot and returns any error as a result.
+func (sna *SnapshotAlpha) Delete(ctx context.Context, name, namespace string) error {
+	if err := sna.dynCli.Resource(VolSnapGVR).Namespace(namespace).Delete(name, &metav1.DeleteOptions{}); !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	// If the Snapshot does not exist, that's an acceptable error and we ignore it
+	return nil
```
bathina2: While I agree this is correct, maybe from a readability standpoint it would be better if 
```suggestion
	if err := sna.dynCli.Resource(VolSnapGVR).Namespace(namespace).Delete(name, &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
		return err
	}
	return nil
```

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
...
+
+	_, err = sna.Get(ctx, cloneName, cloneNamespace)
+	if err == nil {
+		return errors.Errorf("Target snapshot already exists in target namespace, Volumesnapshot: %s, Namespace: %s", cloneName, cloneNamespace)
+	}
+	if !k8errors.IsNotFound(err) {
+		return errors.Errorf("Failed to query target Volumesnapshot: %s, Namespace: %s: %v", cloneName, cloneNamespace, err)
+	}
+
+	src, err := sna.GetSource(ctx, name, namespace)
```
bathina2: You already fetched the original snapshot once above, this would make a second call to get the same object. I suggest calling sna.getContent. or creating something like get source from snapshot

pkg/kube/snapshot/apis/v1alpha1/types.go
```diff
@@ -0,0 +1,207 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
tdmanv: Please update to include Kanister Copyright

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -0,0 +1,317 @@
+// Copyright 2019 The Kanister Authors.
```
tdmanv: for new files, please use 2020

```suggestion
// Copyright 2020 The Kanister Authors.
```

pkg/kube/snapshot/snapshot_alpha.go
tdmanv: @bathina2 for one word changes like this one, you can use github suggestions

```suggestion
// Create creates a VolumeSnapshot and returns it or any error that happened meanwhile.
```

bathina2: Yea I used it elsewhere below

pkg/kube/snapshot/snapshot_alpha.go
PrasadG193: Should we export `PVCKind` though we are not using it anywhere else?

pkg/kube/snapshot/snapshot_alpha.go
bathina2: you are still passing in a pvcObjectRef with namspace not set in sna.Create. See line 87. Let me know if I'm missing something before resolving

pkg/kube/snapshot/snapshot_alpha.go
bathina2: Again you are doing two snapshot.Get() calls. you already have the snapshot object.

pkg/kube/snapshot/snapshot_alpha.go
PrasadG193: Aah, sorry, let me fix it

pkg/kube/snapshot/snapshot_alpha.go
bathina2: Its my opinion but these are well known and commonly used Kinds. Maybe they should be in a common kubernetes library, if they don't already exist. Not sure if the kubernetes client library has these defined but if it does we should use the ones they have instead of maintaining them ourselves.

pkg/kube/snapshot/snapshot_alpha.go
bathina2: Tell me why you want to resolve this. I think you can change the signature of GetSource to take a snapshot object and build a source. The only place this is used in K10 is in phase/snapshot.go line 316. and we're calling snapshot get before it there too. Theres no reason to make 2 calls to kubernetes to get the same snapshot

PrasadG193: I am sorry, I guess clicked "resolve" accidentally. I have another PR which already takes care of the optimisations. But we can address this here itself. In this PR, I have just moved the methods from snapshot.go to snapshot_alpha.go and changed the snapcli to dyncli without making any changes in existing workflow. I was thinking of handling optimisations in another PR

PR #649, opened 2020-04-14
Changes to support azure types

bathina2:
## Change Overview

Adding changes to support azure sku types. We were parsing the volume object incorrectly, and returning the tier instead of the type.

Additionally added support to set the sku type when creating volume from snapshot

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/blockstorage/azure/azuredisk.go
+++ b/pkg/blockstorage/azure/azuredisk.go
@@ -381,7 +381,7 @@ func (s *adStorage) VolumeParse(ctx context.Context, volume interface{}) (*block
 		Size:         int64(azto.Int32(vol.DiskSizeGB)),
 		Az:           az,
 		Tags:         blockstorage.MapToKeyValue(tags),
-		VolumeType:   azto.String(vol.Sku.Tier),
+		VolumeType:   string(vol.Sku.Name),
 		CreationTime: blockstorage.TimeStamp(vol.DiskProperties.TimeCreated.ToTime()),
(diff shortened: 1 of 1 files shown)


SupriyaKasten:
@bathina2 the PR looks good to me. If you haven't already, you can run any one or two blockstorage tests for Azure suite to verify your changes.

pkg/blockstorage/azure/azuredisk.go
```diff
@@ -489,6 +489,13 @@ func (s *adStorage) VolumeCreateFromSnapshot(ctx context.Context, snapshot block
 	if id != "" {
 		createDisk.Zones = azto.StringSlicePtr([]string{id})
 	}
+	for _, saType := range azcompute.PossibleDiskStorageAccountTypesValues() {
```
vkamra, line 492: @bathina2 - what's the behavior if we don't find a match - do we need to return an error?

bathina2: The default is to take the type from the from the snapshot, which is what it was doing originally. You can pass it a sku.Name type and it will change the resulting volume to the new type.

bathina2: I made it a lenient check. Unless you prefer I make it strict?

PR #655, opened 2020-04-19
Add region to Get/DeleteBucket

tdmanv:
## Change Overview

We cannot discover the region for buckets in AWS S3 that use non-standard regions like `us-gov-west-1`. This PR adds region to Get/Delete

## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test

## Test Plan

Ran a manual test with a bucket in `us-gov-west-1`:

https://gist.github.com/844c65e12f4ee9484522292f51c53807

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

PR #657, opened 2020-04-21
Specify ObjectStore region via ProviderConfig

tdmanv:
## Change Overview

We cannot discover the region for buckets in AWS S3 that use non-standard regions like us-gov-west-1. This PR adds region to the provider config

This is an alternative to https://github.com/kanisterio/kanister/pull/655

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/objectstore/bucket.go
+++ b/pkg/objectstore/bucket.go
@@ -47,8 +47,8 @@ type bucket struct {
 }
 
 // CreateBucket creates the bucket. Bucket naming rules are provider dependent.
-func (p *provider) CreateBucket(ctx context.Context, bucketName, region string) (Bucket, error) {
-	location, err := getStowLocation(ctx, p.config, p.secret, region)
+func (p *provider) CreateBucket(ctx context.Context, bucketName string) (Bucket, error) {
+	location, err := getStowLocation(ctx, p.config, p.secret)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


pkg/objectstore/bucket.go
```diff
@@ -131,38 +131,40 @@ func (p *provider) ListBuckets(ctx context.Context) (map[string]Bucket, error) {
...
-	var region string
-	if hostEndPoint == "" {
+	region := p.config.Region
+	if region == "" {
 		var err error
 		region, err = p.getRegionForBucket(ctx, bucketName)
 		if err != nil {
 			return nil, errors.Wrapf(err, "could not get region for bucket %s", bucketName)
 		}
+	}
```
vkamra, line 162: Do you need to set `p.config.Region = region` here?

tdmanv: Yes, thank you. This should probably be hoisted to the provider's initialization so we don't need to do it every time

tdmanv:
Let's not merge til we have the k10 side as well

pkg/objectstore/bucket.go
```diff
@@ -184,8 +186,13 @@ func (p *s3Provider) GetBucket(ctx context.Context, bucketName string) (Bucket,
 }
 
 func (p *s3Provider) DeleteBucket(ctx context.Context, bucketName string) error {
-	region, _ := p.getRegionForBucket(ctx, bucketName)
-	location, err := getStowLocation(ctx, p.config, p.secret, region)
+	cfg := p.config
+	if cfg.Region == "" {
+		// We swalllow this error because region may not be required. If it is,
+		// we'll fail in the next few lines.
+		cfg.Region, _ = p.getRegionForBucket(ctx, bucketName)
```
DeepikaDixit, line 193: We can log this error?

tdmanv: We should. I"ll followup

PR #662, opened 2020-04-26
Use ProviderConfig to check for bucket region.

tdmanv:
## Change Overview

GetRegion call was not honoring settings in the profile

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

Also tested manually in govcloud and minio w/ regions and minio w/o regions

```
go test -v ./pkg/objectstore -ch
eck.v    -check.f TestGetRegionForBucket
=== RUN   Test
PASS: bucket_test.go:129: BucketSuite.TestGetRegionForBucket    2.171s
OK: 1 passed
--- PASS: Test (2.17s)
PASS
ok      github.com/kanisterio/kanister/pkg/objectstore  2.179s

```
- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/objectstore/bucket.go
+++ b/pkg/objectstore/bucket.go
@@ -213,10 +213,10 @@ func (p *s3Provider) DeleteBucket(ctx context.Context, bucketName string) error 
 	return location.RemoveContainer(bucketName)
 }
 
-// GetRegionForBucketreturns the region for a particular bucket. It does not
-// the region set in the provider config is used as a hint, but may differ than
-// the actual region of the bucket. If the bucket does not have a region, then
-// the return value will be "",
+// GetRegionForBucket returns the region for a particular bucket. It does not
+// use the region set in the provider config is used as a hint, but may differ
+// from the actual region of the bucket. If the bucket does not have a region,
+// then the return value will be "".
 func (p *s3Provider) GetRegionForBucket(ctx context.Context, bucketName string) (string, error) {
 	cfg, r, err := awsConfig(ctx, p.config, *p.secret.Aws)
 	if err != nil {
--- a/pkg/objectstore/bucket.go
+++ b/pkg/objectstore/bucket.go
@@ -26,14 +26,15 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/graymeta/stow"
 	"github.com/pkg/errors"
+
+	"github.com/kanisterio/kanister/pkg/field"
+	"github.com/kanisterio/kanister/pkg/log"
 )
 
 var _ Provider = (*provider)(nil)
 
 // provider implements the Provider functionality
 type provider struct {
-	// e.g., s3-us-west-2.amazonaws.com
-	hostEndPoint string
 	// Object store information
 	config ProviderConfig
 	// Secret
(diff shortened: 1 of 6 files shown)


pkg/objectstore/bucket.go
```diff
@@ -210,9 +213,34 @@ func (p *s3Provider) DeleteBucket(ctx context.Context, bucketName string) error
...
+		return "", errors.Wrapf(err, "failed to create session, region = %s", r)
+	}
+	cli := s3.New(s)
+	if cli == nil {
+		return "", errors.New("failed to create s3 client")
+	}
+	gbli := &s3.GetBucketLocationInput{
+		Bucket: aws.String(bucketName),
+	}
+	gblo, err := cli.GetBucketLocation(gbli)
```
miquella, line 236: I worry about using `GetBucketLocation` over `GetBucketRegion` (from the `s3manager`) package.

`GetBucketLocation` is restricted to the owner of the bucket. This means that if a bucket is shared between two accounts, only the owning account will be able to make this call. Other accounts won't be able to determine the region for the bucket.

tdmanv:
Hold off merging until https://github.com/kanisterio/kanister/pull/663 and https://github.com/kanisterio/kanister/pull/664 go in

pkg/objectstore/bucket.go
```diff
@@ -223,3 +201,35 @@ func (p *s3Provider) getOrCreateBucket(ctx context.Context, bucketName string) (
...
+func s3Endpoint(c ProviderConfig) string {
+	if c.Endpoint != "" {
+		return c.Endpoint
+	}
+	r := defaultS3region
+	if c.Region != "" {
+		r = c.Region
+	}
+	return awsS3Endpoint(r)
+
```
tdmanv, line 224:
```suggestion
```

pkg/objectstore/bucket.go
```diff
@@ -145,74 +139,58 @@ func (p *provider) getOrCreateBucket(ctx context.Context, bucketName string) (Bu
...
-		hostEndPoint = awsS3Endpoint(cfg.Region)
-	}
-	bucket := &bucket{
-		directory:    dir,
-		container:    c,
-		location:     location,
-		hostEndPoint: path.Join(hostEndPoint, c.ID()),
-		region:       cfg.Region,
+// GetRegionForBucket returns the region for a particular bucket. It does not
+// use the region set in the provider config is used as a hint, but may differ
```
pavannd1, line 174: [minor] This sentence doesn't seem right
```suggestion
// GetRegionForBucket returns the region for a particular bucket.
// The region set in the provider config is used as a hint, but may differ
```

pkg/objectstore/objectstore_test.go
```diff
@@ -99,8 +100,9 @@ func (s *ObjectStoreProviderSuite) initProvider(c *C, region string) {
 	ctx := context.Background()
 	var err error
 	pc := ProviderConfig{
-		Type:   s.osType,
-		Region: region,
+		Type:     s.osType,
+		Region:   region,
+		Endpoint: s.endpoint,
```
pavannd1, line 105: We haven't set `s.endpoint`. Is it expected to be `""`?

tdmanv: I added this since I used it to test profiles with endpoints. I was planning on following up with adding more buckets to test with, but most other providers have slightly different behavior WRT directories and so not all these tests pass.

pkg/objectstore/bucket.go
tdmanv: sorry, I missed this comment. I'll address in a followup

PR #664, opened 2020-04-29
Use s3manager.GetBucketRegion over GetBucketLocation

tdmanv:
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/objectstore/bucket.go
+++ b/pkg/objectstore/bucket.go
@@ -21,9 +21,8 @@ import (
 	"fmt"
 	"path"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/graymeta/stow"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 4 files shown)


pkg/objectstore/bucket.go
```diff
@@ -191,21 +190,7 @@ func s3BucketRegion(ctx context.Context, cfg ProviderConfig, sec Secret, bucketN
...
-	}
-	gblo, err := cli.GetBucketLocation(gbli)
-	if err != nil {
-		return "", errors.Wrap(err, "failed to get bucket location")
-	}
-	if gblo.LocationConstraint != nil {
-		return *gblo.LocationConstraint, nil
-	}
-	return "", nil
+	return s3manager.GetBucketRegion(ctx, s, bucketName, cfg.Region)
```
vkamra, line 193: What's the reasoning behind switching to `GetBucketRegion`?

tdmanv: They use slightly different auth schemes. I think I'd prefer to move to GetBucketLocation, but there are several tests that assume our behavior is based on GetBucketRegion. 

@miquella was also concerned about cross region support using GetBucketLocation. We should probably try the instructions in this doc to verify: https://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example2.html

PR #676, opened 2020-05-16
Blueprint delete on Namespace object: helm apps

ewhamilton:
## Change Overview

Update blueprints for example helm apps to expect Namespace object for delete action.

## Pull request type

- [x] :bug: Bugfix

### DIFF
--- a/examples/helm/kanister/kanister-elasticsearch/kanister/elasticsearch-blueprint.yaml
+++ b/examples/helm/kanister/kanister-elasticsearch/kanister/elasticsearch-blueprint.yaml
@@ -47,15 +47,15 @@ actions:
           host_name="{{ .Object.metadata.labels.release }}-kanister-elasticsearch-client.{{ .StatefulSet.Namespace }}.svc.cluster.local"
           kando location pull --profile '{{ toJson .Profile }}' --path '{{ .ArtifactsIn.cloudObject.KeyValue.backupLocation }}' - | gunzip | elasticdump --bulk=true --input=$ --output=http://${host_name}:9200
   delete:
-    type: StatefulSet
+    type: Namespace
     inputArtifactNames:
     - cloudObject
     phases:
     - func: KubeTask
       name: deleteFromObjectStore
       args:
-        namespace: "{{ .StatefulSet.Namespace }}"
-        image: "kanisterio/es-sidecar:0.29.0"
+        namespace: "{{ .Namespace.Name }}"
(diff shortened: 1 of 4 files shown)


examples/helm/kanister/kanister-elasticsearch/kanister/elasticsearch-blueprint.yaml
```diff
@@ -47,15 +47,15 @@ actions:
...
     inputArtifactNames:
     - cloudObject
     phases:
     - func: KubeTask
       name: deleteFromObjectStore
       args:
-        namespace: "{{ .StatefulSet.Namespace }}"
-        image: "kanisterio/es-sidecar:0.29.0"
+        namespace: "{{ .Namespace.Name }}"
+        image: kanisterio/kanister-tools:0.29.0
```
ewhamilton, line 58: Didn't have to change the image, but figured this would be more robust if es-sidecar image ever went away.

examples/helm/kanister/kanister-mongodb-replicaset/kanister/mongodb-blueprint.yaml
```diff
@@ -58,16 +58,15 @@ actions:
...
     phases:
-    - func: KubeExec
+    - func: KubeTask
       name: deleteFromBlobStore
       args:
-        namespace: "{{ .StatefulSet.Namespace }}"
-        pod: "{{ index .StatefulSet.Pods 0 }}"
-        container: kanister-sidecar
+        namespace: "{{ .Namespace.Name }}"
+        image: kanisterio/kanister-tools:0.29.0
```
ewhamilton, line 69: Image wasn't otherwise in this file so I used kanister-tools image.  Also advantage for retiring over time.

ewhamilton:
Tests fail because trying to spin up pod in wrong namespace-- the app namespace, not the controller one. Or error is wrong and problem is not waiting for the actionset to complete before tearing down namespace.

TIme will tell.

ewhamilton:
@tdmanv @vkamra I am baffled at how the namespace of the KubeTask appears to be getting cross-wired with a different test namespace.  See the failing output (with test logs enabled).

From the output it appears that Travis is running multiple test Suite's concurrently, but I'm not entirely sure about the process/goroutine hierarchy. Still it seems that would mess up much worse if that weren't generally working right.

Pointing out my obvious error is welcome.

viveksinghggits:
I am trying to repro for ElasticSearch locally with the updated blueprint.

PR #678, opened 2020-05-18
Fix race in setting controller namespace

tdmanv:
## Change Overview

* Uses a well known namespace to run the controller
* Delete/recreates the namespace at test startup

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Test Plan

Fixes integration test

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/testing/integration_test.go
+++ b/pkg/testing/integration_test.go
@@ -18,13 +18,16 @@ package testing
 import (
 	"context"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
 	. "gopkg.in/check.v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
 	crv1alpha1 "github.com/kanisterio/kanister/pkg/apis/cr/v1alpha1"
 	"github.com/kanisterio/kanister/pkg/app"
@@ -75,6 +78,13 @@ func newSecretProfile() *secretProfile {
 	}
 }
 
+var (
+	setUpOnce = sync.Once{}
+	setUpDone = make(chan struct{})
+)
+
+const controllerNamespace = "integration-test-controller"
+
 func (s *IntegrationSuite) SetUpSuite(c *C) {
 	ctx := context.Background()
 	ctx, s.cancel = context.WithCancel(ctx)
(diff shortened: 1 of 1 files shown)


pkg/testing/integration_test.go
```diff
@@ -88,10 +98,42 @@ func (s *IntegrationSuite) SetUpSuite(c *C) {
...
+	setUpOnce.Do(func() {
+		defer close(setUpDone)
+		resetNamespace(ctx, c, s.cli)
+		runController(ctx, c, cfg)
+	})
+	<-setUpDone
+}
+
+func resetNamespace(ctx context.Context, c *C, cli kubernetes.Interface) {
+	// Try to delete namespace and wait til it doesn't exist.
```
pavannd1, line 110:
```suggestion
	// Try to delete namespace and wait until it doesn't exist.
```
[nit]

pkg/testing/integration_test.go
```diff
@@ -75,6 +78,13 @@ func newSecretProfile() *secretProfile {
 	}
 }
 
+var (
+	setUpOnce = sync.Once{}
+	setUpDone = make(chan struct{})
+)
+
+const controllerNamespace = "integration-test-controller"
```
ewhamilton, line 86: You could have controllerNamespace be a variable and tag a rand(6) on the end if you wish.  Just in case there are ever multiple jobs running integration test in parallel.

Or keep it this way.

ewhamilton: That would require a cleanup in Teardown suite with no easy way to know when really done.  This works as is.

ewhamilton:
Approved but added hold off tag so it didn't auto merge.  Remove when you want.

pkg/testing/integration_test.go
ewhamilton: Oh -- it's wouldn't be that hard.  We could move namespace create and cleanup to function `Test` before and after the invocation of `TestingT()`.

That would have advantage of tearing down namespace and cleaning up after itself.

Your call.

pkg/testing/integration_test.go
```diff
@@ -75,6 +78,13 @@ func newSecretProfile() *secretProfile {
...
+
 func (s *IntegrationSuite) SetUpSuite(c *C) {
 	ctx := context.Background()
 	ctx, s.cancel = context.WithCancel(ctx)
```
ewhamilton, line 90: @tdmanv I think you've got to manage the cancel for the controller outside of the Suite construct.

More I think about it the cleaner it is to have a setup and teardown func in this file that is called by the Test func in `testing_test.go`.

pkg/testing/integration_test.go
tdmanv: Deleting it would be difficult - I would need to add a wait group and keep track of the number of suites that have run. Here I just clean up before running.

PR #679, opened 2020-05-19
Use single controller for concurrent suites in integration_test

ewhamilton:
## Change Overview

Fix concurrency issues with concurrent suites and multiple Kanister controller watches.

Alternative approach to fix race in Suite setup.

## Pull request type

- [x] :bug: Bugfix

### DIFF
--- a/pkg/testing/testing_test.go
+++ b/pkg/testing/testing_test.go
@@ -15,14 +15,14 @@
 package testing
 
 import (
-	"testing"
+	test "testing"
 
-	. "gopkg.in/check.v1"
+	check "gopkg.in/check.v1"
 )
 
 // Hook up gocheck into the "go test" runner.
-func Test(t *testing.T) {
+func Test(t *test.T) {
 	integrationSetup(t)
-	TestingT(t)
+	check.TestingT(t)
 	integrationCleanup(t)
 }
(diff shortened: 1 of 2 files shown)


ewhamilton:
Well, looks like there are some packaging issues.  I look forward to seeing the solution in the morning :).

pkg/testing/testing_test.go
```diff
@@ -21,4 +21,8 @@ import (
 )
 
 // Hook up gocheck into the "go test" runner.
-func Test(t *testing.T) { TestingT(t) }
+func Test(t *testing.T) {
+	integrationSetup(t)
```
PrasadG193, line 25: We use build tag in integration_test.go, so if we want to use functions defined in that file, we need to add the `integration` tag in build. The reason was to separate out  integration test and unit test execution 
https://github.com/kanisterio/kanister/pull/679/files#diff-3c0035a2cccefd1a523cfcf89d924e62R1

ewhamilton: Good point.  Didn't notice comment before pushing iteration.  Will think a bit after breakfast.

ewhamilton:
My apologies to anyone who has followed all the compile errors-- I'll set up my kanister build environment before I next do anything on kanister.

PR #680, opened 2020-05-19
Create controller ns before executing any test

viveksinghggits:
## Change Overview

Create controller ns before executing any test and then use that for all the apps.
Highly inspired by @ewhamilton's PR

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/testing/integration_test.go
+++ b/pkg/testing/integration_test.go
@@ -17,7 +17,6 @@ package testing
 
 import (
 	"context"
-	"os"
 	"time"
 
 	"github.com/pkg/errors"
@@ -44,7 +43,6 @@ import (
 const (
 	// appWaitTimeout decides the time we are going to wait for app to be ready
 	appWaitTimeout = 3 * time.Minute
-	controllerSA   = "default"
 )
 
 type secretProfile struct {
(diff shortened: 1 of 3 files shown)


pkg/testing/integration_test.go
```diff
@@ -123,7 +124,7 @@ func (s *IntegrationSuite) TestRun(c *C) {
 	c.Assert(err, IsNil)
 
 	// Set Controller namespace and service account
-	os.Setenv(kube.PodNSEnvVar, s.namespace)
+	os.Setenv(kube.PodNSEnvVar, consts.ControllerNS)
```
ewhamilton, line 127: Use of `os.Setenv` within separate goroutines is problematic.  It is likely safe now because the calls are all setting the same value. For the sake of maintainability it would be better to move these 2 lines to the `integrationSetup()` function in the other file.

ewhamilton:
This approach will still create multiple controller watchers-- one per suite.  They'll either all be operating on the same namespace or on separate namespaces (haven't followed all the paths). That might work, but it  really doesn't mimic the normal runtime environment with a single namespace for controller and actionsets.

pkg/testing/integration_test.go
viveksinghggits: I have addressed this, but as @tdmanv  suggested, I have to make some improvements before be merge this.

PR #683, opened 2020-05-26
Updating tutorial.rst to allow users using other S3 like object storage other than AWS S3.

tellesnobrega:
Since we can't assume that people trying this tutorial are using AWS S3 and this is not necessarily true. By adding the --endpoint we force the command to pull the endpoint from the profile.

## Change Overview

Updating the tutorial to allow people trying with other S3 object storage other than AWS S3.

## Pull request type

- [X] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -321,9 +321,7 @@ data. We can use the templating engine ``toString`` and ``quote`` functions, cou
 ::
 
 **Note**
-
 If you are not using AWS S3 as the Object Storage, you need to add ``--endpoint {{ .Profile.Location.Endpoint }}`` to the end of the aws command so it can connect to the propper S3 object storage.
-
 
 For more on this templating, see :ref:`templates`
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -318,12 +318,12 @@ variables. Because we now have access to the bucket in the ConfigMap, we can
 also push the log to S3. In this Secret, we store the credentials as binary
 data. We can use the templating engine ``toString`` and ``quote`` functions, courtesy of sprig.
 
-___
+```
 **Note**
 
 If you are not using AWS S3 as the Object Storage, you need to add ``--endpoint {{ .Profile.Location.Endpoint }}``
 to the end of the aws command so it can connect to the propper S3 object storage.
-___
+```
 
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -317,6 +317,8 @@ section. We can then consume it through templates and assign it to bash
 variables. Because we now have access to the bucket in the ConfigMap, we can
 also push the log to S3. In this Secret, we store the credentials as binary
 data. We can use the templating engine ``toString`` and ``quote`` functions, courtesy of sprig.
+If you are not using AWS S3 as the Object Storage, you need to add ``--endpoint {{ .Profile.Location.Endpoint }}``
+to the end of the aws command so it can connect to the propper S3 object storage.
```
viveksinghggits, line 321: If we are changing command just for the minio storage, let's have this after the blueprint manifest as a `note`, and there we will mention the updated `aws` command.

docs/tutorial.rst
```diff
@@ -348,7 +350,8 @@ For more on this templating, see :ref:`templates`
             - |
               AWS_ACCESS_KEY_ID={{ .Secrets.aws.Data.aws_access_key_id | toString }}         \
               AWS_SECRET_ACCESS_KEY={{ .Secrets.aws.Data.aws_secret_access_key | toString }} \
-              aws s3 cp /var/log/time.log {{ .ConfigMaps.location.Data.path | quote }}
+              aws s3 cp /var/log/time.log {{ .ConfigMaps.location.Data.path | quote }} \
+              --endpoint {{ .Profile.Location.Endpoint }}
```
viveksinghggits, line 354: If we are changing the command just for minio we should have updated command separately in Note section.

docs/tutorial.rst
pavannd1:
```suggestion
If you are not using AWS S3 as the Object Storage, you need to add ``--endpoint {{ .Profile.Location.Endpoint }}``
to the end of the aws command so it can connect to the appropriate S3 object storage.
```
Agree with @viveksinghggits about having a separate Note. 
This will hold true for all non-aws S3 compatible storage.

PR #693, opened 2020-06-02
Update go to 1.14.2

tdmanv:
## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

PR #648, opened 2020-04-14
Bump Go to 1.14.2 in build image

PrasadG193:
## Change Overview

Upgrade Go to 1.14.2 in build image

## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

PR #694, opened 2020-06-03
Virtual file system for streaming data

pavannd1:
## Change Overview

- Mock file system layer for streaming data using kopia.

TODO: 
- Add owner information to file entry

## Pull request type

- [x] :sunflower: Feature

## Test Plan

- [x] :muscle: Manual


pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Subdir finds a subdirectory with the given name.
+func (d *Directory) Subdir(name ...string) *Directory {
+	i := d
+
+	for _, n := range name {
+		i2 := i.children.FindByName(n)
+		if i2 == nil {
+
```
pavannd1, line 80:
```suggestion
```

pavannd1:
Some ambiguous imports. Fixing locally.

pavannd1:
Fixed gomod 🚀  
Looks like it will require 1.14 from #693 or #648

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	d.children = append(d.children, e)
+	d.children.Sort()
+	return nil
+}
+
+func (d *Directory) resolveSubdir(name string) (parent *Directory, leaf string, err error) {
+	parts := strings.Split(name, "/")
+	for _, n := range parts[0 : len(parts)-1] {
+		if d, err = d.Subdir(n); err != nil {
```
carlbraganza, line 152: Do not override the receiver variable (d)! May work but its not a normal thing to do and any linter would complain!

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	d.children = append(d.children, e)
+	d.children.Sort()
+	return nil
+}
+
+func (d *Directory) resolveSubdir(name string) (parent *Directory, leaf string, err error) {
+	parts := strings.Split(name, "/")
+	for _, n := range parts[0 : len(parts)-1] {
```
carlbraganza, line 151: See my comment in resolveAndAddSubdir on iterating this range.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+	}
+
+	err = d.addChild(subdir)
+
+	return subdir, err
+}
+
+// Subdir finds a subdirectory with the given name.
+func (d *Directory) Subdir(name ...string) (*Directory, error) {
```
carlbraganza, line 84: Why zero or more names?  The doc comment clearly indicates a single name.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	if strings.Contains(e.Name(), "/") {
+		return errors.New("Failed to add child entry: name cannot contain '/'")
+	}
+
+	d.children = append(d.children, e)
+	d.children.Sort()
+	return nil
+}
+
+func (d *Directory) resolveSubdir(name string) (parent *Directory, leaf string, err error) {
```
carlbraganza, line 149: I'm not sure from its usage what this sub is supposed to be doing. Please add a doc comment.

pkg/virtualfs/dir_entry.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// dirEntry is mock in-memory implementation of a directory dirEntry
+type dirEntry struct {
+	name    string
+	mode    os.FileMode
+	size    int64
+	modTime time.Time
+	owner   fs.OwnerInfo
+}
```
carlbraganza, line 31: Please put type assertion vars for the Kopia interfaces dirEntry is supposed to satisfy. In this case I think it is fs.Entry:
```
var _ fs.Entry = (*dirEntry)(nil)
```

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Directory is a mock in-memory implementation of kopia's fs.Directory
+type Directory struct {
+	dirEntry
+
+	children     fs.Entries
+	readdirError error
+	onReaddir    func()
+}
+
```
carlbraganza, line 35: Suggestion:
```
var _ fs.Directory = (*Directory)(nil)
```

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	d, name, err = d.resolveAndAddSubdir(name)
+	if err != nil {
+		return nil, err
+	}
+
+	file = &File{
+		dirEntry: dirEntry{
+			name: name,
+			mode: permissions,
+			// TODO: Add owner information
```
carlbraganza, line 52: Only if you need it - if owner / permission are meaningless for your particular case then you can ignore them.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	return d, parts[len(parts)-1], nil
+}
+
+func (d *Directory) resolveAndAddSubdir(name string) (parent *Directory, leaf string, err error) {
+	parts := strings.Split(name, "/")
+	for _, n := range parts[0 : len(parts)-1] {
+		if d, err = d.AddDir(n, 0777); err != nil {
+			return nil, "", errors.Wrap(err, fmt.Sprintf("Failed to add sub directory '%s'", n))
+		}
+	}
```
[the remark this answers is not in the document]
carlbraganza: Oh wait! You're overriding "d". My bad.

Please rewrite without overwriting the receiver variable!

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Summary returns summary of a directory.
+func (d *Directory) Summary() *fs.DirectorySummary {
+	return nil
+}
+
+// AddFileWithStreamSource adds a mock file with the specified name, permissions and source.
+func (d *Directory) AddFileWithStreamSource(name, sourceEndpoint string, permissions os.FileMode) (file *File, err error) {
+	d, name, err = d.resolveAndAddSubdir(name)
```
carlbraganza, line 43: Please do not override the receiver variable!

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	return append(fs.Entries(nil), d.children...), nil
+}
+
+func (d *Directory) addChild(e fs.Entry) error {
+	if strings.Contains(e.Name(), "/") {
+		return errors.New("Failed to add child entry: name cannot contain '/'")
+	}
+
+	d.children = append(d.children, e)
+	d.children.Sort()
```
carlbraganza, line 145: Pretty expensive sorting with each call just to use FindByName

pkg/virtualfs/dir_entry.go
julio-lopez: These are in the last file below.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/pkg/errors"
+)
+
+// Directory is a mock in-memory implementation of kopia's fs.Directory
+type Directory struct {
+	dirEntry
+
+	children     fs.Entries
+	readdirError error
+	onReaddir    func()
```
julio-lopez, line 33: is `onReaddir` needed?
```suggestion
```

pkg/virtualfs/virtualfs.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+
+	"github.com/kopia/kopia/fs"
+)
+
+// NewDirectory returns new mock directory.
+func NewDirectory(rootName string) *Directory {
+	return &Directory{
+		dirEntry: dirEntry{
+			name: rootName,
```
julio-lopez, line 27: This should validate / sanity check `rootName` to not contain a path separator for instance.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+
+	d.children = newChildren
+}
+
+// OnReaddir invokes the provided function on read.
+func (d *Directory) OnReaddir(cb func()) {
+	d.onReaddir = cb
+}
```
julio-lopez, line 119: ?
```suggestion
```

pkg/virtualfs/file_reader.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// ReaderSeekerCloser implements io.Reader, io.Seeker and io.Closer
+type ReaderSeekerCloser interface {
+	io.Reader
+	io.Seeker
+	io.Closer
+}
+
+type readerSeekerCloser struct {
```
julio-lopez, line 33: The struct names are a little confusing, they don't convey the intent.
Adding a couple of comments helps. At a later time, somebody else will try to understand what this is intended to do, and they do not have the same context you have in your mind right now.

```suggestion
// readSeekerWrapper adds a dummy Close method to a ReadSeeker
type readSeekerWrapper struct {
```

pkg/virtualfs/file_reader.go
```diff
@@ -0,0 +1,58 @@
...
+	io.ReadSeeker
+}
+
+func (c readerSeekerCloser) Close() error {
+	return nil
+}
+
+type readerCloser struct {
```
julio-lopez, line 41:
```suggestion
// readCloserWrapper adds a dummy Seek method to a readCloser
type readCloserWrapper struct {
```

pkg/virtualfs/file_reader.go
```diff
@@ -0,0 +1,58 @@
...
+	io.Reader
+	io.Seeker
+	io.Closer
+}
+
+type readerSeekerCloser struct {
+	io.ReadSeeker
+}
+
+func (c readerSeekerCloser) Close() error {
```
julio-lopez, line 37: The implementation could even attempt to call `Close()` on the `io.ReadSeeker` if it happens to have a `Close` method.

julio-lopez:
@pavannd1 High-level bit: basic unit tests for the directory and entry implementation functionality.

pkg/virtualfs/directory.go
julio-lopez: We may want to set the owner to the user used to run the tool. We expect that restore would run under the same user, so that would facilitate things.

pkg/virtualfs/directory.go
julio-lopez: Looking at the implementation, there is an implicit assumption for `ReadDir` to return sorted entries if `fs.ReadDirAndFindChild` is to be used.

julio-lopez: Also, if `d.children.FindByName` is to be used, then the entries must be sorted.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+			mode: permissions | os.ModeDir,
+		},
+	}
+
+	err = d.addChild(subdir)
+
+	return subdir, err
+}
+
+// Subdir finds a subdirectory with the given name.
```
julio-lopez, line 83: Is there an intended use for `Subdir` or simply copied over? If it is not needed, let's remove it.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+func (d *Directory) resolveAndAddSubdir(name string) (parent *Directory, leaf string, err error) {
+	parts := strings.Split(name, "/")
+	for _, n := range parts[0 : len(parts)-1] {
+		if d, err = d.AddDir(n, 0777); err != nil {
+			return nil, "", errors.Wrap(err, fmt.Sprintf("Failed to add sub directory '%s'", n))
+		}
+	}
+
+	return d, parts[len(parts)-1], nil
+}
```
julio-lopez, line 169: You could replace `resolveAndAddSubdir` with a more general `AddAllDirs` method with behavior similar to `mkdir -p`. That can be leveraged in the helper method mentioned above.

Then, you can also remove the `resolveSubdir` and `Subdir` methods.


```suggestion
// AddAllDirs creates under d all the necessary directories in pathname, similar
// to os.MkdirAll
func (d *Directory) AddAllDirs(pathname string, permissions os.FileMode) (parent *Directory, err error) {
	p, missing, err := d.resolveDirs(pathname)
	if err != nil {
		return nil, err
	}

	for _, n := range missing {
		if p, err = p.AddDir(n, permissions); err != nil {
			return nil, errors.Wrap(err, fmt.Sprintf("Failed to add sub directory '%s'", n))
		}
	}

	return p, nil
}

func (d *Directory) resolveDirs(pathname string) (parent *Directory, missing []string, err error) {
	if pathname == "" {
		return d, nil, nil
	}

	p := d
	parts := strings.Split(path.Clean(pathname), "/")

	for i, n := range parts {
		i2 := p.children.FindByName(n)
		if i2 == nil {
			return p, parts[i:], nil
		}

		if !i2.IsDir() {
			return nil, nil, errors.New(fmt.Sprintf("'%s' is not a directory in '%s'", n, p.Name()))
		}

		p = i2.(*Directory)
	}

	return p, nil, nil
}
```

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	subdir = &Directory{
+		dirEntry: dirEntry{
+			name: name,
+			mode: permissions | os.ModeDir,
+		},
+	}
+
+	err = d.addChild(subdir)
```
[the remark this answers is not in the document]
julio-lopez: @carlbraganza You are right. The validation should be done in `addChild` though.

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+func (d *Directory) Child(ctx context.Context, name string) (fs.Entry, error) {
+	return fs.ReadDirAndFindChild(ctx, d, name)
+}
+
+// Readdir gets the contents of a directory.
+func (d *Directory) Readdir(ctx context.Context) (fs.Entries, error) {
+	if d.readdirError != nil {
+		return nil, d.readdirError
+	}
+
```
julio-lopez, line 131: nuke
```suggestion
```

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+		source: func() (ReaderSeekerCloser, error) {
+			return streamReader(sourceEndpoint)
+		},
+	}
+
+	err = d.addChild(file)
+
+	return file, errors.Wrap(err, "Failed to add file")
+}
```
julio-lopez, line 62: This method should not be part of `Directory` from the API design point of view. It is cleaner to split this into separate methods to:
* Create a `File` with an endpoint / stream source. See comments in `file.go`
* Add subdirectories (akin to `mkdir -p`), that's `func (*Directory) AddAllDirs(...)` below.
* Package level helper to create a stream source and attach it to a directory.

```suggestion
// AddFileWithStreamSource adds a virtual file with the specified name, permissions and source.
func AddFileWithStreamSource(d *Directory, pathname, sourceEndpoint string, dirPermissions, filePermissions os.FileMode) (file *File, err error) {
	dir, name := path.Split(pathname)
	p, err := d.AddAllDirs(dir, dirPermissions)

	if err != nil {
		return nil, err
	}

	f := FileFromEndpoint(name, sourceEndpoint, filePermissions)
	if err := p.addChild(f); err != nil {
		return nil, errors.Wrap(err, "Failed to add file")
	}

	return f, nil
}
```

pkg/virtualfs/directory.go
```diff
@@ -0,0 +1,169 @@
...
+
+// AddDir adds a fake directory with a given name and permissions.
+func (d *Directory) AddDir(name string, permissions os.FileMode) (subdir *Directory, err error) {
+	d, name, err = d.resolveSubdir(name)
+	if err != nil {
+		return nil, err
+	}
+
```
julio-lopez, line 70: Remove sub-directory resolution here. It makes it for cleaner semantics and implementation.
Instead, provide the convenience `AddAllDirs` method below, which the caller could use if they wanted to create intermediate directories.
```suggestion
```

pkg/virtualfs/virtualfs.go
```diff
@@ -0,0 +1,37 @@
...
+
+package virtualfs
+
+import (
+	"os"
+
+	"github.com/kopia/kopia/fs"
+)
+
+// NewDirectory returns new mock directory.
```
julio-lopez, line 23:
```suggestion
// NewDirectory returns new virtual FS root directory.
```

julio-lopez:
Overall, LG.
Let's add unit tests to this PR to exercise the code paths.

pavannd1:
```
=== RUN   Test
START: directory_test.go:49: DirectorySuite.TestAddAllDirs
Add a directory - root/d1
Add third level directory - root/d1/d2
Add third/fourth level dirs - root/d1/d3/d4
Fail adding directory under a file - root/f1/d6
PASS: directory_test.go:49: DirectorySuite.TestAddAllDirs	0.000s

START: directory_test.go:31: DirectorySuite.TestAddDir
PASS: directory_test.go:31: DirectorySuite.TestAddDir	0.000s

START: directory_test.go:88: DirectorySuite.TestAddFile
Add file with stream source - root/f1
Add file with stream source at third level - root/d1/f2
Add file with content at third level - root/d2/f3
PASS: directory_test.go:88: DirectorySuite.TestAddFile	0.000s

START: virtualfs_test.go:29: VirtualFSSuite.TestNewDirectory
PASS: virtualfs_test.go:29: VirtualFSSuite.TestNewDirectory	0.000s

OK: 4 passed
--- PASS: Test (0.00s)
PASS
ok  	github.com/kanisterio/kanister/pkg/virtualfs	1.150s
```

pkg/virtualfs/directory_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	c.Log("Add a directory - root/d1")
+	subdir, err := sourceDir.AddAllDirs("d1/d2", defaultPermissions)
+	c.Assert(err, IsNil)
+	c.Assert(subdir.Name(), Equals, "d2")
+	d1, err := sourceDir.Subdir("d1")
+	c.Assert(err, IsNil)
+	c.Assert(d1, NotNil)
+
+	c.Log("Add third level directory - root/d1/d2")
```
julio-lopez, line 61: This log message seems misleading . The case below is checking whether adding already existing dirs succeeds (does not fail)

pkg/virtualfs/directory_test.go
```diff
@@ -0,0 +1,121 @@
...
+	c.Assert(err, IsNil)
+	c.Assert(d2, NotNil)
+
+	c.Log("Add third/fourth level dirs - root/d1/d3/d4")
+	subdir, err = sourceDir.AddAllDirs("d1/d3/d4", defaultPermissions)
+	c.Assert(err, IsNil)
+	c.Assert(subdir.Name(), Equals, "d4")
+	d3, err := d1.Subdir("d3")
+	c.Assert(err, IsNil)
+	c.Assert(d3, NotNil)
```
julio-lopez, line 75: Check the names and permissions as well?

pkg/virtualfs/directory_test.go
```diff
@@ -0,0 +1,121 @@
...
+	_, err = sourceDir.AddAllDirs("f1/d6", defaultPermissions)
+	c.Assert(err, NotNil)
+}
+
+func (s *DirectorySuite) TestAddFile(c *C) {
+	sourceDir, err := NewDirectory("root")
+	c.Assert(err, IsNil)
+
+	c.Log("Add file with stream source - root/f1")
+	f, err := AddFileWithStreamSource(sourceDir, "f1", "http://test-endpoint", defaultPermissions, defaultPermissions)
```
julio-lopez, line 93: It'd be good to have basic tests that exercise the streaming part to prevent regression.

PR #696, opened 2020-06-04
Add helpers for kando stream push

pavannd1:
## Change Overview

- Required functions to create a virtual fs and kopia snapshots

## Pull request type

- [x] :sunflower: Feature

## Test Plan

- [x] :muscle: Manual


pkg/stream/stream.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if ferr := rep.Flush(ctx); ferr != nil {
+		return errors.Wrap(ferr, "Failed to flush kopia repository")
+	}
+
+	var maybePartial string
+	if manifest.IncompleteReason != "" {
+		maybePartial = " partial"
+	}
+
```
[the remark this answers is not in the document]
julio-lopez: There is also a time bound for snapshots. We may be able to change the value to make it longer, but need to figure out what implications it has, since other components have assumptions about the maximum snapshot duration.

pavannd1:
Okay, now the CI is failing with go1.14 issue:
```
# github.com/kopia/kopia/internal/apiclient
/go/pkg/mod/github.com/kopia/kopia@v0.4.1-0.20200608151401-25934a544df4/internal/apiclient/apiclient.go:60:14: undefined: http.NewRequestWithContext
/go/pkg/mod/github.com/kopia/kopia@v0.4.1-0.20200608151401-25934a544df4/internal/apiclient/apiclient.go:142:48: http.DefaultTransport.(*http.Transport).Clone undefined (type *http.Transport has no field or method Clone)
note: module requires Go 1.14
```
This will be dependent on #693

pavannd1:
Fixed `go1.14` issue. This is ready for review

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+// OpenKopiaRepository connects to the kopia repository based on the config
+// NOTE: This assumes that `kopia repository connect` has been already run on the machine
+func OpenKopiaRepository(ctx context.Context, password string) (repo.Repository, error) {
+	if _, err := os.Stat(defaultConfigFileName()); os.IsNotExist(err) {
+		return nil, errors.New("Failed find kopia configuration file")
+	}
+
+	r, err := repo.Open(ctx, defaultConfigFileName(), password, &repo.Options{})
+	if os.IsNotExist(err) {
```
tdmanv, line 80: Kopia returns an error that matches `os.IsNotExist`?

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, errors.New("Failed to find kopia repository, use `kopia repository connect`")
```
tdmanv, line 81: This block is to return a more specific error, right?
```suggestion
		return nil, errors.New("Failed to find kopia repository, use `kopia repository create`")
```

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	return r, errors.Wrap(err, "Failed to open kopia repository")
+}
+
+func defaultConfigFileName() string {
+	return filepath.Join(os.Getenv("HOME"), ".config", "kopia", "repository.config")
```
tdmanv, line 88: Whats the chance that HOME could change? Should we set this to a well-known path?

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// SnapshotSource creates and uploads a kopia snapshot to the given repository
+func SnapshotSource(ctx context.Context, rep repo.Repository, u *snapshotfs.Uploader, sourceInfo snapshot.SourceInfo, rootDir fs.Entry) error {
+	fmt.Printf("Snapshotting %v ...\n", sourceInfo)
```
tdmanv, line 93: [minor] please use the logging library

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+	_, err = policy.ApplyRetentionPolicy(ctx, rep, sourceInfo, true)
+	if err != nil {
+		return errors.Wrap(err, "Failed to apply kopia retention policy")
+	}
+
+	if ferr := rep.Flush(ctx); ferr != nil {
+		return errors.Wrap(ferr, "Failed to flush kopia repository")
+	}
+
+	fmt.Printf("\nCreated snapshot with root %v and ID %v in %v\n", manifest.RootObjectID(), snapID, time.Since(t0).Truncate(time.Second))
```
tdmanv, line 128: please use the log library

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+			continue
+		}
+
+		if p.IncompleteReason == "" && (previousComplete == nil || p.StartTime.After(previousComplete.StartTime)) {
+			previousComplete = p
+		}
+	}
+
+	if previousComplete != nil {
+		result = append(result, previousComplete)
```
tdmanv, line 156: I found this a little confusing, but I see what you did. This is a slice with zero or one entries, which you later use as a variadic parameter

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+
+func defaultConfigFileName() string {
+	return filepath.Join(os.Getenv("HOME"), ".config", "kopia", "repository.config")
+}
+
+// SnapshotSource creates and uploads a kopia snapshot to the given repository
+func SnapshotSource(ctx context.Context, rep repo.Repository, u *snapshotfs.Uploader, sourceInfo snapshot.SourceInfo, rootDir fs.Entry) error {
+	fmt.Printf("Snapshotting %v ...\n", sourceInfo)
+
+	t0 := time.Now()
```
tdmanv, line 95: [nit] I usually find it easier to test if time is passed in.

pkg/stream/stream.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	manifest.Description = snapshotDescription
+
+	snapID, err := snapshot.SaveSnapshot(ctx, rep, manifest)
+	if err != nil {
+		return errors.Wrap(err, "Failed to save kopia manifest")
+	}
+
+	_, err = policy.ApplyRetentionPolicy(ctx, rep, sourceInfo, true)
```
tdmanv, line 119: is this just a kopia-ism? Does this need to be called?

pkg/stream/stream.go
pavannd1: `repo.Open` checks if the config file is present in `defaultConfigFileName` location and returns `NotExists` if the file is missing

pkg/stream/stream.go
pavannd1: Kopia commands have an option to pass a custom config path. I think we could add that too.
This path is the default linux path Kopia uses when the path is not passed in.

pkg/stream/stream.go
pavannd1: Not sure if I can use the logging library for these since this function runs as a part of `kando` inside the tools pod.
I will try it out, it might also need some changes to existing parsing functions.

pkg/stream/stream.go
pavannd1: Yes, we do need this if custom policy is setup for the repository. Default case has 2 snapshots for each frequency - 48, 14, 25 etc.

pavannd1:
I will merge and iterate

PR #716, opened 2020-07-06
Make caller package compatible with Go modules

miquella:
## Change Overview

When Kanister is used as a Go module, the "kanister" component of the path also includes the version number. The path stripping needs to account for this.

## Pull request type

- [X] :hamster: Trivial/Minor

## Issues

- N/A

## Test Plan

- [X] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/caller/frame.go
+++ b/pkg/caller/frame.go
@@ -14,6 +14,11 @@ type Frame struct {
 	Line     int
 }
 
+// kanisterPathRe matches the kanister component of the filesystem path
+//
+// For example:
+// /home/mark/src/kanister/pkg/kanister.go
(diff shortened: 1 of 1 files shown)


pkg/caller/frame.go
```diff
@@ -13,6 +14,8 @@ type Frame struct {
 	Line     int
 }
 
+var kanisterPathRe = regexp.MustCompile(`/kanister(@v[^/]+)?/`)
```
tdmanv, line 17: [nit] This is pretty readable, but with regex, it's often useful to provide an example of what you'll match against.

PR #738, opened 2020-08-19
Dynamically populate Zones/Regions for GCE

bathina2:
## Change Overview

Dynamically populate Zones/Regions for GCE

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

### DIFF
--- a/pkg/blockstorage/gcepd/gcepd.go
+++ b/pkg/blockstorage/gcepd/gcepd.go
@@ -548,7 +548,25 @@ func isNotFoundError(err error) bool {
 }
 
 func (s *gpdStorage) FromRegion(ctx context.Context, region string) ([]string, error) {
-	return staticRegionToZones(region)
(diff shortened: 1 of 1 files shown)


bathina2:
Need to test but I think this should do the trick.

pkg/blockstorage/gcepd/gcepd.go
```diff
@@ -548,7 +548,25 @@ func isNotFoundError(err error) bool {
 }
 
 func (s *gpdStorage) FromRegion(ctx context.Context, region string) ([]string, error) {
```
ntolia, line 550: How often is this function called? Does this need to be a one-time thing?

bathina2: It's called every restore I believe. I think that's fine.

PR #753, opened 2020-09-09
Register different versions (3.11, 4.4, 4.5) of kanister OCP apps

viveksinghggits:
## Change Overview

This PR adds different versions (3.11, 4.4, 4.5) of the kanister apps that were being tested against openshift cluster. 

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan
Run below command to run apps against 4.5 (default) cluster
```
make openshift-test
```
the logs can be found [here](https://gist.github.com/viveksinghggits/bee57ea544444205527acdee1338afc3).

```
make openshift-test ocp_version=3.11
```
[logs](https://gist.github.com/viveksinghggits/d5a013590d2d8be32ae5548197215a91)

```
make openshift-test ocp_version=4.4
```
[logs](https://gist.github.com/viveksinghggits/e2f32051f57f16c2bb02303404fbc559)

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/build/integration-test.sh
+++ b/build/integration-test.sh
@@ -27,7 +27,9 @@ TEST_OPTIONS="-tags=integration -timeout ${TEST_TIMEOUT} -check.suitep ${DOP}"
 # Regex to match apps to run in short mode
 SHORT_APPS="^PostgreSQL$|^PITRPostgreSQL|MySQL|Elasticsearch|^MongoDB$"
 # OCAPPS has all the apps that are to be tested against openshift cluster
-OC_APPS="MysqlDBDepConfig|MongoDBDepConfig|PostgreSQLDepConfig"
+OC_APPS3_11="MysqlDBDepConfig$|MongoDBDepConfig$|PostgreSQLDepConfig$"
+OC_APPS4_4="MysqlDBDepConfig4_4|MongoDBDepConfig4_4|PostgreSQLDepConfig4_4"
+OC_APPS4_5="MysqlDBDepConfig4_5|MongoDBDepConfig4_5|PostgreSQLDepConfig4_5"
 
 check_dependencies() {
     # Check if minio is already deployed
@@ -49,7 +51,7 @@ Usage: ${0} <app-type>
 Where app-type is one of [short|all]:
   short: Runs e2e integration tests for part of apps
   all: Runs e2e integration tests for all apps
-  oc: Runs e2e integration tests for apps that are to be tetsed against openshift cluster
+  openshift ocp_version=<ocp_version>: Runs e2e integration tests for apps that are to be tetsed against openshift cluster, OCP version can be provided using ocp_version argument
 OR
   You can also provide regex to match apps you want to run.
 EOM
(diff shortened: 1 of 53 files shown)


pkg/testing/integration_register.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/kanisterio/kanister/pkg/app"
 )
 
+const (
+	TemplateVersionOCP3_11 = "release-3.11"
+	TemplateVersionOCP4_4  = "release-4.4"
+	TemplateVersionOCP4_5  = "release-4.5"
+)
+
```
viveksinghggits, line 29: Make this a type.

build/integration-test.sh
```diff
@@ -27,7 +27,9 @@ TEST_OPTIONS="-tags=integration -timeout ${TEST_TIMEOUT} -check.suitep ${DOP}"
 # Regex to match apps to run in short mode
 SHORT_APPS="^PostgreSQL$|^PITRPostgreSQL|MySQL|Elasticsearch|^MongoDB$"
 # OCAPPS has all the apps that are to be tested against openshift cluster
-OC_APPS="MysqlDBDepConfig|MongoDBDepConfig|PostgreSQLDepConfig"
+OC_APPS3_11="MysqlDBDepConfig$|MongoDBDepConfig$|PostgreSQLDepConfig$"
+OC_APPS4_4="MysqlDBDepConfig4_4|MongoDBDepConfig4_4|PostgreSQLDepConfig4_4"
+OC_APPS4_5="MysqlDBDepConfig4_5|MongoDBDepConfig4_5|PostgreSQLDepConfig4_5"
```
PrasadG193, line 32: I think we can just create 2 groups - 3.x and 4.x.
I think db template must be compatible across minor releases.

pkg/app/utils.go
```diff
@@ -21,16 +21,25 @@ import (
...
 	PersistentStorage storage = "persistent" // nolint:varcheck
 
 	// EphemeralStorage can be used if we don't want to deploy database with Persistent
 	EphemeralStorage storage = "ephemeral"
+	// TemplateVersionOCP3_11 stores version of db template 3.11
+	TemplateVersionOCP3_11 DBTemplate = "release-3.11"
+	// TemplateVersionOCP4_4 stores version of db template 4.4
+	TemplateVersionOCP4_4 DBTemplate = "release-4.4"
+	// TemplateVersionOCP4_5 stored version of db template 4.5
+	TemplateVersionOCP4_5 DBTemplate = "release-4.5"
```
PrasadG193, line 35: Can we make it more generic. So version passed with `./build/integration-test.sh openshift <version>` could be any string. And we use that string to generate `dbTemplateURI`, if version is invalid, the test will anyway fail. This saves our efforts of maintaining support for several versions.

viveksinghggits: I dont think we should allow any version to be passed to `integration-test.sh` because 
1. It's better that we let someone know that this version is not supported than let him wait for the time test is running and then after test fails he will have to figure out that this version is not supported.
2. This will help us as well, in order to keep track of the version that we have tested with and are supported.
what do you think

PrasadG193: Sounds good

PrasadG193 (approved):

PR #798, opened 2020-10-20
Bypass bash -c and run kando directly in the pod 

viveksinghggits:
## Change Overview

- Remove bash and run the `kando` command directly in the pod

## Pull request type

- [x] :bug: Bugfix

## Test Plan

```
» cd pkg/kando
» go test -check.f "StreamPushTestSuite"
OK: 1 passed
PASS
ok  	github.com/kanisterio/kanister/pkg/kando	0.016s
```

- [x] :muscle: Manual
- [x] :zap: Unit test


pkg/kando/stream_push_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (s *StreamPushTestSuite) TestEnclosePassword(c *C) {
+
+	for _, tc := range []struct {
+		input, output string
+	}{
+		// we dont handle back ticks yet
+		{
+			input:  "this-is3543%$%#$#()*&)~~`-dummy-pass4534",
+			output: "\"this-is3543%$%#$#()*&)~~\\`-dummy-pass4534\"",
```
PrasadG193, line 19: Does this cause a mismatch between the original password and the transformed password? What will happen if a repo is created with an older format of password and after an upgrade, we try to connect to the repo with a transformed password?

viveksinghggits: @PrasadG193, 
the password that would eventually use the password that user provides. We are not actually transforming the password. I still have to test the restore though.

viveksinghggits: Restore works as expected.

pkg/kando/stream_push.go
```diff
@@ -95,3 +97,23 @@ func GenerateStreamPushCommand(configFile, dirPath, filePath, password, sourceEn
...
+	// add escape char for "
+	reg, err := regexp.Compile("\"")
+	if err != nil || len(password) == 0 {
+		return password
+	}
+
+	op := reg.ReplaceAllString(password, "\\\"")
+
+	// add escape char for `
+	reg, err = regexp.Compile("`")
```
viveksinghggits, line 113: refactor this handle `"` and ` in single compile.

julio-lopez (changes requested):
Directly execute `kando` and bypass the shell.

viveksinghggits:
> Directly execute `kando` and bypass the shell.

Directly executing kando didn't work, I will update details in the ticket that we have.

pkg/kando/stream_push_test.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		{
+			input:  "12345",
+			output: `"12345"`,
+		},
+		{
+			input:  "this-is-dummy-pass",
+			output: "\"this-is-dummy-pass\"",
```
pavannd1, line 26:
```suggestion
			input:  "this-is-test-pass",
			output: "\"this-is-test-pass\"",
```
```suggestion
			input:  "this-is-dummy-pass",
			output: "\"this-is-dummy-pass\"",
```

pkg/kando/stream_push_test.go
```diff
@@ -0,0 +1,48 @@
...
+var _ = Suite(&StreamPushTestSuite{})
+
+func (s *StreamPushTestSuite) TestEnclosePassword(c *C) {
+	for _, tc := range []struct {
+		input, output string
+	}{
+		// we dont handle back ticks yet
+		{
+			input:  "this-is3543%$%#$#()*&)~~`-dummy-pass4534",
+			output: "\"this-is3543%$%#$#()*&)~~\\`-dummy-pass4534\"",
```
pavannd1, line 18:
```suggestion
			input:  "this-is3543%$%#$#()*&)~~`-test-pass4534",
			output: "\"this-is3543%$%#$#()*&)~~\\`-test-pass4534\"",
```
```suggestion
			input:  "this-is3543%$%#$#()*&)~~`-dummy-pass4534",
			output: "\"this-is3543%$%#$#()*&)~~\\`-dummy-pass4534\"",
```

pkg/kando/stream_push.go
```diff
@@ -95,3 +97,23 @@ func GenerateStreamPushCommand(configFile, dirPath, filePath, password, sourceEn
...
 }
+
+// enclosePassword adds escape char \ before every double quote " or backtick ` in the password
+// and then encloses the result between " (double quote) to handle special chars in the password
+func enclosePassword(password string) string {
+	// add escape char for "
+	reg, err := regexp.Compile("\"")
+	if err != nil || len(password) == 0 {
+		return password
+	}
```
pavannd1, line 108:
```suggestion
	if password == "" {
		return password
	}

	// add escape char for "
	reg, err := regexp.Compile("\"")
	if err != nil {
		return password
	}
```
Return early to avoid regexp compile.

pavannd1:
An alternative would be to completely remove `--password` flag from the command helper. We may not require it after all!

pavannd1 (approved):
LGTM

julio-lopez (approved):
🎊 🎉 🎈 🏆

PR #800, opened 2020-10-26
Split Initialize function in apps from Reset, allow aws region to be passed to aws-rds app

ankitjain235:
## Change Overview

Fix dump based rds-postgres blueprint for failing delete action and update the app/blueprint to be in sync with latest changes.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

`$ make integration-test rds-postgres-dump`


pkg/testing/integration_test.go
```diff
@@ -259,8 +259,8 @@ func (s *IntegrationSuite) TestRun(c *C) {
 		err = a.Reset(ctx)
 		c.Assert(err, IsNil)
 
-		count, err := a.Count(ctx)
-		c.Assert(err, IsNil)
+		count, _ := a.Count(ctx)
+		// c.Assert(err, IsNil)
```
PrasadG193, line 263: missed while cleanup?

ankitjain235: After reset call, the schema won't be present in cases where we are not using default schema. There, the Count would always return 0, err. So removed this Assert.

pkg/app/mysql.go
```diff
@@ -195,14 +195,18 @@ func (mdb *MysqlDB) Reset(ctx context.Context) error {
 		return errors.Wrapf(err, "Error while dropping the mysql table: %s", stderr)
 	}
 
+	log.Print("Reset of the application was successful.", field.M{"app": mdb.name})
+	return nil
+}
+
+// Initialize is used initialize the database or create schema
```
viveksinghggits, line 202:
```suggestion
// Initialize is used to initialize the database or create schema
```

viveksinghggits: and everywhere.

pkg/testing/integration_test.go
PrasadG193: why does it return an error? We should not get an error if count is 0.

viveksinghggits (approved):

pkg/app/mysql.go
ankitjain235: Done

pkg/testing/integration_test.go
ankitjain235: Earlier we used to initialize the schema in Reset call so count would always be successful. Now, in cases where we there is no table, the count query would fail but we still return 0 there.

PrasadG193: Ideally, Count() should not return an error because having no data present doesn't mean the count query has failed. It could be misleading. We need to refactor count() implementation accordingly. For now, we can just remove the validation from here.

ankitjain235: Right, removed the validation for now.

PrasadG193 (approved):

PR #801, opened 2020-10-26
Workaround Minio directory semantics

vkamra:
## Change Overview

This PR works around the difference in semantics/support of "directory objects" in Minio.

Other object stores have a flat namespace backend which supports creating a "directory marker" 
object to represent a "directory".

Minio uses a filesystem backend which makes this hard to support.
They've also stated they will not add support for this - [here](https://github.com/minio/minio/issues/7335) and [here](https://github.com/minio/minio/issues/7397) for example.

The difference in semantics can be illustrated with the following operations
where once an object is created in Minio with the same prefix as one of the
directory markers - that object is no longer visible.

**S3**:
```
PUT (dir1/, <empty>)                            // Create a directory marker
GET (dir1/) -> 200                              // Get the directory marker object -> OK
LIST(objects with prefix "dir1/") -> "dir1/"    // List objects with prefix "dir1/" 
PUT (dir1/file1, <data>)                        // Create an object with the prefix "dir1/"
GET(dir1/) -> 200                               // Get the directory marker object -> OK
LIST(objects with prefix "dir1/") -> "dir1/", "dir1/file1" // List objects with prefix "dir1/" 
```
**Minio**:
```
PUT (dir1/, <empty>)                                // Create a directory marker
GET (dir1/) -> 200                                  // Get the directory marker object -> OK
LIST(objects with prefix "dir1/") -> "dir1/"        // List objects with prefix "dir1/" 
PUT (dir1/file1, <data>)                            // Create an object with the prefix "dir1/"
GET(dir1/) -> 404                                   // Get the directory marker object -> 404
LIST(objects with prefix "dir1/") -> "dir1/file1"   // List objects with prefix "dir1/" 
```

Earlier tests did not surface this issue but Minio recently fixed a [bug](https://github.com/minio/minio/pull/10315) in their standalone version (which was hiding this issue).

Follow-ups based on offline discussions with @julio-lopez 

- The fixes here remove the requirement for "directory marker" objects.
   We should be able to remove that logic in the `objectstore` package.
- Leverage `Delimiter` in `ListObjectsInput` API for S3 stores to improve performance
   of List object calls.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

Ran `objectstore` `TestDirectories` test against a local minio deployment - 

https://gist.github.com/vkamra/67f57d0ec496d0878a36e955f461e1b3


pkg/objectstore/directory.go
```diff
@@ -296,10 +310,11 @@ func isDirectoryObject(path string) (string, bool) {
 		// No '/'
 		return "", false
 	case 2:
+		// dir1/dir2
```
vkamra, line 313:
```suggestion
		// dir1/
```

vkamra, line 313: ListDirs("p/"

p/dir1/dir2/elem

S3 -> Walk ("p/") -> "p/", "p/dir1/", "p/dir1/dir2/", "p/dir1/dir2/elem"

Minio -> Walk ("p/") -> "p/dir1/dir2/elem"

pkg/objectstore/directory.go
```diff
@@ -61,9 +61,22 @@ func (d *directory) GetDirectory(ctx context.Context, dir string) (Directory, er
...
 	dir = d.absDirName(dir)
-	_, err := d.bucket.container.Item(cloudName(dir))
-	if err != nil {
+	if _, err := d.bucket.container.Item(cloudName(dir)); err == nil {
+		return &directory{
+			bucket: d.bucket,
+			path:   dir,
+		}, nil
+	}
+
```
tdmanv, line 70: [minor] you could check for a 404 here and return the error otherwise.

tdmanv (commented):
Question on a behavior change in aws:
Lets say my director contains a file `foobar` and I call `GetDirectory(`foo`). Will this return `&directory{path: "foo"}`?
Can we guarantee that `dir` will always contain a trailing slash? If so, we should probably validate it.

vkamra:
> Question on a behavior change in aws:
> Lets say my director contains a file `foobar` and I call `GetDirectory(`foo`). Will this return `&directory{path: "foo"}`? Can we guarantee that `dir` will always contain a trailing slash? If so, we should probably validate it.

Yep - I did look into this. We do the following in `GetDirectory` which will always add a trailing slash - so it won't return `foobar` in this case:
```
dir = d.absDirName(dir)
```

tdmanv (approved):

pkg/objectstore/directory.go
vkamra: Makes sense - will follow up. I think I might need to account for error differences b/w GCS, S3, etc - so would need some refactoring to do that.

julio-lopez (approved):

PR #802, opened 2020-10-28
Add mariadb readme in examples

viveksinghggits:
## Change Overview

This change adds the maridb example. Restore is failing because of below error 
```
ERROR 1556 (HY000) at line 819: You can't use locks with log tables
```
even after this error, the deleted databases are restored successfully but the restore job is being failed
At some places people are saying these logs can be ignored.

Even if we try to skip the logs table, it looks like those tables are not being skipped 

```
mysqldump --column-statistics=0 -u root --password=${root_password} -h {{ .StatefulSet.Name }} --single-transaction --ignore-table=mysql.slow_log --ignore-table=mysql.general_log  --all-databases
```
I am still looking into how we can get rid of that error.

[this is](https://jira.mariadb.org/browse/MDEV-8494) a related discussion that was raised earlier. 

I [raised an issue](https://jira.mariadb.org/browse/MDEV-24050) with maria DB team as well, let's see what they respond with.

#### Update

We have updated the backup process to just include the non-system databases by excluding `performance_schema`, `infromation_schema` and `mysql` database while taking the dump.
And things work as expected now.

## Pull request type

- [x] :hamster: Trivial/Minor

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


viveksinghggits:
In the recent commits we excluded `performance_schema`, and the reason is if we don't exclude the database, this is the error that we get 
```
ERROR 1050 (42S01) at line 117: Table 'accounts' already exists
```

examples/stable/maria/README.md
```diff
@@ -0,0 +1,265 @@
+# MariaDB
+
+[MariaDB]https://mariadb.org/) is open source relation database that that is made by the developers of MySQL.
```
ankitjain235, line 3:
```suggestion
[MariaDB](https://mariadb.org/) is open source relation database that that is made by the developers of MySQL.
```

ankitjain235 (commented):
Just a minor README comment, otherwise LGTM.

ankitjain235 (approved):

PR #881, opened 2021-01-16
Silence errors on integration test teardown

tdmanv:
## Change Overview

There have been intermittent failures in the integration tests on teardown. This PR will make those failures log-only.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues

- #XXX

### DIFF
--- a/pkg/testing/integration_test.go
+++ b/pkg/testing/integration_test.go
@@ -419,13 +419,19 @@ func (s *IntegrationSuite) TearDownSuite(c *C) {
 	// Uninstall app
 	if !s.skip {
 		err := s.app.Uninstall(ctx)
-		c.Assert(err, IsNil)
+		if err != nil {
+			c.Log("Failed to uninstall app %s: %v", s.name, err.Error())
+		}
+
 	}
 
 	// Uninstall implementation of the apps doesn't delete namespace
 	// Delete the namespace separately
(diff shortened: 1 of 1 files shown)


julio-lopez:
The PR subject is misleading because it is not silencing the errors, but instead preventing the test/job from failing, or something like that.

pkg/testing/integration_test.go
```diff
@@ -419,13 +419,19 @@ func (s *IntegrationSuite) TearDownSuite(c *C) {
...
+		}
+
 	}
 
 	// Uninstall implementation of the apps doesn't delete namespace
 	// Delete the namespace separately
 	err := s.cli.CoreV1().Namespaces().Delete(ctx, s.namespace, metav1.DeleteOptions{})
 	c.Assert(err, IsNil)
+	if err != nil {
+		c.Log("Failed to delete namespace: %v", err.Error())
```
julio-lopez, line 433: Will this execute after the Assert above?

What is the desired behavior?
* Is it to avoid stopping execution? then in that case `c.Check()` may be more appropriate and there is no need for an explicit `c.Log()` since `c.Check()` will do _"the right thing"_ ™️ 
* Not causing a test (teardown) failure on error? then remove the prior `c.Assert()`

PR #895, opened 2021-02-10
Snapshot class in storageclass override

bathina2:
## Change Overview

This changes allows the storageclass to be annotated with its volumesnapshotclass 

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/snapshot/snapshot_alpha.go
+++ b/pkg/kube/snapshot/snapshot_alpha.go
@@ -417,6 +417,8 @@ func TransformUnstructured(u *unstructured.Unstructured, value interface{}) erro
 	return errors.Wrapf(err, "Failed to Unmarshal unstructured object")
 }
 
+// GetSnapshotClassbyAnnotation checks if the provided annotation is present in either the storageclass
+// or volumesnapshotclass and returns the volumesnapshotclass.
 func GetSnapshotClassbyAnnotation(dynCli dynamic.Interface, kubeCli kubernetes.Interface, gvr schema.GroupVersionResource, annotationKey, annotationValue, storageClass string) (string, error) {
 	// fetch storageClass
 	sc, err := kubeCli.StorageV1().StorageClasses().Get(context.TODO(), storageClass, metav1.GetOptions{})
--- a/pkg/kube/snapshot/snapshot_alpha.go
+++ b/pkg/kube/snapshot/snapshot_alpha.go
@@ -19,7 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
+	v1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	k8errors "k8s.io/apimachinery/pkg/api/errors"
@@ -84,7 +84,7 @@ func (sna *SnapshotAlpha) CloneVolumeSnapshotClass(sourceClassName, targetClassN
 
 // GetVolumeSnapshotClass returns VolumeSnapshotClass name which is annotated with given key.
 func (sna *SnapshotAlpha) GetVolumeSnapshotClass(annotationKey, annotationValue, storageClassName string) (string, error) {
-	return getSnapshotClassbyAnnotation(sna.dynCli, sna.kubeCli, v1alpha1.VolSnapClassGVR, annotationKey, annotationValue, storageClassName)
+	return GetSnapshotClassbyAnnotation(sna.dynCli, sna.kubeCli, v1alpha1.VolSnapClassGVR, annotationKey, annotationValue, storageClassName)
 }
 
 // Create creates a VolumeSnapshot and returns it or any error that happened meanwhile.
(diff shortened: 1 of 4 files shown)


pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -417,19 +417,27 @@ func TransformUnstructured(u *unstructured.Unstructured, value interface{}) erro
 	return errors.Wrapf(err, "Failed to Unmarshal unstructured object")
 }
 
-func getSnapshotClassbyAnnotation(dynCli dynamic.Interface, kubeCli kubernetes.Interface, gvr schema.GroupVersionResource, annotationKey, annotationValue, storageClass string) (string, error) {
+func GetSnapshotClassbyAnnotation(dynCli dynamic.Interface, kubeCli kubernetes.Interface, gvr schema.GroupVersionResource, annotationKey, annotationValue, storageClass string) (string, error) {
```
onkarbhat, line 420: Add comments for the exported function ?

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -417,19 +417,27 @@ func TransformUnstructured(u *unstructured.Unstructured, value interface{}) erro
...
+	// fetch storageClass
+	sc, err := kubeCli.StorageV1().StorageClasses().Get(context.TODO(), storageClass, metav1.GetOptions{})
```
onkarbhat, line 422: @bathina2 - Not applicable for this PR. But should we propagate the context from the caller instead of using context.TODO ?

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -417,19 +417,27 @@ func TransformUnstructured(u *unstructured.Unstructured, value interface{}) erro
...
+	if err != nil {
+		return "", errors.Errorf("Failed to find StorageClass (%s) in the cluster: %v", storageClass, err)
+	}
+	// Check if storageclass annotation override is present.
+	if val, ok := sc.Annotations[annotationKey]; ok {
+		vsc, err := dynCli.Resource(gvr).Get(context.TODO(), val, metav1.GetOptions{})
+		if err != nil {
```
onkarbhat, line 429: Should we check the annotation value and the driver ( we check for this later at Line 447 ) ? 

```
if val, ok := ans[annotationKey]; ok && val == annotationValue && driver == sc.Provisioner {
```

pkg/kube/snapshot/snapshot_test.go
```diff
@@ -1003,3 +1003,83 @@ func (s *SnapshotTestSuite) TestCreateFromSourceStable(c *C) {
...
+			sc:         "sc1",
+			errChecker: NotNil,
+		},
+		{ // annotation on sc
+			dyncli:     dynfake.NewSimpleDynamicClient(scheme, vsc2),
+			kubecli:    fake.NewSimpleClientset(sc2),
+			gvr:        v1beta1.VolSnapClassGVR,
+			key:        "key",
+			value:      "value",
+			sc:         "sc2",
```
onkarbhat, line 1065: Based on my comment above about checking for the value , in this test case, value is being passed as `value` . But value in the annotations map in sc2 is actually `vsc2` .

pkg/kube/snapshot/snapshot_test.go
```diff
@@ -1003,3 +1003,83 @@ func (s *SnapshotTestSuite) TestCreateFromSourceStable(c *C) {
...
+			errChecker: IsNil,
+			retVSC:     "vsc2",
+		},
+		{ // missing vsc
```
onkarbhat, line 1069: What is the difference between this test case and the one with the comment `no vsc available` ? Should we remove one of them ?

onkarbhat (changes requested):

pkg/kube/snapshot/snapshot_alpha.go
bathina2: will add

pkg/kube/snapshot/snapshot_alpha.go
bathina2: I think overall we should be doing that,yes

pkg/kube/snapshot/snapshot_alpha.go
bathina2: No this may not always be the case (even though it mostly is). The below check is more strict and enforces that. This is an override that allows any volumesnapshotclass to be specified

pkg/kube/snapshot/snapshot_test.go
bathina2: In the function signature the value being passed in was previously important. In this case its not. We check to see what if the key exists in a storageclass and see if its a valid volumesnapshot class.

pkg/kube/snapshot/snapshot_test.go
bathina2: They are both similar but testing two different code paths.

onkarbhat: I see, yes. Makes sense.

pkg/kube/snapshot/snapshot_test.go
onkarbhat: Sounds good. Can you please update the values in the test cases to indicate that it is true/false when we use the annotation from the VSC , and to indicate that it is the VSC name when we use the annotation in the storage class ?

pkg/kube/snapshot/snapshot_alpha.go
onkarbhat: Sounds good.

onkarbhat (approved):
LGTM. Minor comment about the values in the test cases.

PR #917, opened 2021-03-03
Remove github actions from CI

viveksinghggits:
## Change Overview

Even though we disabled it not to run on PRs, by removing
`on` field from the main.yml file, it was complaining that
`on` field is not found in the main.yml.
I think we should get it removed forever.

## Pull request type

## Issues

- #XXX

## Test Plan

NA


chaitanya-baraskar (approved):

viveksinghggits:
I think we will have to merge this manually.

PR #927, opened 2021-03-11
Upgrade Kanister CRDs to v1 from v1beta1

viveksinghggits:
## Change Overview

This PR tries to update the version of CRDs that we create to `v1` from `v1beta1`. Some of the issue that I faced 
- When we create CRD using kube-buidler, the fields are marked as `required` but we don't necessarily provide all the fields while creating the CRs. So we had to remove those `required` fields from CRDs.

After above changes, things were working on cluster version 1.20, but in kanister CI we run kubernetes version 1.18, and there things were failing because when we read the blueprint from `examples/` dir to actually marshal it to types json, the values that we didnt specify in the manifest were being set as `null`. To resolve that we had to add `omitempty` in the types.go for some fields.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [x] :green_heart: E2E

Change `integration_test.sh` to run the test for
```
SHORT_APPS="^MongoDB$"
```

With the Kubernetes version that runs in CI
```
» k version
» make integration-test
```
logs can be found [here](https://gist.github.com/e64a1d984a9a1c337e38fa4081588bbd).

With kubernetes version 1.20
```
 » make integration-test 
```

and the logs can be found [here](https://gist.github.com/08f73024af54c1c9805c98ee4cf0cb14).


pkg/apis/cr/v1alpha1/types.go
```diff
@@ -178,12 +172,6 @@ type ActionSetList struct {
 	Items           []*ActionSet `json:"items"`
 }
 
-// These names are used to query Blueprint API objects.
-const (
-	BlueprintResourceName       = "blueprint"
-	BlueprintResourceNamePlural = "blueprints"
-)
-
```
viveksinghggits, line 186: I have moved these to `consts` package to resolve import cycle issue that we faced when we tried to use them from `customresource` package.

viveksinghggits:
Per current implementation if CRDs are already created in the cluster nothing would happen. But in our case we would like to re-create them with `v1 apiVersion` to achieve that, if we delete and create the CRD we lose the CRs that were already created.
And we are getting below error while updating the CRD if its already present 
```
customresourcedefinitions.apiextensions.k8s.io \"profiles.cr.kanister.io\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```
I am continuing looking into this.

pkg/consts/consts.go
```diff
@@ -9,3 +9,13 @@ const (
 	LabelKeyCreatedBy        = "createdBy"
 	LabelValueKanister       = "kanister"
 )
+
+// These names are used to query ActionSet API objects.
```
pavannd1, line 13:
```suggestion
// These names are used to query API objects.
```

pkg/customresource/actionset-crd.go
```diff
@@ -0,0 +1,254 @@
...
+                            description: 'Namespace of the referent. More info: http://kubernetes.io/docs/user-guide/namespaces'
+                            type: string
+                          resource:
+                            description: Resource name of the referent.
+                            type: string
+                        type: object
+                      description: ConfigMaps that we'll get and pass into the blueprint.
+                      type: object
+                    name:
+                      description: 'Name is the action we''ll perform. For example:
```
pavannd1, line 83:
```suggestion
                      description: 'Name is the action we perform. For example:
```

pkg/customresource/actionset-crd.go
```diff
@@ -0,0 +1,254 @@
...
+                            description: 'Name of the referent. More info: http://kubernetes.io/docs/user-guide/identifiers#names'
+                            type: string
+                          namespace:
+                            description: 'Namespace of the referent. More info: http://kubernetes.io/docs/user-guide/namespaces'
+                            type: string
+                          resource:
+                            description: Resource name of the referent.
+                            type: string
+                        type: object
+                      description: ConfigMaps that we'll get and pass into the blueprint.
```
pavannd1, line 80:
```suggestion
                      description: ConfigMaps that we get and pass into the blueprint.
```

pkg/customresource/actionset-crd.go
```diff
@@ -0,0 +1,254 @@
...
+                      type: object
+                    name:
+                      description: 'Name is the action we''ll perform. For example:
+                        backup or restore.'
+                      type: string
+                    object:
+                      description: Object refers to the thing we'll perform this action
```
pavannd1, line 87:
```suggestion
                      description: Object refers to the resource we perform this action
```

pkg/customresource/actionset-crd.go
```diff
@@ -0,0 +1,254 @@
...
+                            description: 'Name of the referent. More info: http://kubernetes.io/docs/user-guide/identifiers#names'
+                            type: string
+                          namespace:
+                            description: 'Namespace of the referent. More info: http://kubernetes.io/docs/user-guide/namespaces'
+                            type: string
+                          resource:
+                            description: Resource name of the referent.
+                            type: string
+                        type: object
+                      description: Secrets that we'll get and pass into the blueprint.
```
pavannd1, line 168:
```suggestion
                      description: Secrets that we get and pass into the blueprint.
```

pkg/customresource/actionset-crd.go
```diff
@@ -0,0 +1,254 @@
...
+                            x-kubernetes-preserve-unknown-fields: true
+                        type: object
+                      description: Artifacts created by this phase.
+                      type: object
+                    blueprint:
+                      description: Blueprint with instructions on how to execute this
+                        action.
+                      type: string
+                    name:
+                      description: 'Name is the action we''ll perform. For example:
```
pavannd1, line 198:
```suggestion
                      description: 'Name is the action we perform. For example:
```

pkg/customresource/actionset-crd.go
```diff
@@ -0,0 +1,254 @@
...
+                        backup or restore.'
+                      type: string
+                    object:
+                      description: Object refers to the thing we'll perform this action
```
pavannd1, line 202:
```suggestion
                      description: Object refers to the resource we perform this action
```

pkg/customresource/customresource.go
```diff
@@ -93,48 +97,76 @@ func CreateCustomResources(context Context, resources []CustomResource) error {
...
+		return nil, err
 	}
+	return crd, nil
+}
 
-	_, err := context.APIExtensionClientset.ApiextensionsV1beta1().CustomResourceDefinitions().Create(contextpkg.TODO(), crd, metav1.CreateOptions{})
+func decodeSpecIntoObject(spec []byte, intoObj runtime.Object) error {
+	d := serializer.NewCodecFactory(runtime.NewScheme()).UniversalDeserializer()
+	if _, _, err := d.Decode(spec, nil, intoObj); err != nil {
+		return fmt.Errorf("Failed to decode spec into object %s and spec %s\n", err.Error(), spec)
```
pavannd1, line 111:
```suggestion
		return fmt.Errorf("Failed to decode spec into object: %s; spec %s\n", err.Error(), spec)
```

pkg/customresource/customresource.go
```diff
@@ -93,48 +97,76 @@ func CreateCustomResources(context Context, resources []CustomResource) error {
...
+	_, err = context.APIExtensionClientset.ApiextensionsV1().CustomResourceDefinitions().Create(ctx, crd, metav1.CreateOptions{})
 	if err != nil {
 		if !errors.IsAlreadyExists(err) {
 			return fmt.Errorf("failed to create %s CRD. %+v", resource.Name, err)
 		}
+
+		// if CRD already exists, get the resource version and create the CRD with that resource version
+		c, err := context.APIExtensionClientset.ApiextensionsV1().CustomResourceDefinitions().Get(ctx, crd.Name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("Failed to get CRD to get resource version %s\n", err.Error())
```
pavannd1, line 131:
```suggestion
			return fmt.Errorf("Failed to get CRD to get resource version: %s\n", err.Error())
```

pkg/customresource/customresource.go
```diff
@@ -93,48 +97,76 @@ func CreateCustomResources(context Context, resources []CustomResource) error {
...
+		}
+
+		crd.ResourceVersion = c.ResourceVersion
+		_, err = context.APIExtensionClientset.ApiextensionsV1().CustomResourceDefinitions().Update(ctx, crd, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("Failed to delete already present CRD %s\n", err.Error())
```
pavannd1, line 137:
```suggestion
			return fmt.Errorf("Failed to delete already present CRD: %s\n", err.Error())
```

pkg/customresource/actionset-crd.go
viveksinghggits: these things are actually being generated by I will change them anyway.

pavannd1 (commented):
Do we need to change the permissions of the `controller/Dockerfile` or is it just a remnant of testing?

viveksinghggits:
> Do we need to change the permissions of the `controller/Dockerfile` or is it just a remnant of testing?

Hi @pavannd1 
Sorry about that we don't need that, I have fixed that now.

pavannd1 (approved):
LGTM

PR #956, opened 2021-04-23
Removing checks for pods that are still running

bathina2:
## Change Overview

WaitForPodCompletion initially returned an error when a container inside the pod failed. This behavior should be controlled by the pods RestartPolicy simplifying the logic. This function will now only return when a pod is in the terminal state of success or failure. Otherwise it exits when the context times out or keep restarting according to the policy.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod.go
+++ b/pkg/kube/pod.go
@@ -279,7 +279,7 @@ func checkPVCAndPVStatus(vol v1.Volume, p *v1.Pod, cli kubernetes.Interface, nam
 	return nil
 }
 
-// WaitForPodCompletion waits for a pod to reach a terminal state
+// WaitForPodCompletion waits for a pod to reach a terminal state, or timeout
 func WaitForPodCompletion(ctx context.Context, cli kubernetes.Interface, namespace, name string) error {
 	err := poll.Wait(ctx, func(ctx context.Context) (bool, error) {
 		p, err := cli.CoreV1().Pods(namespace).Get(ctx, name, metav1.GetOptions{})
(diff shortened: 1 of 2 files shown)


pkg/kube/pod.go
```diff
@@ -282,14 +282,8 @@ func WaitForPodCompletion(ctx context.Context, cli kubernetes.Interface, namespa
...
 		switch p.Status.Phase {
-		case v1.PodRunning:
-			for _, con := range p.Status.ContainerStatuses {
-				if con.State.Terminated != nil {
-					return false, errors.Errorf("Container %v is terminated, while Pod %v is Running. Container termination status: %v. Pod details (%s)", con.Name, name, con.State.Terminated, p.String())
-				}
-			}
 		case v1.PodFailed:
-			return false, errors.Errorf("Pod %s failed", name)
+			return false, errors.Errorf("Pod %s failed. Pod details (%s)", name, p.String())
```
onkarbhat, line 286: @bathina2  - Does this mean the callers of this function have to ensure that they create their pod using the correct RestartPolicy ? How do we make sure that all consumers switch to the correct policy there?

bathina2: @onkarbhat good catch. I've added the ability to pass down the the restart policy as part of the PodOptions. I've also set the default policy to be Never which was what the original code was intending. Going forward if the consumers want to change the policy they can pass it in via the PodOptions.

tdmanv (approved):

pkg/kube/pod.go
onkarbhat: Discussed this on a call with @tdmanv and @bathina2 . LGTM 👍🏼

onkarbhat (approved):

PR #961, opened 2021-04-28
Update kafka integration test to use kafka-bridge

A-kanksh-a:
## Change Overview

This PR is to modify the integration test for Kafka to use rest API exposed through kafkaBridge for producing and consuming the message instead of using `kafka-go` library
link to the log : [https://gist.github.com/A-kanksh-a/c635b96f90dc67a479c8ce2d0026a6c7](https://gist.github.com/A-kanksh-a/c635b96f90dc67a479c8ce2d0026a6c7) 
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

> ./build/integration-test.sh Kafka


pkg/app/kafka.go
```diff
@@ -260,36 +282,79 @@ func (kc *KafkaCluster) Initialize(ctx context.Context) error {
 	return nil
 }
 
-// to produce messages to kafka
-func produce(ctx context.Context, topic string, namespace string) error {
-	partition := 0
-	forwarder, err := K8SServicePortForward(ctx, bootstrapServerHost, namespace, bootstrapServerPort)
+type Message struct {
```
PrasadG193, line 285: Please add comment on all the structs

pkg/app/kafka.go
```diff
@@ -260,36 +282,79 @@ func (kc *KafkaCluster) Initialize(ctx context.Context) error {
...
+	FetchMinBytes            int    `json:"fetch.min.bytes"`
+	ConsumerRequestTimeoutMs int    `json:"consumer.request.timeout.ms"`
+}
+
+type SubscriptionPayload struct {
+	Topics []string `json:"topics"`
+}
+
+// InsertRecord creates a topic and insert a record
+func InsertRecord(ctx context.Context, namespace string) error {
```
PrasadG193, line 316: Let's implement this method on `KafkaCluster`
```suggestion
func (kc *KafkaCluster) InsertRecord(ctx context.Context, namespace string) error {
```

pkg/app/kafka.go
```diff
@@ -260,36 +282,79 @@ func (kc *KafkaCluster) Initialize(ctx context.Context) error {
...
-	_, err = conn.WriteMessages(
-		kafka.Message{Value: []byte("{'title':'The Matrix','year':1999,'cast':['Keanu Reeves','Laurence Fishburne','Carrie-Anne Moss','Hugo Weaving','Joe Pantoliano'],'genres':['Science Fiction']}")},
-	)
-	return err
+	defer resp.Body.Close()
+	bytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	log.Print(string(bytes))
```
PrasadG193, line 356: Please log at debug level or remove it

pkg/app/kafka.go
```diff
@@ -260,36 +282,79 @@ func (kc *KafkaCluster) Initialize(ctx context.Context) error {
...
 
-	defer conn.Close()
+	req, err := http.NewRequest("POST", uri+"/topics/"+topic, body)
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/vnd.kafka.json.v2+json")
 
-	err = conn.SetWriteDeadline(time.Now().Add(10 * time.Second))
+	resp, err := http.DefaultClient.Do(req)
```
PrasadG193, line 347: We should also check for the correct response code. It will not raise an error when the response code is not 200/201

pkg/app/kafka.go
```diff
@@ -370,49 +435,133 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
...
+	body := bytes.NewReader(payloadBytes)
+
+	req, err := http.NewRequest("POST", uri+"/consumers/"+topic+"-consumer-group/instances/"+topic+"-consumer/subscription", body)
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/vnd.kafka.v2+json")
 
-	conn, err := kafka.DialLeader(ctx, "tcp", uri, topic, partition)
+	resp, err := http.DefaultClient.Do(req)
```
PrasadG193, line 489: Let's check for correct response status code

pkg/app/kafka.go
```diff
@@ -370,49 +435,133 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	bytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	log.Print(string(bytes))
```
PrasadG193, line 498: Log at debug level or remove if not required

pkg/app/kafka.go
```diff
@@ -370,49 +435,133 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
...
+	data := SubscriptionPayload{
+		Topics: []string{topic},
+	}
+	payloadBytes, err := json.Marshal(data)
+	if err != nil {
+		return err
+	}
+	body := bytes.NewReader(payloadBytes)
+
+	req, err := http.NewRequest("POST", uri+"/consumers/"+topic+"-consumer-group/instances/"+topic+"-consumer/subscription", body)
```
PrasadG193, line 483: Let's declare the URL format as const and construct string from the format here. This applies to other places as well
```
const subscriptionURLPathFormat = "/consumers/%s-consumer-group/instances/%s-consumer/subscription"
```

pkg/app/kafka.go
```diff
@@ -370,49 +435,133 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
...
+// consumeMessage consumes the message from a topic
+func consumeMessage(uri string) (int, error) {
+	req, err := http.NewRequest("GET", uri+"/consumers/"+topic+"-consumer-group/instances/"+topic+"-consumer/records", nil)
 	if err != nil {
 		return 0, err
 	}
+	req.Header.Set("Accept", "application/vnd.kafka.json.v2+json")
 
-	defer conn.Close()
+	resp, err := http.DefaultClient.Do(req)
```
PrasadG193, line 510: check for response status code

pkg/app/kafka.go
```diff
@@ -370,49 +435,133 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
...
+	bytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return 0, err
 	}
-	if !messageEnd {
-		if err := batch.Close(); err != nil {
-			return 0, err
+	responseBody := string(bytes)
+	var Message []Message
+	_ = json.Unmarshal([]byte(responseBody), &Message)
```
PrasadG193, line 522: Please capture and handle error

pkg/app/kafka.go
```diff
@@ -370,49 +435,133 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
...
+				return true, nil
+			}
 		}
+		return false, nil
+	})
+	if err != nil {
+		return 0, err
 	}
-	return count, nil
+	return recordCount, nil
```
PrasadG193, line 566: we can simply `return recordCount, err` here

pkg/app/kafka.go
```diff
@@ -370,49 +435,133 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
 	return f, nil
 }
 
-func consume(ctx context.Context, topic string, namespace string) (int, error) {
-	partition := 0
-	forwarder, err := K8SServicePortForward(ctx, bootstrapServerHost, namespace, bootstrapServerPort)
+// createConsumerGroup creates a consumer group in kafka cluster
+func createConsumerGroup(uri string) error {
+	data := Payload{
+		Name:                     topic + "-consumer",
```
PrasadG193, line 441: Why are we adding `-consumer` here? We are doing same while constructing URL as well. Can we just use topic and avoid combining string `topic+"-consumer"` while building URLs also?

A-kanksh-a: yeah we can do that, but we have appended `-consumer` as it would differentiate the `topic` and `consumer-group` instances that would subscribe to the topic. So its just a format we are using here.

pkg/app/kafka.go
```diff
@@ -370,49 +447,145 @@ func K8SServicePortForward(ctx context.Context, svcName string, ns string, pPort
...
+	defer resp.Body.Close()
+	bytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	log.Debug().Print(string(bytes))
+	return nil
+}
+
+// Subscribe subscribes a consumer to topic provided in kafka cluster
```
pavannd1, line 489:
```suggestion
// subscribe subscribes a consumer to topic provided in kafka cluster
```

pavannd1 (commented):
Based on whatever I could understand, changes look good. Will let @PrasadG193 take a look and +2 this.

PrasadG193 (approved):
Looks good

PR #963, opened 2021-04-28
Add options field in kopia location profile

PrasadG193:
## Change Overview

Add options field in kopia location profile so that additional options can be passed to kopia client

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/apis/cr/v1alpha1/types.go
+++ b/pkg/apis/cr/v1alpha1/types.go
@@ -290,7 +290,7 @@ type KopiaServerSecret struct {
 	Hostname       string                `json:"hostname,omitempty"`
 	UserPassphrase *KopiaServerSecretRef `json:"userPassphrase,omitempty"`
 	TLSCert        *KopiaServerSecretRef `json:"tlsCert,omitempty"`
-	Options        map[string]int        `json:"options,omitempty"`
+	ConnectOptions map[string]int        `json:"connectOptions,omitempty"`
 }
 
 // KopiaServerSecretRef refers to K8s secrets containing Kopia creds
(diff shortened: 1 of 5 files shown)


pkg/apis/cr/v1alpha1/types.go
```diff
@@ -290,6 +290,7 @@ type KopiaServerSecret struct {
 	Hostname       string                `json:"hostname,omitempty"`
 	UserPassphrase *KopiaServerSecretRef `json:"userPassphrase,omitempty"`
 	TLSCert        *KopiaServerSecretRef `json:"tlsCert,omitempty"`
+	Options        map[string]int        `json:"options,omitempty"`
```
pavannd1, line 293: Should we say `CacheOptions` or `ConnectOptions`?

PrasadG193: @pavannd1 I intentionally kept if generic so that in future if we have to pass other options for different actions, we can use the same map

pkg/param/param.go
```diff
@@ -499,6 +500,7 @@ func fetchKopiaCredential(ctx context.Context, cli kubernetes.Interface, ks *crv
 			Username: ks.Username,
 			Password: string(password),
 			Cert:     string(tlsCert),
+			Options:  ks.Options,
```
pavannd1, line 503: Same as above

pavannd1 (approved):
LGTM. Good to merge after name change!

PR #995, opened 2021-05-12
Change helm version to 2.16.10 from 3

viveksinghggits:
## Change Overview

This PR changes helm version from build docs image to 2.16.10 from 3.

## Pull request type

- [x] :hamster: Trivial/Minor
- [x] :bug: Bugfix

## Issues

- #XXX

## Test Plan

NA

### DIFF
--- a/docker/docs-build/Dockerfile
+++ b/docker/docs-build/Dockerfile
@@ -13,4 +13,4 @@ RUN curl "https://s3.amazonaws.com/aws-cli/awscli-bundle-${AWS_VERSION}.zip" -o 
     ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
 
 # add helm
-COPY --from=alpine/helm:3.1.0 /usr/bin/helm /usr/local/bin/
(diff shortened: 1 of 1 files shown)


ankitjain235 (approved):

docker/docs-build/Dockerfile
```diff
@@ -13,4 +13,4 @@ RUN curl "https://s3.amazonaws.com/aws-cli/awscli-bundle-${AWS_VERSION}.zip" -o
     ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
 
 # add helm
-COPY --from=alpine/helm:3.1.0 /usr/bin/helm /usr/local/bin/
+COPY --from=alpine/helm:2.16.10 /usr/bin/helm /usr/local/bin/
```
pavannd1, line 16: What is the reason for this change?

viveksinghggits: we were running the respective command from shippable node Pavan, where we have helm v2. but now we are running that command from this image. That is the reason we want to change that to v2.

PR #1002, opened 2021-05-19
Update location commands to produce kopia snapshot output

pavannd1:
## Change Overview

- Kopia snapshot struct with encode/decode helpers
- location push command update to produce the snapshot info as output
- location pull and delete commands update to accept kopia snapshot info json as input

Should be merged after #1001 

## Pull request type

- [x] :sunflower: Feature

## Test Plan

- [x] :muscle: Manual
Tested with updated mysql blueprint - Will be added in a follow-up PR


PrasadG193:
We should also update v2 blueprints as per these changes

pkg/kopia/stream.go
```diff
@@ -36,18 +36,28 @@ const (
 	pullRepoPurpose = "kando location pull"
 )
 
+// SnapshotInfo tracks kopia snapshot information produced by a kando command in a phase
+type SnapshotInfo struct {
+	// BackupID is the snapshot ID produced by kopia snapshot operation
+	BackupID string `json:"backupID"`
+	// LogicalSize is the sum of cached and hashed file size in bytes
+	LogicalSize string `json:"logicalSize"`
```
tdmanv, line 44: [nit] I prefer using an int in our internal structures and letting serialization handle parsing

tdmanv (approved):

pkg/kopia/stream.go
```diff
@@ -36,18 +36,28 @@ const (
...
+	// LogicalSize is the uploaded size in bytes
+	PhysicalSize string `json:"physicalSize"`
+}
+
 // Write creates a kopia snapshot from the given reader
 // A virtual directory tree rooted at filepath.Dir(path) is created with
 // a kopia streaming file with filepath.Base(path) as name
-func Write(ctx context.Context, path string, source io.Reader) (string, string, error) {
+func Write(ctx context.Context, path string, source io.Reader) (string, error) {
```
tdmanv, line 52: I'd suggest keeping the amount of time that this struct is serialized as short as possible - can the caller perform the serialization when needed as output?

PR #1003, opened 2021-05-19
Support path without parent directory

pavannd1:
## Change Overview

- If the `path` flag from `kando` command does not have a parent directory, use a default parent directory as the root of the kopia snapshot.
Example:
If `--path 'mysql.dump.gz'` is provided, then the default parent directory will be used.
If `--path '/mysql-backups/mysql.dump.gz` is provided, then `/mysql-backups` will be used.

## Pull request type

- [x] :hamster: Trivial/Minor
- [x] :sunflower: Feature

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kopia/stream.go
```diff
@@ -60,11 +64,19 @@ func Write(ctx context.Context, path string, source io.Reader) (string, error) {
 		return "", errors.Wrap(err, "Failed to open kopia repository")
 	}
 
+	// If the path provided does not have a parent directory OR
+	// has just root (`/`) directory as the parent,
+	// use the default directory as root of the snapshot
+	rootPath := filepath.Dir(path)
+	if rootPath == dotDirString || rootPath == slashDirString {
```
tdmanv, line 71:
```suggestion
	parentPath := filepath.Dir(path)
	if parentPath == dotDirString || parentPath == slashDirString {
```

pkg/kopia/stream.go
```diff
@@ -60,11 +64,19 @@ func Write(ctx context.Context, path string, source io.Reader) (string, error) {
...
+		rootPath = defaultRootDir
+	}
+
 	// Populate the source info with source path
 	sourceInfo := snapshot.SourceInfo{
 		UserName: rep.ClientOptions().Username,
 		Host:     rep.ClientOptions().Hostname,
-		Path:     filepath.Dir(path),
+		Path:     rootPath,
```
tdmanv, line 79:
```suggestion
		Path:     parentPath,
```

tdmanv (commented):
This defaulting behavior seems a bit strange to me - could you please elaborate a bit on the use case?

pavannd1:
@tdmanv 
Kopia stdin builds a virtual directory tree with a static directory as the root. This static directory has a single streaming file entry which is the name used for the file created with the stdin data.
In our case, the path provided to kando could include a directory or just be a file name.
Examples:
- If `kando --path 'mysql.dump.gz'` is used, then the static directory ends up being `.` if we use `filepath.Dir(path)` as the root path for the kopia snapshot. 
With this defaulting behavior, we would have `/kanister-backups` as the root path of the snapshot with `mysql.dump.gz` as the streaming file entry.

- If `kando --path '/mysql-backups/mysql.dump.gz` is used, then `filepath.Dir(path)` will produce `/mysql-backups` which would be the root path of the snapshot

tdmanv (approved):

pavannd1:
Apologies. I may have sniped the `go fmt` commit!

PR #1009, opened 2021-05-26
Validate snapshot ID in SnapshotInfo

PrasadG193:
## Change Overview

Verify Kopia snapshot ID is not empty while marshal, unmarshal SnapshotInfo

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kopia/utils.go
+++ b/pkg/kopia/utils.go
@@ -194,8 +194,8 @@ func GetDataStoreGeneralMetadataCacheSize(opt map[string]int) int {
 
 // MarshalKopiaSnapshot encodes kopia SnapshotInfo struct into a string
 func MarshalKopiaSnapshot(snapInfo *SnapshotInfo) (string, error) {
-	if snapInfo.ID == "" {
-		return "", errors.New("kopia snapshot ID cannot be empty")
+	if err := snapInfo.Validate(); err != nil {
+		return "", err
 	}
 	snap, err := json.Marshal(snapInfo)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pavannd1 (approved):
LGTM

pkg/kopia/utils.go
```diff
@@ -194,6 +194,9 @@ func GetDataStoreGeneralMetadataCacheSize(opt map[string]int) int {
 
 // MarshalKopiaSnapshot encodes kopia SnapshotInfo struct into a string
 func MarshalKopiaSnapshot(snapInfo *SnapshotInfo) (string, error) {
+	if snapInfo.ID == "" {
```
tdmanv, line 197: Do you need a nil check here?

Perhaps a better pattern would be to add a receiving method:

```
func (si*SnapshotInfo) Validate() error{
    if si == nil {
        return errors.New("Invalid Kopia snapshot: cannot be nil")
    }
    if si.ID == "" {
        return errors.New("Invalid Kopia snaoshot: ID cannot be empty")
    }
    return nil
}
```

tdmanv (approved):

pkg/kopia/utils.go
PrasadG193: @tdmanv makes sense. I have made the changes

PR #1011, opened 2021-05-27
Remove `type` field from stable blueprints

viveksinghggits:
## Change Overview

Since we are upgrading the CRDs to `v1` from `v1beta1`, and we
don't actually specify `type` field in our `types.go` we wont
have `types` field in the CR schema. And because of that if
we create a blueprint with `types` in it, the validation would
failed.

I see there would be a problem when these applications are being
called from k10, because it looks like k10 needs type for `delete`
action.

I still am looking into if we really need that or we can change that.

## Pull request type

## Issues

- #XXX

## Test Plan

Changed blueprint to remove `type` field from it and ran mysql integration test.

```
» make integration-test
```
the logs can be found [here](https://gist.github.com/6043fefec8f22f5565a004d64320000b)

I have tested the change with k10 as well, I created a blueprint without `type` field and then did backup and restore of application. And have tested e2e as well. The logs of e2e can be found [here]().https://gist.github.com/fdd0fe3032f2736e8e88376150389c48

### DIFF
--- a/examples/stable/cassandra/cassandra-blueprint.yaml
+++ b/examples/stable/cassandra/cassandra-blueprint.yaml
@@ -4,7 +4,6 @@ metadata:
   name: cassandra-blueprint
 actions:
   backup:
-    type: StatefulSet
     outputArtifacts:
       params:
         keyValue:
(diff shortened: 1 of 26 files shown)


PrasadG193:
@viveksinghggits we would still need `Kind` field, right?

viveksinghggits:
> @viveksinghggits we would still need `Kind` field, right?

yes, we would need that. Did I remove that somewhere.

chaitanya-baraskar (approved):

PR #1028, opened 2021-06-18
Add dir support to kando with kopia location

PrasadG193:
Signed-off-by: Prasad Ghangal <prasad.ghangal@gmail.com>

## Change Overview

This approach has the following limitations:
1. The backup performed with stdin as a source, can be restored ONLY as stdout target
2. The backup performed with file/dir path as a source, can be restored ONLY as file/dir path as target

Please see the test plan for the details.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

Test cases:
1. Backup file with stdin as source, restore file with stdout as target - PASS
2. Backup file with stdin as source, restore file with filepath as target - FAIL
3. Backup file with filepath as source, restore file with filepath as target - PASS
4. Backup file with filepath as source, restore file with stdout as target - FAIL
5. Backup dir with filepath as source, restore file with stdout as target and `--path` arg with same value as source filepath  - PASS
6. Backup dir with dirpath as source, restore dir with dirpath as target - PASS


pavannd1:
I think test case 2 should work if the target path is same as the backup file path. Please try it out.

pkg/kopia/stream.go
```diff
@@ -114,6 +117,78 @@ func Write(ctx context.Context, path string, source io.Reader) (*SnapshotInfo, e
 	return snapshotInfo, nil
 }
 
+func WriteFile(ctx context.Context, path string, sourcePath string) (*SnapshotInfo, error) {
```
pavannd1, line 120: Please add a comment

pkg/kopia/stream.go
```diff
@@ -146,6 +221,42 @@ func Read(ctx context.Context, backupID, path string, target io.Writer) error {
 	return errors.Wrap(err, "Failed to copy snapshot data to the target")
 }
 
+func ReadFile(ctx context.Context, backupID, target string) error {
```
pavannd1, line 224: Same here. Requires a comment

PrasadG193:
> I think test case 2 and 4 should work if the target path is same as the backup file path. Please try it out.

@pavannd1 it works on master where we create Reader/Writer from file and pass it to stream handling. But since we have added separate handling for files and streaming data, they are no longer compatible.

pkg/kopia/stream.go
```diff
@@ -146,6 +221,42 @@ func Read(ctx context.Context, backupID, path string, target io.Writer) error {
...
+	if err != nil {
+		return errors.Wrap(err, "Unable to resolve path")
+	}
+	// TODO: Do we want to keep this flags configurable?
+	output := &restore.FilesystemOutput{
+		TargetPath:             p,
+		OverwriteDirectories:   true,
+		OverwriteFiles:         true,
+		OverwriteSymlinks:      true,
+		IgnorePermissionErrors: true,
```
PrasadG193, line 251: @pavannd1 do we want to have these flags configurable?

pavannd1: Should be okay for now. We can improve later if the requirement comes up.

pavannd1 (approved):
LGTM

PR #1032, opened 2021-06-21
Improve error message for undefined template key

A-kanksh-a:
## Change Overview

Improve error message while rendering undefined template key. Currently, we get an error message `" map has no entry for key "`

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- K10-7240

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

Error message when `'{{ .ArtifactsIn.cloudObject.KeyValue.paths }}' ` is being rendered and `paths` is undefined.

`time="2021-06-28T06:42:27.261951017Z" level=info msg="Failed to execute phase: v1alpha1.Phase{Name:\"pullFromBlobStore\", State:\"pending\", Output:map[string]interface {}(nil)}:" ActionSet=restore-backup-ggzfj-xbmgf File=pkg/controller/controller.go Function="github.com/kanisterio/kanister/pkg/controller.(*Controller).runAction.func1" Line=438 Phase=pullFromBlobStore cluster_name=6fe9c799-279c-4f46-bfda-c6e470080eaa error="Can't render template: key <paths> Not found" hostname=myrelease-kanister-operator-68ff8c8d54-tgp7v`

### DIFF
--- a/pkg/param/render.go
+++ b/pkg/param/render.go
@@ -140,7 +140,7 @@ func newUndefinedKeyError(err string) error {
 	adjustedPos := pos + len(undefinedKeyErrorMsg)
 	key := err[adjustedPos:]
 	key = strings.Trim(key, "\"")
-	return errors.WithStack(errors.New("Error rendering template key: " + key + " is undefined"))
+	return errors.WithStack(errors.New("Can't render template: key <" + key + "> Not found"))
 }
 
 // RenderObjectRefs function renders object refs from TemplateParams
--- a/pkg/param/render.go
+++ b/pkg/param/render.go
@@ -120,18 +120,24 @@ func renderStringArg(arg string, tp TemplateParams) (string, error) {
 	}
 	buf := bytes.NewBuffer(nil)
 	if err = t.Execute(buf, tp); err != nil {
+		// Check if Error is because of undefined key,
+		// which will lead on execute error on first undefined (missingkey=error).
+		// Get the undefined key name from the error message.
 		if strings.Contains(err.Error(), "map has no entry for key ") {
-			// Check if Error is because of undefined key,
-			// which will lead on execute error on first undefined (missingkey=error).
-			// Get the undefined key name from the error message.
-			pos := strings.LastIndex(err.Error(), "map has no entry for key ")
-			adjustedPos := pos + len("map has no entry for key ")
-			key := err.Error()[adjustedPos:len(err.Error())]
-			return "", errors.WithStack(errors.New(key + " key is undefined"))
+			errMsg := NewErrorMessage(err.Error(), "map has no entry for key ")
+			return "", errors.WithStack(errors.New(errMsg))
 		}
 		return "", errors.WithStack(err)
 	}
 	return buf.String(), nil
+}
+
+func NewErrorMessage(err string, value string) string {
+	pos := strings.LastIndex(err, value)
+	adjustedPos := pos + len(value)
+	key := err[adjustedPos:len(err)]
+	key = strings.Trim(key, "\"")
+	return "Error rendering template key: " + key + " is undefined"
 }
 
 // RenderObjectRefs function renders object refs from TemplateParams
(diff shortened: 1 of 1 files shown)


pkg/param/render.go
```diff
@@ -119,11 +120,26 @@ func renderStringArg(arg string, tp TemplateParams) (string, error) {
...
 	}
 	return buf.String(), nil
 }
 
+func NewErrorMessage(err string, value string) string {
+	pos := strings.LastIndex(err, value)
+	adjustedPos := pos + len(value)
+	key := err[adjustedPos:]
+	key = strings.Trim(key, "\"")
+	return "Error rendering template key: " + key + " is undefined"
```
viveksinghggits, line 140: This still doesn't look very clear to me, let's try to further improve this.

pkg/param/render.go
```diff
@@ -119,11 +120,26 @@ func renderStringArg(arg string, tp TemplateParams) (string, error) {
 	}
 	buf := bytes.NewBuffer(nil)
 	if err = t.Execute(buf, tp); err != nil {
+		// Check if Error is because of undefined key,
+		// which will lead on execute error on first undefined (missingkey=error).
+		// Get the undefined key name from the error message.
+		if strings.Contains(err.Error(), "map has no entry for key ") {
```
pavannd1, line 126: Add `"map has no entry for key "` into a string constant and use here

pkg/param/render.go
```diff
@@ -119,11 +120,26 @@ func renderStringArg(arg string, tp TemplateParams) (string, error) {
...
+			errMsg := NewErrorMessage(err.Error(), "map has no entry for key ")
+			return "", errors.WithStack(errors.New(errMsg))
+		}
 		return "", errors.WithStack(err)
 	}
 	return buf.String(), nil
 }
 
+func NewErrorMessage(err string, value string) string {
```
pavannd1, line 135: The name is not appropriate. This is not a generic function but can be used only with the error above. I would recommend using something more specific. `newUndefinedKeyError(err)`. No need to export the function and no need to pass in the `map has no entry for key` since we are changing it to a string const above.
Also, it can directly return `errors.WithStack` instead of a message.

pkg/param/render.go
```diff
@@ -119,11 +124,25 @@ func renderStringArg(arg string, tp TemplateParams) (string, error) {
...
 		return "", errors.WithStack(err)
 	}
 	return buf.String(), nil
 }
 
+func newUndefinedKeyError(err string) error {
+	pos := strings.LastIndex(err, undefinedKeyErrorMsg)
+	adjustedPos := pos + len(undefinedKeyErrorMsg)
+	key := err[adjustedPos:]
+	key = strings.Trim(key, "\"")
```
viveksinghggits, line 142:
```suggestion
	key = strings.Trim(err[adjustedPos:], "\"")
```

pkg/param/render.go
```diff
@@ -119,11 +124,25 @@ func renderStringArg(arg string, tp TemplateParams) (string, error) {
...
+	return errors.WithStack(errors.New("Can't render template: key <" + key + "> Not found"))
```
viveksinghggits, line 143: let's use `fmt.Sprintf` here.
I was thinking if we can say something like this `key` is not already set, or was not set by previous actions.

viveksinghggits:
```suggestion
	return errors.WithStack(errors.New("Can't render template: "key" is not set"))
```

pkg/param/render.go
```diff
@@ -119,11 +125,24 @@ func renderStringArg(arg string, tp TemplateParams) (string, error) {
...
 		return "", errors.WithStack(err)
 	}
 	return buf.String(), nil
 }
 
+func newUndefinedKeyError(err string) error {
+	pos := strings.LastIndex(err, undefinedKeyErrorMsg)
+	adjustedPos := pos + len(undefinedKeyErrorMsg)
+	key := strings.Trim(err[adjustedPos:], "\"")
+	return errors.WithStack(errors.New(fmt.Sprintf("Can't render template: \"%s\" Not found", key)))
```
pavannd1, line 143: [nit]
```suggestion
	return errors.WithStack(errors.New(fmt.Sprintf("Failed to render template: \"%s\" not found", key)))
```

pavannd1 (approved):

PR #1048, opened 2021-07-12
Add actionsetTimeout helm flag

akankshakumari393:
## Change Overview

Add actionset Timeout helm flag, which would set the context timeout for `handleActionset() `

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/controller/controller.go
```diff
@@ -409,6 +425,24 @@ func (c *Controller) runAction(ctx context.Context, as *crv1alpha1.ActionSet, aI
 			}
 			var rf func(*crv1alpha1.ActionSet) error
 			if err != nil {
+				// If error is because of context canceled && timeout then update the actionset with failed status 
+				if ctx.Err() != nil {
+					bpName := as.Spec.Actions[aIDX].Blueprint
+					bp, _ := c.crClient.CrV1alpha1().Blueprints(as.GetNamespace()).Get(context.Background(), bpName, v1.GetOptions{})
```
PrasadG193, line 431: Why are we using `context.Background()` here? Should we use `ctx` instead?
Same for other occurrences in the same function

A-kanksh-a: @PrasadG193 I believe If we use `ctx` that will give an error since `ctx` is already canceled at this point

pkg/controller/controller.go
```diff
@@ -350,7 +353,9 @@ func (c *Controller) handleActionSet(as *crv1alpha1.ActionSet) (err error) {
 	if as, err = c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Update(context.TODO(), as, v1.UpdateOptions{}); err != nil {
 		return errors.WithStack(err)
 	}
-	ctx := context.Background()
+	iv := getEnvAsIntOrDefault("ACTIONSET_TIMEOUT", 30)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(iv)*time.Second)
+	defer cancel()
 	ctx = field.Context(ctx, consts.ActionsetNameKey, as.GetName())
 	for i := range as.Status.Actions {
 		if err = c.runAction(ctx, as, i); err != nil {
```
A-kanksh-a, line 361: so here we are defining a context `ctx` and calling `runAction(ctx, as, i)`. `runAction()` has some synchronous execution, we only wait for that result in line 361. Apart from that `runAction()` method execute phases in a goroutine asynchronously(line 414) and we don't wait for that goroutine to finish. the goroutines also use the same context.
 So here the `handleActionSet()` completes the execution and calls `cancel()` on context even before asynchronous part of execution that uses the same context `ctx` completes. 
 cc @pavannd1

PR #1066, opened 2021-08-09
pvc migration example

michaelcourcy:
## Change Overview

Migrate a PVC from a kubernetes cluster to another one 

We'll demonstrate how to migrate a PVC from a cluster (source cluster ) to another one (destination cluster) with Kanister.

This example was tested to migrate a PVC from an openshift 3.9 cluster to an openshift 4.7 cluster.
It's for this reason we install kanister in version 0.61.0 because it's the last kanister version still  compatible with openshift 3.9. If you don't have this constraint you can install kanister in the latest version.

the PVC must not be mounted by any pods because the blueprint use the `CopyVolumeData`, `PrepareData` and `RestoreData` functions.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


examples/migrate-pvc/README.MD
```diff
@@ -0,0 +1,125 @@
+# Migrate a PVC from a kubernetes cluster to another one 
+
```
viveksinghggits, line 2: I was thinking should the title be, Migrating volume data  or PVC data from a cluster to another.

viveksinghggits: if we change this, we will also have to change the dir name where this README is at.

examples/migrate-pvc/README.MD
```diff
@@ -0,0 +1,125 @@
...
+version.
+
+## Requirement 
+
+The PVC must not be mounted by any pods during the backup or during the restore.
+
+## Prerequisite  
+
+
+### Install kanister on both cluster
```
viveksinghggits, line 18: instead of giving the command to install kanister, and kanctl can be just mention the prerequisite like we are doing for apps in `examples/stable`.

examples/migrate-pvc/README.MD
```diff
@@ -0,0 +1,125 @@
...
+```
+
+PVC content has been sent to the S3 bucket defined in default-profile.
+
+### Prepare a restore-action for the destination cluster.
+
+Create a dry-run restore action that will execute in the destination cluster.
+
+```
+kanctl --namespace kanister create actionset -a restore --from backup-mvwhh --dry-run > restore-backup-mvwhh.yaml
```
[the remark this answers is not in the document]
michaelcourcy: No that won't work, because this action set is built from an existing action set in the source cluster. Once the user have changed his context to the destination cluster it won't be able to generate it anymore as this "from" action set does not exist in the destination cluster.

examples/migrate-pvc/README.MD
michaelcourcy: You want to add "volume data" because you think there could be some confusion about just migrating the spec of the pvc without the data that come along ? If that's the case I would prefer saying "Migrating the volume data **of** a PVC from a cluster to another". This is what we're doing actually, we are not migrating the pvc spec, we recreate manually an empty pvc on the destination cluster. Does it work for you ?

examples/migrate-pvc/README.MD
michaelcourcy: ok

PR #1074, opened 2021-08-25
Change the default duration for assuming an AWS role from 90m to 60m 

onkarbhat:
## Change Overview

Change the default duration for assuming an AWS role from 90m to 60m  since newly created IAM role's have a default max duration of 60m.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -52,7 +52,7 @@ const (
 	roleARNEnvKey                  = "AWS_ROLE_ARN"
 
 	// TODO: Make this configurable via `config`
-	AssumeRoleDurationDefault = 90 * time.Minute
+	AssumeRoleDurationDefault = 60 * time.Minute
 	AssumeRoleDuration        = "assumeRoleDuration"
 )
 


DeepikaDixit (approved):

vkamra:
@onkarbhat - what is the default used in AWS SDK when one is not specified? Should we use that - since it matches previous behavior (and override in specific cases - like Kopia)?

The other question is that if we are going to use a higher value for Kopia operations - isn't it better to catch the issue earlier during profile creation/validation?

I may not be able to re-review - so please don't block behind this if we need something for the release.

onkarbhat:
@vkamra  - The default is 60m when no duration is specified. See the row for AssumeRoleWithWebIdentity here - https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use.html#id_roles_use_view-role-max-session . I am not sure why the previous default in Kanister was 90 minutes.  Based on the information in this link, and also considering that the max duration for a new IAM role also default to 60 minutes, the default of 60 minutes in Kanister makes sense. 

Yes, improving profile validation will be helpful. It will require that we get the settings (including the max session duration ) for an IAM role and compare that with the default/configured duration.  We will have to check if this requires that a user/role be setup with the necessary permissions to query for IAM roles though - I'll confirm if this is needed. But we can add this later ( should not block this change )

PR #1128, opened 2021-11-01
Copy volumeSnapshotClass parameters when cloning

bathina2:
## Change Overview

Copy parameters when cloning.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/snapshot/snapshot_beta.go
+++ b/pkg/kube/snapshot/snapshot_beta.go
@@ -416,6 +416,8 @@ func UnstructuredVolumeSnapshotClass(gvr schema.GroupVersionResource, name, driv
 	}
 }
 
+// Mss2msi takes a map of string:string and returns a string:inteface map.
+// This is useful since the unstructured type take map[string]interface{} as values.
 func Mss2msi(in map[string]string) map[string]interface{} {
 	if in == nil {
 		return nil
--- a/pkg/kube/snapshot/snapshot_beta.go
+++ b/pkg/kube/snapshot/snapshot_beta.go
@@ -63,7 +63,7 @@ func cloneSnapshotClass(ctx context.Context, dynCli dynamic.Interface, snapClass
 	for _, key := range excludeAnnotations {
 		delete(existingAnnotations, key)
 	}
-	usNew := UnstructuredVolumeSnapshotClass(snapClassGVR, targetClassName, sourceSnapClass.Driver, newDeletionPolicy)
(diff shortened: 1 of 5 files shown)


pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -411,6 +411,18 @@ func UnstructuredVolumeSnapshotClass(gvr schema.GroupVersionResource, name, driv
 			},
 			VolSnapClassBetaDriverKey: driver,
 			"deletionPolicy":          deletionPolicy,
+			"parameters":              Mss2msi(params),
 		},
 	}
 }
+
+func Mss2msi(in map[string]string) map[string]interface{} {
```
PrasadG193, line 419: Can you please add comment to the function? It will help understand its purpose.

PrasadG193 (commented):
Overall looks good

pkg/kube/snapshot/snapshot_beta.go
bathina2: Added a comment to clarify

onkarbhat (approved):
LGTM

pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -63,7 +63,7 @@ func cloneSnapshotClass(ctx context.Context, dynCli dynamic.Interface, snapClass
 	for _, key := range excludeAnnotations {
 		delete(existingAnnotations, key)
 	}
-	usNew := UnstructuredVolumeSnapshotClass(snapClassGVR, targetClassName, sourceSnapClass.Driver, newDeletionPolicy)
+	usNew := UnstructuredVolumeSnapshotClass(snapClassGVR, targetClassName, sourceSnapClass.Driver, newDeletionPolicy, sourceSnapClass.Parameters)
```
viveksinghggits, line 66: Hi @bathina2,
By introducing this argument it looks like we are assuming `params` field will always be there in the `VolumeSnapshotClass`. But I have seen some examples where params field is not present in the `VoluemSnapshotClass` resoruce, and because of that when we try to create the copied `VolumeSnapshotClass` with `nil` as parameter value VolumeSnapshotClass Creation fails with below error 
```
"k10-clone-csi-hostpath-snapclass" is invalid: parameters: Invalid value: "null": parameters in body must be of type object: "null"
```
I was thinking we shouldn't pass the parameters field if the values is nil.

PR #1138, opened 2021-11-12
Functionality to call ingress operations after discovering APIVersion

viveksinghggits:
## Change Overview

To support k8s version 1.17 and 1.22 we will have to make sure that we are getting the ingress resources from the api version that is present on the cluster, for example we can not get ingresses from api version networking/v1 on k8s versions 1.22.
This PR adds wrappers that can be used to figure out which api version ingresses are part and then call the helper methods to eventually call the verb methods (get/list) from respective apis.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/discover.go
+++ b/pkg/kube/discover.go
@@ -41,11 +41,12 @@ func IsOSRouteGroupAvailable(ctx context.Context, cli discovery.DiscoveryInterfa
 }
 
 func IsResAvailableInGroupVersion(ctx context.Context, cli discovery.DiscoveryInterface, groupName, version, resource string) (bool, error) {
-	gv := fmt.Sprintf("%s/%s", groupName, version)
 	resList, err := cli.ServerPreferredResources()
 	if err != nil {
 		return false, err
 	}
+
+	gv := fmt.Sprintf("%s/%s", groupName, version)
 	for _, res := range resList {
 		for _, r := range res.APIResources {
 			if r.Name == resource && gv == res.GroupVersion {
--- a/pkg/kube/discover.go
+++ b/pkg/kube/discover.go
@@ -41,13 +41,14 @@ func IsOSRouteGroupAvailable(ctx context.Context, cli discovery.DiscoveryInterfa
 }
 
 func IsResAvailableInGroupVersion(ctx context.Context, cli discovery.DiscoveryInterface, groupName, version, resource string) (bool, error) {
+	gv := fmt.Sprintf("%s/%s", groupName, version)
 	resList, err := cli.ServerPreferredResources()
 	if err != nil {
 		return false, err
 	}
 	for _, res := range resList {
 		for _, r := range res.APIResources {
-			if r.Name == resource && fmt.Sprintf("%s/%s", groupName, version) == res.GroupVersion {
+			if r.Name == resource && gv == res.GroupVersion {
 				return true, nil
 			}
 		}
--- a/pkg/kube/discover.go
+++ b/pkg/kube/discover.go
@@ -40,6 +40,21 @@ func IsOSRouteGroupAvailable(ctx context.Context, cli discovery.DiscoveryInterfa
 	return false, nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/kube/discover.go
```diff
@@ -40,6 +40,21 @@ func IsOSRouteGroupAvailable(ctx context.Context, cli discovery.DiscoveryInterfa
...
 }
 
+func IsResAvailableInGroupVersion(ctx context.Context, cli discovery.DiscoveryInterface, groupName, version, resource string) (bool, error) {
+	resList, err := cli.ServerPreferredResources()
+	if err != nil {
+		return false, err
+	}
+	for _, res := range resList {
+		for _, r := range res.APIResources {
+			if r.Name == resource && fmt.Sprintf("%s/%s", groupName, version) == res.GroupVersion {
```
denisvmedia, line 50: I think, we can have a minor improvement here: `fmt.Sprintf("%s/%s", groupName, version)` is called multiple times, but always evaluates to the same value.

viveksinghggits: Hi @denisvmedia,
Sorry, I didn't get that. `fmt.Sprintf("%s/%s", groupName, version)` will result into different values in every iteration of the loop.

viveksinghggits: aah, hold on.

denisvmedia: Hm... are you sure? `groupName` and `version` are passed as string arguments in the function. Therefore they seem to remain unchanged for every iteration of the loop.

pkg/kube/ingress/ingress_extbeta.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/pkg/errors"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+)
+
+type IngressExtBeta struct {
```
pavannd1, line 29: As suggested below, I believe the struct and interface names are better when they have complete words describing them. We could use short forms for instances. This could be named `ExtensionsV1beta1` similar to k8s pkg client. It would easier to understand the purpose.
Also, add a comment saying `// <Struct> implements ingress.Manager interface`

pkg/kube/ingress/ingress_netbeta.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/pkg/errors"
+	netv1beta1 "k8s.io/api/networking/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+)
+
+type IngressNetBeta struct {
```
pavannd1, line 29: Same here. Could be `NetworkingV1beta1`

pkg/kube/ingress/ingress_netv1.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/pkg/errors"
+	netv1 "k8s.io/api/networking/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+)
+
+type IngressNetV1 struct {
```
pavannd1, line 29: And `NetworkingV1` here

pavannd1 (approved):
LGTM

PR #1145, opened 2021-11-19
Don't use params field, if they are nil, while cloning volumesnapshotclass

viveksinghggits:
## Change Overview

While cloning `VolumeSnapshotClass`, as part of https://github.com/kanisterio/kanister/pull/1128
we started passing params. But we should not pass params if the values
is `nil`, because in that case creation `VolumeSnapshotClass` fails with
`nil` not allowed.

## Pull request type

- [x] :bug: Bugfix

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual

Updated *** with kanister commit id and then updated the image in the env where we have been facing problem to make sure things have been resolved.

### DIFF
--- a/pkg/kube/snapshot/snapshot_alpha.go
+++ b/pkg/kube/snapshot/snapshot_alpha.go
@@ -422,30 +422,22 @@ func UnstructuredVolumeSnapshotContentAlpha(name, snapshotName, snapshotNs, dele
 }
 
 func UnstructuredVolumeSnapshotClassAlpha(name, driver, deletionPolicy string, params map[string]string) *unstructured.Unstructured {
-	if params == nil {
-		return &unstructured.Unstructured{
-			Object: map[string]interface{}{
-				"apiVersion": fmt.Sprintf("%s/%s", v1alpha1.GroupName, v1alpha1.Version),
-				"kind":       VolSnapClassKind,
-				"metadata": map[string]interface{}{
-					"name": name,
-				},
-				VolSnapClassAlphaDriverKey: driver,
-				"deletionPolicy":           deletionPolicy,
-			},
-		}
-	}
-	return &unstructured.Unstructured{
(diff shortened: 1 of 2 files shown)


pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -422,6 +422,19 @@ func UnstructuredVolumeSnapshotContentAlpha(name, snapshotName, snapshotNs, dele
...
+				"apiVersion": fmt.Sprintf("%s/%s", v1alpha1.GroupName, v1alpha1.Version),
+				"kind":       VolSnapClassKind,
+				"metadata": map[string]interface{}{
+					"name": name,
+				},
+				VolSnapClassAlphaDriverKey: driver,
+				"deletionPolicy":           deletionPolicy,
+			},
+		}
+	}
```
onkarbhat, line 437: @viveksinghggits - There is a lot of common code in the map being constructed. Can we refactor this so that we prepare the map without the "parameters" first and then add the parameters if params is non nil ?

pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -402,6 +402,19 @@ func UnstructuredVolumeSnapshotContent(gvr schema.GroupVersionResource, name, sn
 }
 
 func UnstructuredVolumeSnapshotClass(gvr schema.GroupVersionResource, name, driver, deletionPolicy string, params map[string]string) *unstructured.Unstructured {
+	if params == nil {
```
onkarbhat, line 405: Same comment as above.

onkarbhat (changes requested):

onkarbhat (approved):
LGTM

bathina2 (approved):

bathina2:
@viveksinghggits  thanks for catching this.

PR #1156, opened 2021-12-14
Fallback to `extensions/v1beta1` for ingress APIs in case of fake CLI

viveksinghggits:
## Change Overview

While creating ingress manager to interact with ingress resources based on apiVersion, we were failing when fake kubeclient was passed.
This PR tries to default to `extensions/v1beta1` in that case.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/ingress/ingressmgr.go
+++ b/pkg/kube/ingress/ingressmgr.go
@@ -58,7 +58,7 @@ func NewManager(ctx context.Context, kubeCli kubernetes.Interface) (Manager, err
 	if err != nil {
 		return nil, errors.New("Error finding if the CLI was fake CLI")
 	}
-	if len(list) == 0 {
+	if list == nil {
 		return nil, nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/kube/ingress/ingressmgr.go
```diff
@@ -53,6 +53,15 @@ type Manager interface {
 // NewManager can be used to get the Manager based on the APIVersion of the ingress resources on the cluster
 // so that, respecitve methods from that APIVersion can be called
 func NewManager(ctx context.Context, kubeCli kubernetes.Interface) (Manager, error) {
+	// return early if its fake CLI because fakecli doesn't implement `ServerPreferredResources` correctly
```
pavannd1, line 56:
```suggestion
	// Kubernetes fake CLI  doesn't implement `ServerPreferredResources`. Return early when used.
```

pkg/kube/ingress/ingressmgr.go
```diff
@@ -53,6 +53,15 @@ type Manager interface {
...
+	list, err := kubeCli.Discovery().ServerPreferredResources()
+	if err != nil {
+		return nil, errors.New("Error finding if the CLI was fake CLI")
```
pavannd1, line 59:
```suggestion
		return nil, errors.Wrap(err, "Failed to discover the server preferred resources")
```
Not sure if we should mention fake cli in production code.

chaitanya-baraskar (approved):

PR #1177, opened 2022-01-08
Changes to pick largest Restore Size 

bathina2:
## Change Overview

When creating a PVC from a Snapshot, if the restoreSize is specified we use it instead of the Snapshots restore size. 
This leads to issues if the specified restoreSize is less than the snapshots restore size.

This change picks the greater of both and uses that to create the resulting PVC.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/volume/volume.go
+++ b/pkg/kube/volume/volume.go
@@ -119,25 +119,9 @@ type CreatePVCFromSnapshotArgs struct {
 // CreatePVCFromSnapshot will restore a volume and returns the resulting
 // PersistentVolumeClaim and any error that happened in the process.
 func CreatePVCFromSnapshot(ctx context.Context, args *CreatePVCFromSnapshotArgs) (string, error) {
-	var size *resource.Quantity
-	if args.RestoreSize == "" {
-		sns, err := snapshot.NewSnapshotter(args.KubeCli, args.DynCli)
-		if err != nil {
-			return "", err
-		}
-		snap, err := sns.Get(ctx, args.SnapshotName, args.Namespace)
-		if err != nil {
-			return "", err
-		}
-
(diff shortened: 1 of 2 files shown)


pkg/kube/volume/volume.go
```diff
@@ -183,6 +167,41 @@ func CreatePVCFromSnapshot(ctx context.Context, args *CreatePVCFromSnapshotArgs)
...
+		return nil, errors.Wrap(err, "Failed to get snapshot")
+	}
+	if snap.Status != nil && snap.Status.RestoreSize != nil {
+		quantities = append(quantities, snap.Status.RestoreSize)
+	}
+	if args.RestoreSize != "" {
+		s, err := resource.ParseQuantity(args.RestoreSize)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse quantity (%s)", args.RestoreSize)
+		}
```
carlbraganza, line 188: Why fail here if there is a possibility that we have additional values in `quantities`?

pkg/kube/volume/volume.go
```diff
@@ -183,6 +167,41 @@ func CreatePVCFromSnapshot(ctx context.Context, args *CreatePVCFromSnapshotArgs)
...
+		quantities = append(quantities, &s)
+	}
```
carlbraganza, line 190: Should probably do this before attempting `sns.Get`.

carlbraganza (approved):

PR #1199, opened 2022-01-20
Update KubeOps to support core group & wait for delete completion

shlokchaudhari9:
## Change Overview

This PR consists of updates to KubeOps function to support create & delete for core group resources

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #1204 

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


helm/kanister-operator/templates/rbac.yaml
```diff
@@ -16,8 +16,10 @@ rules:
   - ""
   resources:
```
shlokchaudhari9, line 17: @PrasadG193 Should we use * here? Since we claim create and delete of **Kubernetes resources** on docs.

PrasadG193: @shlokchaudhari9 in my opinion, we should not give access by default. If someone wants to do that, they can update the clusterroles accordingly. We can add a note in docs for that

pkg/function/kubeops_test.go
```diff
@@ -82,6 +82,18 @@ metadata:
 spec:
   deploymentName: example-foo
   replicas: 1`
+
+	pvcSpec = `apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: test-pvc
+spec:
+  storageClassName: do-block-storage
```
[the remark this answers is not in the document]
shlokchaudhari9: Done

pkg/function/kubeops_test.go
```diff
@@ -74,6 +74,18 @@ spec:
...
+kind: Service
+metadata:
+  name: test-svc
+spec:
+  selector:
+    app: MyApp
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 9376`
```
PrasadG193, line 87: Can't we use the existing service declared above `serviceSpec`?

pkg/kube/kubectl.go
```diff
@@ -110,10 +110,11 @@ func (k *KubectlOperation) Delete(objRef crv1alpha1.ObjectReference, namespace s
...
-	err := k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Delete(context.Background(), objRef.Name, metav1.DeleteOptions{})
-	if apierrors.IsNotFound(err) {
-		return &objRef, nil
+	_ = k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Delete(context.Background(), objRef.Name, metav1.DeleteOptions{})
+	for {
+		_, err := k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Get(context.Background(), objRef.Name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return &objRef, nil
+		}
 	}
```
PrasadG193, line 119: Let's use a polling mechanism here instead of an infinite loop. This is not the optimized way to wait for certain condition to be met

shlokchaudhari9:
@PrasadG193 Please have a look

pkg/kube/kubectl.go
```diff
@@ -103,17 +104,21 @@ func (k *KubectlOperation) Create(spec io.Reader, namespace string) (*crv1alpha1
...
 		namespace = metav1.NamespaceDefault
 	}
 	if objRef.Namespace != "" {
 		namespace = objRef.Namespace
 	}
-	err := k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Delete(context.Background(), objRef.Name, metav1.DeleteOptions{})
-	if apierrors.IsNotFound(err) {
-		return &objRef, nil
-	}
+	_ = k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Delete(ctx, objRef.Name, metav1.DeleteOptions{})
```
PrasadG193, line 114: We should handle returned error here as well

pkg/kube/kubectl.go
```diff
@@ -103,17 +104,21 @@ func (k *KubectlOperation) Create(spec io.Reader, namespace string) (*crv1alpha1
...
 
+	err := poll.Wait(ctx, func(context.Context) (done bool, err error) {
```
PrasadG193, line 116: Lets add a comment summarizing what we are waiting for

pkg/kube/kubectl.go
```diff
@@ -103,17 +104,29 @@ func (k *KubectlOperation) Create(spec io.Reader, namespace string) (*crv1alpha1
...
+	if err := poll.Wait(ctx, func(context.Context) (done bool, err error) {
+		_, err = k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Get(ctx, objRef.Name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}); err != nil {
+		return &objRef, err
+	}
+	return &objRef, nil
```
PrasadG193, line 131:
```suggestion
	err := poll.Wait(ctx, func(context.Context) (done bool, err error) {
		_, err = k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Get(ctx, objRef.Name, metav1.GetOptions{})
		if apierrors.IsNotFound(err) {
			return true, nil
		}
		return false, err
	})
	return &objRef, err
```
we can just return the error, the check is redundant

pkg/kube/kubectl.go
```diff
@@ -103,17 +104,29 @@ func (k *KubectlOperation) Create(spec io.Reader, namespace string) (*crv1alpha1
 }
 
 // Delete k8s resource referred by objectReference
-func (k *KubectlOperation) Delete(objRef crv1alpha1.ObjectReference, namespace string) (*crv1alpha1.ObjectReference, error) {
+func (k *KubectlOperation) Delete(ctx context.Context, objRef crv1alpha1.ObjectReference, namespace string) (*crv1alpha1.ObjectReference, error) {
```
PrasadG193, line 107: We should also pass context to Create() method as well. But as per my understanding, ` k.factory.NewBuilder()` method doesn't support handling context. We might need to handle the context in Create()

shlokchaudhari9: I don't think we can add context to Create() method if we want to create resource using `k.factory.NewBuilder()`
Following are the methods being called by Create() at this line:

```bash
func (m *Helper) Create(namespace string, modify bool, obj runtime.Object) (runtime.Object, error) {
	return m.CreateWithOptions(namespace, modify, obj, nil)
}

func (m *Helper) CreateWithOptions(namespace string, modify bool, obj runtime.Object, options *metav1.CreateOptions) (runtime.Object, error) {
	if options == nil {
		options = &metav1.CreateOptions{}
	}
	if m.ServerDryRun {
		options.DryRun = []string{metav1.DryRunAll}
	}
	if m.FieldManager != "" {
		options.FieldManager = m.FieldManager
	}
	if modify {
		// Attempt to version the object based on client logic.
		version, err := metadataAccessor.ResourceVersion(obj)
		if err != nil {
			// We don't know how to clear the version on this object, so send it to the server as is
			return m.createResource(m.RESTClient, m.Resource, namespace, obj, options)
		}
		if version != "" {
			if err := metadataAccessor.SetResourceVersion(obj, ""); err != nil {
				return nil, err
			}
		}
	}

	return m.createResource(m.RESTClient, m.Resource, namespace, obj, options)
}

func (m *Helper) createResource(c RESTClient, resource, namespace string, obj runtime.Object, options *metav1.CreateOptions) (runtime.Object, error) {
	return c.Post().
		NamespaceIfScoped(namespace, m.NamespaceScoped).
		Resource(resource).
		VersionedParams(options, metav1.ParameterCodec).
		Body(obj).
		Do(context.TODO()).
		Get()
}
```
There is no argument that accepts `ctx` from the method call

shlokchaudhari9: We might need to explore a different way to create resources that also takes `context.Context` as argument

shlokchaudhari9: As discussed we will make changes w.r.t. this improvement as a separate PR. Filed a bug here https://github.com/kanisterio/kanister/issues/1205

pkg/function/kubeops_test.go
```diff
@@ -241,6 +241,41 @@ func (s *KubeOpsSuite) TestKubeOps(c *C) {
...
 
+func (s *KubeOpsSuite) TestKubeOpsCreateDeleteWithCoreResource(c *C) {
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+	tp := param.TemplateParams{}
+	action := "test"
+	gvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}
+	serviceName := "test-deployment-2"
+
+	bp := newCreateResourceBlueprint(deletePhase(gvr, serviceName, s.namespace))
```
pavannd1, line 252: It seems to me that this test is attempting deletion of a resource created in the previous test. It means that one must always run the previous test before running this one which is not a good test design. When run by itself, this test will always fail.
Please modify this to create and delete the resource so that it can be run independently

pkg/kube/kubectl.go
pavannd1: Agree. Will be helpful later

pkg/kube/kubectl.go
```diff
@@ -103,17 +104,27 @@ func (k *KubectlOperation) Create(spec io.Reader, namespace string) (*crv1alpha1
 }
 
 // Delete k8s resource referred by objectReference
```
pavannd1, line 106: Update the comment to mention wait for deletion

pkg/kube/kubectl.go
```diff
@@ -102,18 +103,29 @@ func (k *KubectlOperation) Create(spec io.Reader, namespace string) (*crv1alpha1
...
-		return &objRef, nil
+	err := k.dynCli.Resource(schema.GroupVersionResource{Group: objRef.Group, Version: objRef.APIVersion, Resource: objRef.Resource}).Namespace(namespace).Delete(ctx, objRef.Name, metav1.DeleteOptions{})
+	if err != nil {
+		return &objRef, err
 	}
+	return waitForResourceDeletion(ctx, k, objRef, namespace)
+}
 
+// Wait for resource to get deleted by repeatedly checking for NotFound error
+func waitForResourceDeletion(ctx context.Context, k *KubectlOperation, objRef crv1alpha1.ObjectReference, namespace string) (*crv1alpha1.ObjectReference, error) {
```
pavannd1, line 122:
```suggestion
// waitForResourceDeletion waits for resource to get deleted by repeatedly checking for NotFound error
func waitForResourceDeletion(ctx context.Context, k *KubectlOperation, objRef crv1alpha1.ObjectReference, namespace string) (*crv1alpha1.ObjectReference, error) {
```
General golang convention is to start the comment with the function name we are describing

pavannd1 (approved):
LGTM

PR #1200, opened 2022-01-21
Redact commands and args while logging Pod manifest

PrasadG193:
## Change Overview

This PR adds a helper function that can be used to log pod object manifest safely by hiding the commands and args field values which can hold sensitive info.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

```
$ go test -v -check.f="TestSafeDumpPodObject" .
=== RUN   Test
OK: 1 passed
--- PASS: Test (0.00s)
PASS
ok      github.com/kanisterio/kanister/pkg/log  0.008s
```


pavannd1 (approved):
LGTM

pkg/log/log.go
```diff
@@ -236,3 +239,20 @@ func entryToJSON(entry *logrus.Entry) []byte {
...
 	return bytes
 }
+
+// SafeDumpPodObject redacts commands and args in Pod manifest to hide sensitive info,
+// converts Pod object into string and returns it
+func SafeDumpPodObject(pod *v1.Pod) string {
+	if pod == nil {
+		return ""
+	}
+	for i, _ := range pod.Spec.Containers {
```
pavannd1, line 249:
```suggestion
	for i := range pod.Spec.Containers {
```
caught by linter:
```
pkg/log/log.go:249:9: S1005: should omit value from range; this loop is equivalent to `for i := range ...` (gosimple)
	for i, _ := range pod.Spec.Containers {
```

pkg/log/log.go
```diff
@@ -236,3 +239,20 @@ func entryToJSON(entry *logrus.Entry) []byte {
 
 	return bytes
 }
+
+// SafeDumpPodObject redacts commands and args in Pod manifest to hide sensitive info,
+// converts Pod object into string and returns it
+func SafeDumpPodObject(pod *v1.Pod) string {
```
julio-lopez, line 245: This function modifies `pod`. Either:
- it should be noted in the function documentation; or
- the function should create a deep copy.

This is not an issue in the current usage of this function. But, if the function were to be reused elsewhere, this is surprising behavior and it may actually introduce software defects (a.k.a bugs).

PrasadG193: That makes sense. Thanks a lot for letting me know. I will keep that in mind. Also, will fix this in the follow-up

Issue #1126, opened 2021-10-28
Kanister Functions for creating CSI snapshots


pavannd1:
Add new Kanister functions to create, restore and delete Volume Snapshots via CSI

PR #1208, opened 2022-01-24
Add new integration test for CSI Snapshot Blueprint

shlokchaudhari9:
## Change Overview

This PR consists of an integration test for the CSI Snapshot example Blueprint

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues

- #1126 

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


viveksinghggits:
Let's divide this PR into two PRs, one to add enhancements into how we install k10 to install CSI host drive and other one to actually implement the test.

build/local_kubernetes.sh
```diff
@@ -59,6 +60,29 @@ start_localkube() {
     kind get kubeconfig --name="kanister" > "${HOME}/.kube/config"
     wait_for_nodes
     wait_for_pods
+
+    SNAPSHOTTER_VERSION=v5.0.0
+
+    # Install VolumeSnapshot CRDs
+    kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/client/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+    kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/client/config/crd/snapshot.storage.k8s.io_volumesnapshotcontents.yaml
+    kubectl apply -f https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/client/config/crd/snapshot.storage.k8s.io_volumesnapshots.yaml
```
viveksinghggits, line 69: let's see if we can use one kubectl statement to create all these resources. Can a syntax like this be used
https://github.com/digitalocean/csi-digitalocean#3-deploy-the-csi-plugin-and-sidecars

shlokchaudhari9: Yeah. We can merge the kubectl commands in the way done on csi-digitalocean.

shlokchaudhari9:
@PrasadG193 @viveksinghggits @pavannd1 This PR ready to be reviewed

pkg/app/csi-snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (csi CSISnapshot) getAppPersistentVolumeClaimObj() (*v1.PersistentVolumeClaim, error) {
+	pvcManifest := `apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: time-log-pvc
+  labels:
+    app: time-logger
+spec:
+  storageClassName: csi-hostpath-sc
```
viveksinghggits, line 216: @shlokchaudhari9 I somehow think that we should change the `csi-hostpath-sc` to default storage class and then use that automatically instead of manually specifying the storage class here.
What I have concluded is, the other database applications are somehow not working with `csi-hostpath-sc`, can we figure that out? If we can figure out why those are not working with `csi-hostpath-sc`, we can set this as default storage class and then we will not have to set the storeageclass anywhere else.

shlokchaudhari9: @viveksinghggits This is the issue I am facing on all the database apps https://github.com/kubernetes/minikube/issues/13098

The problem is that, database apps in our tests are being run by non-root users (some helm charts explicitly forbid running their containers as root) That's why we get permission denied error in most of them or helm chart gets timed-out during installation.

I also found out why only CSISnapshot (Time logger) integration test passes, while other apps fail. Turns out, the image used in time-logger deployment is being run as a root user. Hence, we do not get permission denied error when using the mount path within the container.

In the attached issue thread, many people have suggested switching the volume provisioner. Which is what we are doing by not setting the hostpath storageClass as default. We run the CSISnapshot (Time logger) integration test explicitly by mentioning the storageClassName in PVC, while the other database apps automatically use standard volume provisioner.

viveksinghggits: makes sense. We can get this in.

viveksinghggits: I will re-review.

pkg/app/csi-snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes"
+
+	crv1alpha1 "github.com/kanisterio/kanister/pkg/apis/cr/v1alpha1"
+	"github.com/kanisterio/kanister/pkg/field"
+	"github.com/kanisterio/kanister/pkg/kube"
+	"github.com/kanisterio/kanister/pkg/log"
+)
+
+type CSISnapshot struct {
```
pavannd1, line 33: I wonder if we should rename this application. It's an application used to test CSISnapshot but the snapshot itself is not the application. I would recommend something like TimeLogCSI

pkg/app/csi-snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+	podname, containername, err := kube.GetPodContainerFromDeployment(ctx, csi.cli, csi.namespace, csi.name)
+	if err != nil || podname == "" {
+		return "", errors.Wrapf(err, "Error getting pod and containername %s.", csi.name)
+	}
+	_, stderr, err := kube.Exec(csi.cli, csi.namespace, podname, containername, command, nil)
+	return stderr, err
+}
+
+func (csi CSISnapshot) getAppDeploymentObj() (*appsv1.Deployment, error) {
+	deploymentManifest := `apiVersion: apps/v1
```
pavannd1, line 176: I think we should move these to Go structs. Maintaining raw strings is hard and is vulnerable to typos.

pkg/app/csi-snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+          claimName: time-log-pvc`
+
+	var deployment *appsv1.Deployment
+	err := yaml.Unmarshal([]byte(deploymentManifest), &deployment)
+	return deployment, err
+}
+
+func (csi CSISnapshot) getAppPersistentVolumeClaimObj() (*v1.PersistentVolumeClaim, error) {
+	pvcManifest := `apiVersion: v1
+kind: PersistentVolumeClaim
```
pavannd1, line 210: Same here. Move it Go 👍🏼

pkg/app/csi-snapshot.go
```diff
@@ -0,0 +1,226 @@
...
+
+type CSISnapshot struct {
+	cli        kubernetes.Interface
+	namespace  string
+	name       string
+	deployment appsv1.Deployment
+	pvc        v1.PersistentVolumeClaim
+}
+
+func NewCSISnapshot(name string) App {
```
pavannd1, line 41: Let's also start adding comments to exported functions and structs.

pkg/app/csi-snapshot.go
shlokchaudhari9: Cool Thanks

pavannd1 (approved):
LGTM

Issue #1207, opened 2022-01-24
Add webhook to validate bluerpint resources


viveksinghggits:
**Is your feature request related to a problem? Please describe.**
Right now blueprints can only be validated using kanctl command with the syntax

```
kanctl validate blueprint -f <blueprint-manifest>
```
But if someone doesn't do this manually and tries to create a blueprint, the incorrect blueprint might get created.

**Describe the solution you'd like**
We should install a validating webhook along with kanister controller to make sure the validation is done on the blueprint that is being even using kubectl. We can try to validate the same checks that we are doing when we run `kanctl validate blueprint`.

- [x] Add validating webhook
- [ ] Update docs

PR #1213, opened 2022-01-27
Add validating webhook controller for blueprint resource

viveksinghggits:
## Change Overview

Add validaating webhook for the blueprint resources that is going to do some basic check even before the blueprint is actually created in the cluster.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #1207

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

Deploy kanister from local repo and make sure invalid blueprints are not created

```
 » k create -f examples/stable/mysql/mysql-blueprint.yaml -n kanister
Error from server (Invalid blueprint, validation of phase restoreFromBlobStore in action restore, error: Required arg missing: command
): error when creating "examples/stable/mysql/mysql-blueprint.yaml": admission webhook "blueprints.cr.kanister.io" denied the request: Invalid blueprint, validation of phase restoreFromBlobStore in action restore, error: Required arg missing: command
```

Install kanister `0.72.0` and upgrade that to new changes make sure things are working.

- [ ] Update docs


pkg/handler/handler.go
```diff
@@ -36,8 +42,6 @@ type Info struct {
 	Version string `json:"version"`
 }
 
-var _ http.Handler = (*healthCheckHandler)(nil)
-
 type healthCheckHandler struct{}
 
 //NewHealthCheckHandler function returns pointer to an empty healthCheckHandler
```
pavannd1, line 47: Looks like this is no longer used anywhere. Can we remove it?

pkg/kancontroller/kancontroller.go
```diff
@@ -83,6 +71,11 @@ func Execute() {
 		return
 	}
 
+	err = handler.RunWebhookServer()
+	if err != nil {
+		log.WithError(err).Print("Failed to start kanister controller server")
```
pavannd1, line 76:
```suggestion
		log.WithError(err).Print("Failed to start validating webhook server")
```
Any reason for not specifying this?
And should we return here?

viveksinghggits: apart from webhook there were some other endpoints that we register that is the reason I said kanister controller. But I think validating webhook controller is better. I can do that.

go.mod
```diff
@@ -9,34 +9,76 @@ replace (
 	gopkg.in/check.v1 => github.com/kastenhq/check v0.0.0-20180626002341-0264cfcea734
 )
 
+require (
+	github.com/Azure/azure-sdk-for-go v54.0.0+incompatible
```
viveksinghggits, line 13: we see some of the un necessary changes here because I think my vs code grouped the dependencies in direct and indirect dependencies and I liked it this way.

pavannd1: Should we add comments about direct and indirect so that in the future people add new dependencies in the right list?

pkg/blueprint/validate/validate.go
```diff
@@ -33,14 +34,14 @@ func Do(bp *crv1alpha1.Blueprint, funcVersion string) error {
 		phases, err := kanister.GetPhases(*bp, name, funcVersion, param.TemplateParams{})
 		if err != nil {
 			utils.PrintStage(fmt.Sprintf("validation of action %s", name), utils.Fail)
-			return err
+			return errors.Wrapf(err, fmt.Sprintf("Failed to validate action %s", name))
```
pavannd1, line 37:
```suggestion
			return errors.Wrapf(err, "Failed to validate action %s", name)
```

pavannd1:
LGTM - minor comments

pavannd1 (approved):

PR #1223, opened 2022-02-04
Use CONTAINER_NAME if set

ewhamilton:
## Change Overview

Use CONTAINER_NAME if set instead of attempting to parse /proc/self/cgroup.

Supports cgroups v2 systems such as Docker Desktop for Mac 4.3+.

Approach of CONTAINER_NAME has been used successfully for K10. To maintain independence of Kanister from K10 in workflows that do not set CONTAINER_NAME, this change detects being in a container if either CONTAINER_NAME set or if detected from /proc/self/cgroup.

## Pull request type

- [x] :hamster: Trivial/Minor

### DIFF
--- a/build/package.sh
+++ b/build/package.sh
@@ -24,7 +24,10 @@ build_licenses_info_image() {
     local src_dir="$(pwd)"
     local target_file="$(pwd)/licenses"
     local mount_cmd="-v $(pwd):$(pwd)"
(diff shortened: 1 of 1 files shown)


miquella (commented):
LGTM :+1: Will let @pavannd1 weigh in though.

pavannd1 (approved):

Issue #1243, opened 2022-02-11
Kanister CRDs should be distributed with helm chart


viveksinghggits:
### Discussed in https://github.com/kanisterio/kanister/discussions/1232

<div type='discussions-op-text'>

<sup>Originally posted by **viveksinghggits** February  7, 2022</sup>
Kanister CRDs right now are distributed with kanister controller code and not the helm chart. Ideally we should try to distribute them with helm chart. 
In some scenarios people were not able to upgrade to newer version because they didn't want to give kanister access to manage CRDs and because of that upgrade to recent version failed.</div>

pavannd1:
@viveksinghggits We should probably add something to the docs too

viveksinghggits:
right, I will do that. thanks @pavannd1 .

viveksinghggits:
https://github.com/kanisterio/kanister/pull/1278

PR #1244, opened 2022-02-11
Distribute CRDs using helm chart

viveksinghggits:
## Change Overview

CRDs for now were being deployed using controller, we
should make sure CRDs are being distributed by helm
chart preferrably but if we are consuming kanister not
as sub chart in that case they can set an env var
in the kanister pod to make sure CRDs are being created.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #1243

## Test Plan

Helm install manifests

```
helm install kanister helm/kanister-operator/ -n kanister --dry-run > default.yaml
```
the manifest can be found [here](https://gist.github.com/74d2a849910bf27c2074f32a93019b4f).

with `createCRDs: true` 
```
helm install kanister helm/kanister-operator/ -n kanister --set controller.createCRDs=true  --dry-run > createcrds.yaml
```
the generated manifest can be found [here](https://gist.github.com/9f8f30473f78e0daaf800cb0b8f08b56)

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


ihcsim (commented):
Thanks for working on this. I am sort of hesitant to let Helm manage CRDs because by default, Helm can't upgrade nor delete CRDs in future chart versions. We are not gaining much other than the "one chart" installation experience, and now users become responsible for upgrading the CRDs.

I wonder if we should default `createCRDs` to `true` to continue let the controller manage the CRDs, with `false` being the 'break glass' approach. (Also see also comment below on upgrading existing Helm releases.) WDYT?

helm/kanister-operator/values.yaml
```diff
@@ -13,6 +13,10 @@ serviceAccount:
 controller:
   service:
     port: 443
+  # createCRDs specifies if kanister controller should create the CRDs
+  # false : CRDs would be created by helm
+  # true : CRDs would be created by kanister controller
+  createCRDs: false
```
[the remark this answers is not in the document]
viveksinghggits: So, even if we set the `createCRDs` field in the `values.yaml` to be true, our controller is going to update the CRDs which should be ok I think.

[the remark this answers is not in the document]
viveksinghggits: > That's assuming the controller has the required RBAC permissions to manage CRDs, right? Looking back at the slack [convo](https://kanisterio.slack.com/archives/C85C8V22J/p1639747585010400), I thought the issue started because users changed the controller's RBAC to not be able to manage CRDs.

Controller not having authz to manage the CRDs is the case where people are going to use `createCRDs : false`. And in that case helm will be responsible to create the CRDs.

> Also, I am more interested in how the default of createCRDs: false will affect existing users where their existing CRDs aren't managed by Helm. On their next helm upgrade, Helm might fail due to conflicting existing resources.

To test this I  installed released (latest) version of kanister and then upgraded that to locally chnaged chart with changes  `createCRDs :  false` and it was upgraded successfully without failure.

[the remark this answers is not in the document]
ihcsim: Thanks for confirming the behaviour of `createCRDs : false`. The inclusion of the CRD rules in `helm/kanister-operator/templates/rbac.yaml` confused me earlier.

Also, thanks for testing the upgrade using the latest version. Just out of curiosity, does that mean we can remove the controller-generated CRDs completely? This PR is going to default to using the Helm-managed CRDs anyway, and upgrade seems to work.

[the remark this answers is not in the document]
viveksinghggits: Hi @ihcsim,
Like we discussed in this thread https://kanisterio.slack.com/archives/C85C8V22J/p1644962226053899 we need to have `crds` in helm template as well controller independently should also be able to create CRDs for the cases where kanister is not being consumed as sub chart.

pkg/customresource/embed.go
```diff
@@ -0,0 +1,14 @@
...
+
+// embed.go embeds the CRD yamls (actionset, profile, blueprint) with the
+// controller binary so that we can read these manifests in runtime.
+
+// We need these manfiests at two places, at `pkg/customresource/` and at
+// `helm/kanister-operator/crds`. To make sure we are not duplicating the
+// things we have original files at `pkg/customresource` and have soft links
+// at `helm/kanister-operator/crds`.
+
+//go:embed actionset.yaml profile.yaml blueprint.yaml
```
miquella, line 13: It might make the list easier to maintain if it were sorted and on individual lines:
```suggestion
//go:embed actionset.yaml
//go:embed blueprint.yaml
//go:embed profile.yaml
```

helm/kanister-operator/values.yaml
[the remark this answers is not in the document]
miquella: One additional consideration is that I don't believe Helm will update the CRDs when they change. However, the controller's code will create **or update** the CRDs.

With this in mind, perhaps calling the option `updateCRDs` instead might make more sense.

It seems like this would make it more obvious how a consumer could adopt updates to the CRDs versus manually installing the updates (or knowing that they need to set `createCRDs=true` even though the CRDs are already installed).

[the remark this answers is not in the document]
viveksinghggits: thanks Mark.

miquella (approved):
Looks great to me!

pkg/customresource/customresource.go
```diff
@@ -114,15 +113,21 @@ func decodeSpecIntoObject(spec []byte, intoObj runtime.Object) error {
...
+		return errors.Wrap(err, "Getting CRD object from CRD bytes")
 	}
+
 	ctx := contextpkg.Background()
 	_, err = context.APIExtensionClientset.ApiextensionsV1().CustomResourceDefinitions().Create(ctx, crd, metav1.CreateOptions{})
 	if err != nil {
-		if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create %s CRD. %+v", resource.Name, err)
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("Failed to create %s CRD. %+v", resource.Name, err)
```
pavannd1, line 130:
```suggestion
			return errors.Errorf("Failed to create %s CRD. %+v", resource.Name, err)
```
Not your change but noticed that all the errors here are using fmt errors - perhaps create an issue to move all of them to errors pkg

pkg/kancontroller/kancontroller.go
```diff
@@ -112,3 +120,17 @@ func isCACertMounted() bool {
 
 	return true
 }
+
+func createOrUpdateCRDs() bool {
+	createOrUpdateCRD := os.Getenv(createOrUpdateCRDEnvVar)
+	if createOrUpdateCRD == "" {
+		return false
```
pavannd1, line 127: Shouldn't this be true by default? What happens to the current installations where the CRDs are handled by the controller?

pkg/kancontroller/kancontroller.go
```diff
@@ -112,3 +120,17 @@ func isCACertMounted() bool {
...
+	}
+	c, err := strconv.ParseBool(createOrUpdateCRD)
+	if err != nil {
+		log.Print("environment variable", field.M{"CREATE_CRDS": createOrUpdateCRD})
+		return false
+	}
+
+	return c
```
pavannd1, line 135: Could this be simplified to
```suggestion
	return createOrUpdateCRD == "true"
```

pkg/kancontroller/kancontroller.go
viveksinghggits: we decided to use the default to be `false` because we want CRDs to be primarily managed/installed by helm instead of kanister controller.
if the CRDs are already installed by  controller and we try to upgrade that release it is going to go through and release would be upgraded without any problem.

ihcsim: how much impact will this default have on existing users, who are currently relying on the controller to manage the CRDs? with this change, these users need to be made aware that by default, they will be responsible for manually upgrading the CRDs going forward because Helm doesn't do it for them, unless they change this configuration to `true` in all future upgrades.

pkg/kancontroller/kancontroller.go
```diff
@@ -112,3 +118,12 @@ func isCACertMounted() bool {
...
 	return true
 }
+
+func createOrUpdateCRDs() bool {
+	createOrUpdateCRD := os.Getenv(createOrUpdateCRDEnvVar)
+	if createOrUpdateCRD == "" {
+		return false
+	}
+
+	return createOrUpdateCRD == "true"
```
ihcsim, line 128: TIOLI: what about `True`, `TRUE`, `t`, `T`? Feels like the previous approach of using `strconv.ParseBool()` is more robust.

viveksinghggits: I agree, I am also inclined to change it back to what it was.

pkg/kancontroller/kancontroller.go
viveksinghggits: Should we document this @ihcsim ?

ihcsim: Per discussion in community meeting, we will keep the controller-managed approach as the default, to avoid disruption to existing upgrade path. User can use Helm to manage the CRDs by changing the `controller.updateCRDs` configuration to `false`.

viveksinghggits: I have  made the changes, by default controller is going to be responsible for managing CRDs.

pkg/kancontroller/kancontroller.go
```diff
@@ -112,3 +120,18 @@ func isCACertMounted() bool {
...
+func createOrUpdateCRDs() bool {
+	createOrUpdateCRD := os.Getenv(createOrUpdateCRDEnvVar)
+	if createOrUpdateCRD == "" {
+		return true
+	}
+
+	c, err := strconv.ParseBool(createOrUpdateCRD)
+	if err != nil {
+		log.Print("environment variable", field.M{"CREATEORUPDATE_CRDS": createOrUpdateCRD})
+		return false
```
ihcsim, line 133: We should also default this to `true`.

helm/kanister-operator/templates/deployment.yaml
```diff
@@ -26,6 +26,9 @@ spec:
         volumeMounts:
           - name: webhook-certs
             mountPath: /var/run/webhook/serving-cert
+        env:
+        - name: CREATEORUPDATE_CRDS
```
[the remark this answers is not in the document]
viveksinghggits: right, I also felt that but `createOrUpdateCRDs` in helm values wont be a good idea I think. But it makes sense in go code. That is the reason I decided to go with this.

pkg/customresource/customresource.go
```diff
@@ -114,15 +113,21 @@ func decodeSpecIntoObject(spec []byte, intoObj runtime.Object) error {
...
+		return errors.Wrap(err, "Getting CRD object from CRD bytes")
 	}
+
 	ctx := contextpkg.Background()
 	_, err = context.APIExtensionClientset.ApiextensionsV1().CustomResourceDefinitions().Create(ctx, crd, metav1.CreateOptions{})
 	if err != nil {
-		if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create %s CRD. %+v", resource.Name, err)
+		if !apierrors.IsAlreadyExists(err) {
+			return errors.Errorf("Failed to create %s CRD. %+v", resource.Name, err)
```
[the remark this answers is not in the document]
viveksinghggits: I think we should handle that as separate PR because we have same format in this entire source file. WDYT?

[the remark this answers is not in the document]
ihcsim: I see what you mean. I vouch for consistency. So this LGTM :+1:.

pavannd1 (approved):

PR #1249, opened 2022-02-16
Azure gov client changes

bathina2:
## Change Overview

The following PR takes the environment  name config value and uses it to set the other environment variables. It also allows the other provided config values to override these.

We also need to uses the `WithBaseURI` versions of calls to create clients. This will ensure the client points to the right endpoint.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/blockstorage/azure/client.go
+++ b/pkg/blockstorage/azure/client.go
@@ -79,7 +79,7 @@ func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
 
 	baseURI, ok = config[blockstorage.AzureResurceMgrEndpoint]
 	if !ok {
-		baseURI = compute.DefaultBaseURI
+		baseURI = env.ResourceManagerEndpoint
 	}
 
 	disksClient := compute.NewDisksClient(subscriptionID)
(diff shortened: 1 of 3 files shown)


pkg/blockstorage/azure/client.go
```diff
@@ -118,12 +116,12 @@ func getAuthorizer(env azure.Environment, config map[string]string) (*autorest.B
 	}
 
 	credConfig := auth.NewClientCredentialsConfig(clientID, clientSecret, tenantID)
-	if aDDEndpoint, ok := config[blockstorage.AzureActiveDirEndpoint]; ok {
-		credConfig.AADEndpoint = aDDEndpoint
+	if credConfig.AADEndpoint, ok = config[blockstorage.AzureActiveDirEndpoint]; !ok {
```
onkarbhat, line 119: Can we add unit tests to verify this logic where we default to env.Something if ok is false, and verify that we use the config passed in config[some field] when ok is true ?

onkarbhat (approved):
LGTM. Minor comment about a unit test.

pkg/blockstorage/azure/client.go
bathina2: Yea I'm trying to find a way to unit test this

PR #1260, opened 2022-02-23
Adding context to log statements

chaitanya-baraskar:
## Change Overview

Adding context to log statements. ActionSet name is being passed as context field. Updated log statements to use context so all relevant logs from ActionSet can be traced together.
Added feature where kanister picks labels with prefix `kanister.io/` for logging.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

For testing, I tried MsSQL blueprint - https://github.com/kanisterio/kanister/tree/master/examples/stable/mssql
Deployed application using steps given in README.md and tried backup restore command. Attaching gist of generated output.

Output before changes - https://gist.github.com/chaitanya-baraskar/4db6c3495905e43065163049c37f52d6
Output after changes in this PR - https://gist.github.com/chaitanya-baraskar/822a9148f0b4085fbf6da81a6f7cb4a6


pkg/controller/controller.go
```diff
@@ -218,20 +218,26 @@ func (c *Controller) onUpdateActionSet(oldAS, newAS *crv1alpha1.ActionSet) error
 		log.Print("Updated ActionSet", field.M{"ActionSetName": newAS.Name})
 		return err
 	}
+	ctx := context.TODO()
+	ctx = field.Context(ctx, consts.ActionsetNameKey, newAS.GetName())
+	// adding labels in the context as field for better logging
+	for key, value := range newAS.GetLabels() {
```
ihcsim, line 224: What are the benefits of logging the actionset labels? Since labels are something that user can change over time, will they actually want to aggregate these fields?

pkg/function/delete_volume_snapshot.go
```diff
@@ -79,15 +79,15 @@ func deleteVolumeSnapshot(ctx context.Context, cli kubernetes.Interface, namespa
 		snapshot, err := provider.SnapshotGet(ctx, pvcInfo.SnapshotID)
 		if err != nil {
 			if strings.Contains(err.Error(), SnapshotDoesNotExistError) {
-				log.Debug().Print("Snapshot already deleted", field.M{"SnapshotID": pvcInfo.SnapshotID})
+				log.WithContext(ctx).Print("Snapshot already deleted", field.M{"SnapshotID": pvcInfo.SnapshotID})
```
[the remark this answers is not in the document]
chaitanya-baraskar: @ihcsim As per code, I think this shouldn't be part of Debug. We are checking if snapshot exist or not, without Debug level log it will be difficult for user to track what's going wrong.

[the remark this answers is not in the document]
ihcsim: Agreed with your approach :+1:.

pkg/controller/controller.go
```diff
@@ -218,20 +218,26 @@ func (c *Controller) onUpdateActionSet(oldAS, newAS *crv1alpha1.ActionSet) error
 		log.Print("Updated ActionSet", field.M{"ActionSetName": newAS.Name})
 		return err
 	}
+	ctx := context.TODO()
+	ctx = field.Context(ctx, consts.ActionsetNameKey, newAS.GetName())
```
pavannd1, line 222: The action set name is useful in the function logs below but looks like all the logs in the controller already have it as a field. How do we avoid duplication? Should we update the logs to not explicitly log the name?

pavannd1, line 222:
```suggestion
	ctx := field.Context(context.Background(), consts.ActionsetNameKey, newAS.GetName())
```

chaitanya-baraskar: @pavannd1 I will update functions to not use ActionName in log.

ihcsim (approved):
LGTM :+1:!

pkg/function/kube_exec.go
```diff
@@ -96,8 +96,8 @@ func (kef *kubeExecFunc) Exec(ctx context.Context, tp param.TemplateParams, args
 	ctx = field.Context(ctx, consts.PodNameKey, pod)
 	_ = field.Context(ctx, consts.ContainerNameKey, container)
 	stdout, stderr, err := kube.Exec(cli, namespace, pod, container, cmd, nil)
-	format.Log(pod, container, stdout)
-	format.Log(pod, container, stderr)
+	format.LogWithCtx(ctx, pod, container, stdout)
```
pavannd1, line 99: I feel like the pod name and container name might get logged twice here since we are adding those to the context above on lines 96/97. Could you please confirm this with a blueprint using `KubeExec` function?

pkg/function/kube_exec_all.go
```diff
@@ -91,8 +91,8 @@ func execAll(ctx context.Context, cli kubernetes.Interface, namespace string, ps
 				ctx = field.Context(ctx, consts.PodNameKey, p)
 				ctx = field.Context(ctx, consts.ContainerNameKey, c)
 				stdout, stderr, err := kube.Exec(cli, namespace, p, c, cmd, nil)
-				format.Log(p, c, stdout)
-				format.Log(p, c, stderr)
+				format.LogWithCtx(ctx, p, c, stdout)
```
pavannd1, line 94: Same as above

pkg/function/kube_exec.go
chaitanya-baraskar: @pavannd1 you are right. I will update function to not log these fields for now. Later we can get rid of unused arguments.

pkg/format/format.go
```diff
@@ -55,3 +51,18 @@ func info(podName string, containerName string, l string) {
 		log.Print("Pod Update", field.M{"Pod": podName, "Container": containerName, "Out": l})
 	}
 }
+
+func LogWithCtx(ctx context.Context, podName string, containerName string, output string) {
+	if output != "" {
+		logs := regexp.MustCompile("[\r\n]").Split(output, -1)
```
julio-lopez, line 57: Perform the reg exp compilation only once by initializing it in the package and then reusing it on all calls to `LogWithCtx`

pkg/function/kube_exec.go
chaitanya-baraskar: @pavannd1 I am fixing where this is the case.

pkg/format/format.go
```diff
@@ -24,13 +25,14 @@ import (
 	"github.com/kanisterio/kanister/pkg/log"
 )
 
+var regex *regexp.Regexp
+
+func init() {
+	regex = regexp.MustCompile("[\r\n]")
+}
```
julio-lopez, line 32: nit: does the following work?

```suggestion
var regex = regexp.MustCompile("[\r\n]")
```

chaitanya-baraskar: @julio-lopez let me check

chaitanya-baraskar: @julio-lopez you are right. It's working. I have tested changes end to end and updated gist with new logs.

pavannd1 (commented):
LGTM

julio-lopez (commented):
+1: LG 👍

PR #1278, opened 2022-03-11
Update docs about how CRDs are managed

viveksinghggits:
## Change Overview

We recently [changed the](https://github.com/kanisterio/kanister/pull/1244) way CRDs are managed by kanister. This PR updates the docs to specify the same.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

run `make docs` at the root of repo to build the docs.

image


docs/install.rst
```diff
@@ -45,6 +45,16 @@ This will install the Kanister controller in the ``kanister`` namespace
...
 
 
+Kanister Custom Resource Definitions (CRD)
+==========================================
+
+`Custom Resource Definitions <https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/>`_
+that Kanister needs are, by default, managed by Kanister controller. If Kanister
+controller doesn't have the authorization to create or update CRDs, you can disable
+that using ``controller.updateCRDs`` helm flag. If set to ``false``, helm will
+manage the CRDs instead of Kanister controller.
```
pavannd1, line 55:
```suggestion
`Custom Resource Definitions <https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/>`_
that Kanister needs are, by default, managed by Kanister controller. If the Kanister
controller doesn't have the authorization to create or update CRDs, you can disable
that using the ``controller.updateCRDs`` helm flag. If set to ``false``, helm will
manage the CRDs instead of the Kanister controller.
```

[nit]

pavannd1 (approved):

docs/install.rst
ihcsim: I think we can add some example use cases on why the controller won't have the RBAC permissions to manage CRDs.

```suggestion
Kanister defines a few `Custom Resource Definitions (CRDs) <https://docs.kanister.io/architecture.html#custom-resources>` to express operation blueprints and action tasks. By default, these CRDs are managed by the Kanister controller. 

In deployment workflows where the Kanister's YAML manifests must be version-controlled, one can disable these auto-managed CRDs, by setting the Helm ``controller.updateCRDs`` flag to ``false``. This causes the Kanister controller to not output the CRDs YAML during installation. Users can then use commands like ``helm template --set controller.updateCRDS=false`` to output the static YAML of the CRDs and the Kanister application.

Note that use this flag on existing CRDs during an upgrade is not supported.
```

viveksinghggits: Hi @ihcsim 
`Kanister's YAML manifests` this is a bit confusing, what do we mean by Kanister's YAML do we mean YAML of CRDs that Kanister uses. Also, that was not the main use cases why we decided to make have this option. The main use case was that they were not willing to run kanister with authz to manage CRDs. 
`the Kanister controller to not output the CRDs YAML during installation` this sentence is also a bit confusing, Kanister controller doesn't actually output CRDs, it either creates or updates them directly.

docs/install.rst
```diff
@@ -45,6 +45,19 @@ This will install the Kanister controller in the ``kanister`` namespace
                                        --namespace kanister
 
 
+Kanister Custom Resource Definitions (CRDs)
+===========================================
+
+Kanister defines a few `Custom Resource Definitions (CRDs) <https://docs.kanister.io/architecture.html#custom-resources>`_
+to express backup workflow and their execution. By default, these CRDs are
```
ihcsim, line 52: How about this?

> to express and execute data protection workflows

docs/install.rst
```diff
@@ -45,6 +45,19 @@ This will install the Kanister controller in the ``kanister`` namespace
...
+managed (created/updated) by the Kanister controller.
+
+If the Kanister controller doesn't have the authorization to create or
```
ihcsim, line 55: s/authorization/permissions

PR #1296, opened 2022-03-21
Enable automatic dependabot pull requests

julio-lopez:
## Pull request type

- [x] :hamster: Trivial/Minor
- [x] :robot: Automation


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    open-pull-requests-limit: 10
+    schedule:
+      interval: daily
```
pavannd1, line 6:
```suggestion
      interval: daily

```
[nit] may not be required

pavannd1 (approved):

.github/dependabot.yml
julio-lopez: true, that's the default. Leaving it there for reference in case we want to change it

ihcsim (approved):

Issue #1345, opened 2022-03-31
Change the way the certificate for webhook is generated 


muffl0n:
Currently, the certificate for the ValidatingWebhook is generated on installation.
This has the drawback, that a `helm diff` reports a change on every run. It reports changes (to the certificate) that makes it hard to distinguish what is due to configuration changes made by oneself.

I took a look into several solutions and found the way "ingress-nginx" does it the smartest one.
This is a rough breakdown on how they do it:
* On `helm install` the `ValidatingWebhook` is installed, without cert-data
* Also a `Job` is installed, that runs immediately, creates a certificate+secret and patches the `ValidatingWebhook` to use that secret

They use the image [jet/kube-webhook-certgen](https://github.com/jet/kube-webhook-certgen/), see https://github.com/eclipse/che/issues/17855#issuecomment-705393366 for some more details.

I would love to contribute on this, once it get's accepted as "we should do this". I'm relatively new to webhooks, so this could cost me a bit more of my time. But I'm eager to learn. :)

See https://kanisterio.slack.com/archives/C85C8V22J/p1647870529989079?thread_ts=1647866137.638459&cid=C85C8V22J for reference.

ihcsim:
Will it be acceptable to fall back to using the Kubernetes apiserver trust root? That seems to be the behaviour if the [`caBundle` field](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#webhookclientconfig-v1-admissionregistration-k8s-io) isn't specified in the validating webhook configuration.

muffl0n:
But you still have to put a certificate in the pod, that acts as the ValidatingWebhook, right? 
Currently the certificate is generated in validating-webhook.yaml and put in the secret "kanister-webhook-certs". This is mounted as "webhook-certs" in the operator in deployment.yaml

ihcsim:
I wonder if the webhook can issue a [CSR](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/) to the apiserver and get a signed certificate from it.

ihcsim:
Looks like it uses the same K8s CSR API. Just trying to see if we can avoid introducing external components.

ihcsim:
Thanks for putting https://github.com/kanisterio/kanister/pull/1369 together. It helps to see what the code looks like :+1:. At this point, we want to keep Kanister unopinionated in regards to cert management solutions. Depending on the user's requirements, there will always be legit reasons to pick one solution over the other.

[...]

PR #1369, opened 2022-04-07
Create certificate for webhook with "kube-webhook-certgen"

muffl0n:
## Change Overview

This PR uses the tool "kube-webhook-certgen" to generate the certificate for the webhook. The yaml files are shamelessly stolen from https://github.com/kubernetes/ingress-nginx/tree/main/charts/ingress-nginx/templates/admission-webhooks

I'm sure there are things to optimize, though this works pretty well in my test cluster. I also tidied up the template after copying them over to keep it simple. But maybe it's better to keep them as original as possible for easier updating.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #1345

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


ihcsim:
Is there a way to auto-renew  the certs based on the cert validity period, not the chart upgrade cycle?

Does the tool let us change the default (100y) cert validity period? Some users have stricter requirements. I get having a long-lived CA, but disagree with long-lived leaf certs. It also doesn't seem to let us revoke certificates in case of security incidents.

How would this work if we introduce another TLS server?

muffl0n:
As the job only runs on deployment of the chart, there is no recurring check if the certificate is still valid between those runs. Also you can't change the default validity period of 100y. Documentation is here: https://github.com/kubernetes/ingress-nginx/tree/main/images/kube-webhook-certgen

I strongly believe that this solution is more than sufficient for our use case. If a big project like ingress-nginx goes this way, we can safely "offload" our concerns as others already evaluated them. Also, if somebody has stricter requirements, they need a different solution like cert-manager anyway. Neither the current solution (creating certificate with 1y validity on deployment) nor my proposed solution (creating certificate with 100y validity once) would be sufficient. So we can safely ignore those for this case cause it's out of scope.

If we introduce another TLS server/webhook, we'd need another job. We can use the same roles, serviceaccount, etc so we would greatly benefit from the proposed additions. If we introduced an additional "mutatingwebhook", we'd just need some small changes to the templates.

muffl0n:
Using cert-manager for certification creation would be relatively easy and even does automatic recreation!
issuer.yaml
```
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: selfsigned-issuer
spec:
  selfSigned: {}
```
certificate.yaml
```
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: {{ include "kanister-operator.fullname" . }}-admission
spec:
  commonName: {{ include "kanister-operator.fullname" . }}.{{ .Release.Namespace }}.svc
  dnsNames:
    - {{ include "kanister-operator.fullname" . }}
    - {{ include "kanister-operator.fullname" . }}.{{ .Release.Namespace }}.svc
  issuerRef:
    kind: Issuer
    name: selfsigned-issuer
  secretName: {{ include "kanister-operator.fullname" . }}-admission
```
Add annotation to validating-webhook.yaml
```
....
metadata:
  name: "blueprints.cr.kanister.io"
  annotations:
    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "kanister-operator.fullname" . }}-admission
....
```

Reference: https://www.elastic.co/guide/en/cloud-on-k8s/1.2/k8s-webhook-cert-manager.html

Pitfall: After certificate rotation, there is no way to trigger the operator to pick up the new certificate. See cert-manager/cert-manager#1440. So we would have to add a watcher that restarts the pod once the certificate changes.

It would be worth the effort for "high-security" environments, but is fairly out of scope for this PR.

PR #1390, opened 2022-04-16
Strictly match Group in ResourceTypeMatcher: K8s core group

ewhamilton:
## Change Overview

This PR supports strict matching of Group in ResourceTypeMatcher.

This allows use of K8s core resources (group == "") in filters without
accidentally matching resources with same name in other groups. Maintains
wildcard behavior for "" in version and resource so that matches can be
done for entire groups and all versions of a specific group:resource.

Note: the filter package has been contributed to Kanister in the hope that it
may be useful. It does not appear that the Matches() method that is modified
here is used by anything in Kanister other than test code. This package is
used by K10 and an additional unit test also exercises this change.

## Pull request type

- [x] :bug: Bugfix

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/filter/filter_test.go
+++ b/pkg/filter/filter_test.go
@@ -18,7 +18,7 @@ import (
 	"testing"
 
 	. "gopkg.in/check.v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
@@ -337,12 +337,12 @@ func (s *FilterSuite) TestGroupVersionResourceIncludeExclude(c *C) {
 					Group:   "mygroup",
 					Version: "yourversion",
 				},
+			},
+			exclude: []schema.GroupVersionResource{
 				{
 					Group:   "yourgroup",
 					Version: "myversion",
 				},
-			},
-			exclude: []schema.GroupVersionResource{
 				{
 					Group:   "yourgroup",
 					Version: "yourversion",
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -15,7 +15,7 @@
 package filter
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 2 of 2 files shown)


pkg/filter/filter_test.go
```diff
@@ -18,7 +18,7 @@ import (
 	"testing"
 
 	. "gopkg.in/check.v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
ihcsim, line 21: Not needed?

ihcsim (commented):
I am not familiar enough with the `pkg/discovery/discovery.go` code to tell how this change will affect it: 

```sh
 ✗ grep -rnI "\.Any(" *                                                                                                
pkg/discovery/discovery.go:63:          if !exclude.Any(gvr) {
```

`Any()` eventually calls into `Matches()`.

Otherwise, I agree with this change since [upstream](https://github.com/kubernetes/api/blob/cd271e7523ef7b6b485aa42c0218c294ca00a001/core/v1/register.go#L25-L26) uses the same value.

pkg/filter/filter_test.go
ewhamilton: VSCode likely added it automatically. I don't think it is needed. OTOH, 72 of the 82 imports of `"k8s.io/api/core/v1"` in kanister code use it. So I'd do it for consistency.

ewhamilton:
> I am not familiar enough with the `pkg/discovery/discovery.go` code to tell how this change will affect it: exclude.Any() uses it.

Within the Kanister package, the two uses of ignoreGroupErrors are just above, `AllGVRsIgnoreGroupErrs()` and `NamespacedGVRsIgnoreGroupErrs()`. Neither are currently called from within Kanister.

`NamespacedGVRsIgnoreGroupErrs()` is used by K10 where it passes in
```
		filter.ResourceTypeMatcher{
			filter.ResourceTypeRequirement{Group: "metrics.k8s.io"},
			filter.ResourceTypeRequirement{Group: "external.metrics.k8s.io"},
			filter.ResourceTypeRequirement{Group: "custom.metrics.k8s.io"},
		}
```
Since the Group values here are not `""`, there is no effect. If for some reason we wanted to exclude errors for the core group, we wouldn't want it to match either, so the change would be desirable. If we wanted to ignore errors from all groups, an exclude filter of `""` would do so.

bathina2 (approved):

PR #1395, opened 2022-04-21
Update ubi-minimal base image to ubi-minimal:987

infraq:
Update ubi-minimal base image to ubi-minimal:987


pavannd1:
Closing since this was an experiment
cc: @kale-amruta

PR #1447, opened 2022-05-27
Add infrastructure for user displayable errors

viveksinghggits:
## Change Overview

This PR adds infrastructure to enable users to add errors that should be shown to the end users instead of showing them the entire stack trace. The usage can be figured out by looking at the test.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

```
~/work/opensource/kanister/pkg/errors (user-errors*) » go test 
OK: 3 passed
PASS
ok  	github.com/kanisterio/kanister/pkg/errors	0.001s
```

### DIFF
--- a/pkg/errors/user_errors.go
+++ b/pkg/errors/user_errors.go
@@ -34,7 +34,9 @@ func (ue *UserError) UserMessage() string {
 	return ue.msg
 }
 
-func (ue *UserError) Unwrap() error { return ue.error }
+func (ue *UserError) Unwrap() error {
+	return ue.error
+}
 
 // UserErrorWithMessage initializes the UserError with given error and a message.
 // Needs to accept error if we want to initialize the UserError in between the
@@ -49,7 +51,7 @@ func UserErrorWithMessage(err error, msg string) UserMessenger {
 // UserMessagesInError gets us messages of all the `UserError`s that are wrapped
 // in the passed error's chain
 func UserMessagesInError(err error) []string {
-	userMessages := []string{}
+	var userMessages []string
 	if err == nil {
 		return nil
 	}
--- a/pkg/errors/user_errors.go
+++ b/pkg/errors/user_errors.go
@@ -1,0 +1,63 @@
+// Copyright 2022 The Kanister Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package errors
+
+import (
+	"errors"
+)
+
+type UserMessenger interface {
+	error
+	UserMessage() string
+}
+
+// UserError is an error type that be checked to be present in the
+// error chain. And if present, can be shown to the end users
+type UserError struct {
+	error
+	msg string
+}
+
+func (ue *UserError) UserMessage() string {
+	return ue.msg
+}
+
+func (ue *UserError) Unwrap() error { return ue.error }
+
+// UserErrorWithMessage initializes the UserError with given error and a message.
+// Needs to accept error if we want to initialize the UserError in between the
+// error chain
+func UserErrorWithMessage(err error, msg string) UserMessenger {
+	if err == nil {
+		err = errors.New(msg)
+	}
+	return &UserError{err, msg}
+}
+
+// UserMessagesInError gets us messages of all the `UserError`s that are wrapped
+// in the passed error's chain
+func UserMessagesInError(err error) []string {
+	userMessages := []string{}
+	if err == nil {
+		return nil
+	}
+	for err != nil {
+		if e, ok := err.(*UserError); ok {
+			userMessages = append(userMessages, e.UserMessage())
+		}
+		err = errors.Unwrap(err)
+	}
+	return userMessages
+}
--- a/pkg/errors/user_errors_test.go
+++ b/pkg/errors/user_errors_test.go
@@ -1,0 +1,56 @@
+// Copyright 2022 The Kanister Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package errors
+
+import (
+	"testing"
+
+	"github.com/pkg/errors"
+	. "gopkg.in/check.v1"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type UserErrorSuite struct{}
+
+var _ = Suite(&UserErrorSuite{})
+
+func (u *UserErrorSuite) TestUserMessagesInError(c *C) {
+	expectedUserErrors := []string{"User Error Two", "User Error One"}
+	err1 := errors.New("error 1")
+	err2 := errors.Wrap(err1, "error2")
+
+	ue1 := UserErrorWithMessage(err2, expectedUserErrors[1])
+	err3 := errors.Wrap(ue1, "errro 3")
+
+	ue2 := UserErrorWithMessage(err3, expectedUserErrors[0])
+	err4 := errors.Wrap(ue2, "error 4")
+
+	c.Assert(UserMessagesInError(err4), DeepEquals, expectedUserErrors)
+}
+
+// The case where there are no `UserError`s in error chain
+func (u *UserErrorSuite) TestUserMessagesInErrorWithoutUserError(c *C) {
+	err1 := errors.New("error 1")
+	err2 := errors.Wrap(err1, "error 2")
+	err3 := errors.Wrap(err2, "error 3")
+	err4 := errors.Wrap(err3, "error 4")
+
+	c.Assert(len(UserMessagesInError(err4)), Equals, 0)
+}
+
+func (u *UserErrorSuite) TestUserMessagesInErrorNilError(c *C) {
+	c.Assert(UserMessagesInError(nil), Equals, nil)
+}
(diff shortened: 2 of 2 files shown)


viveksinghggits:
cc @faradayfan

pkg/errors/user_errors.go
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package errors
+
+import (
+	"errors"
+)
+
+type UserMessenger interface {
```
ihcsim, line 21: If the idea is that the caller should use the exported `UserErrorWithMessage()` to fetch the displayable message, is this interface necessary? The Go stdlib exposes many [errors by value](https://pkg.go.dev/io#pkg-variables).

pkg/errors/user_errors.go
```diff
@@ -0,0 +1,63 @@
...
+func UserMessagesInError(err error) []string {
+	userMessages := []string{}
+	if err == nil {
+		return nil
+	}
+	for err != nil {
+		if e, ok := err.(*UserError); ok {
+			userMessages = append(userMessages, e.UserMessage())
+		}
+		err = errors.Unwrap(err)
```
ihcsim, line 60: Is this reliable? What if there is an intermediate error in the chain that didn't implement `Unwrap()`, causing ` err == nil`? Any subsequent `UserError`s will be lost.

pkg/errors/user_errors.go
```diff
@@ -0,0 +1,63 @@
...
+
+type UserMessenger interface {
+	error
+	UserMessage() string
+}
+
+// UserError is an error type that be checked to be present in the
+// error chain. And if present, can be shown to the end users
+type UserError struct {
+	error
```
ihcsim, line 29: Will there be cases where `UserError` needs to be a standalone error, without a parent `error`? Why not let `UserError` implements `errors` directly and the caller decide when/how they want to wrap `UserError` in a chain of errors?

ihcsim (commented):
Are there some other follow-up tasks to this PR? I am just curious why it needs to be implemented in Kanister and not some downstream packages?

viveksinghggits:
Hi Ivan,
Yes, there will be follow up PRs to this PR.

pkg/errors/user_errors.go
denisvmedia: Imagine, you got an error:

```go
err := errors.New("somewhat")
```

And then you need an instance of `UserError` to present it to the user in a user-friendly manner. At the same time you want to preserve the behavior of `errors.Is`/`errors.As` for the error wrapped with `UserError`. How would you do that?

pkg/errors/user_errors.go
denisvmedia: Interesting, how come there can be an error in the chain, which doesn't implement `Unwrap`, but at the same time wraps another error? I can imagine that only as a consequence of a potential bug (a struct that implements `error`, that allows wrapping an error, but someone forgot to implement `Unwrap` for that struct).

pkg/errors/user_errors.go
```diff
@@ -0,0 +1,63 @@
...
+	if err == nil {
+		err = errors.New(msg)
+	}
+	return &UserError{err, msg}
+}
+
+// UserMessagesInError gets us messages of all the `UserError`s that are wrapped
+// in the passed error's chain
+func UserMessagesInError(err error) []string {
+	userMessages := []string{}
```
denisvmedia, line 52: We could probably start with a nil slice, which should be treated equivalent to an empty slice.

```suggestion
	var userMessages []string
```

pkg/errors/user_errors_test.go
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package errors
+
+import (
+	"testing"
+
+	"github.com/pkg/errors"
+	. "gopkg.in/check.v1"
```
denisvmedia, line 21: Maybe avoid dot import?

pkg/errors/user_errors.go
denisvmedia: Yeah, I agree with the points. But if the interface is really required, I'd put a description on top for the future implementations (if any). Also, we should probably add `Unwrap` to the interface to guarantee that the implementing errors implement it as well.

pkg/errors/user_errors.go
```diff
@@ -0,0 +1,63 @@
...
+type UserError struct {
+	error
+	msg string
+}
+
+func (ue *UserError) UserMessage() string {
+	return ue.msg
+}
+
+func (ue *UserError) Unwrap() error { return ue.error }
```
denisvmedia, line 37: These two lines seem to be inconsistent in the style (oneliner vs. block)

pkg/errors/user_errors.go
viveksinghggits: mm, good points. Let me try to writer some test around it to see the behaviour.

pkg/errors/user_errors_test.go
viveksinghggits: hmm, can we use gocheck  without importing the module?

pkg/errors/user_errors.go
ihcsim: I think updating `UserError` to implement `errors` is more idiomatic. There are probably other ways to go about this, but here is an example, where:

i. `UserError` can be treated like an independent `error`
i. The constructor uses `errors.As` to find and append existing `UserError`s to the new `UserError`, so that there is always one `UserError` to worry about during unwrapping

```diff
diff --git a/pkg/errors/user_errors.go b/pkg/errors/user_errors.go
index aa964207..5363b8fa 100644
--- a/pkg/errors/user_errors.go
+++ b/pkg/errors/user_errors.go
@@ -16,48 +16,51 @@ package errors
 
 import (
 	"errors"
+	"fmt"
+	"strings"
 )
 
-type UserMessenger interface {
-	error
-	UserMessage() string
-}
-
 // UserError is an error type that be checked to be present in the
 // error chain. And if present, can be shown to the end users
 type UserError struct {
-	error
 	msg string
 }
 
-func (ue *UserError) UserMessage() string {
+func (ue *UserError) Error() string {
 	return ue.msg
 }
 
-func (ue *UserError) Unwrap() error { return ue.error }
-
 // UserErrorWithMessage initializes the UserError with given error and a message.
 // Needs to accept error if we want to initialize the UserError in between the
 // error chain
-func UserErrorWithMessage(err error, msg string) UserMessenger {
+func WrapUserError(err error, msg string) error {
+	ue := &UserError{msg}
 	if err == nil {
-		err = errors.New(msg)
+		return ue
+	}
+
+	// append any existing UserErrors in err to msg, so there is always one
+	// msg to unwrap.
+	var u *UserError
+	if errors.As(err, &u) {
+		ue.msg += fmt.Sprintf("\n%s", u.Error())
 	}
-	return &UserError{err, msg}
+
+	return fmt.Errorf("%w %s", ue, err)
 }
 
 // UserMessagesInError gets us messages of all the `UserError`s that are wrapped
 // in the passed error's chain
-func UserMessagesInError(err error) []string {
+func UnwrapUserError(err error) []string {
 	userMessages := []string{}
 	if err == nil {
 		return nil
 	}
-	for err != nil {
-		if e, ok := err.(*UserError); ok {
-			userMessages = append(userMessages, e.UserMessage())
-		}
-		err = errors.Unwrap(err)
+
+	var ue *UserError
+	if errors.As(err, &ue) {
+		userMessages = strings.Split(ue.Error(), "\n")
 	}
+
 	return userMessages
 }
diff --git a/pkg/errors/user_errors_test.go b/pkg/errors/user_errors_test.go
index 5ab64162..91ce297e 100644
--- a/pkg/errors/user_errors_test.go
+++ b/pkg/errors/user_errors_test.go
@@ -32,13 +32,13 @@ func (u *UserErrorSuite) TestUserMessagesInError(c *C) {
 	err1 := errors.New("error 1")
 	err2 := errors.Wrap(err1, "error2")
 
-	ue1 := UserErrorWithMessage(err2, expectedUserErrors[1])
+	ue1 := WrapUserError(err2, expectedUserErrors[1])
 	err3 := errors.Wrap(ue1, "errro 3")
 
-	ue2 := UserErrorWithMessage(err3, expectedUserErrors[0])
+	ue2 := WrapUserError(err3, expectedUserErrors[0])
 	err4 := errors.Wrap(ue2, "error 4")
 
-	c.Assert(UserMessagesInError(err4), DeepEquals, expectedUserErrors)
+	c.Assert(UnwrapUserError(err4), DeepEquals, expectedUserErrors)
 }
 
 // The case where there are no `UserError`s in error chain
@@ -48,9 +48,9 @@ func (u *UserErrorSuite) TestUserMessagesInErrorWithoutUserError(c *C) {
 	err3 := errors.Wrap(err2, "error 3")
 	err4 := errors.Wrap(err3, "error 4")
 
-	c.Assert(len(UserMessagesInError(err4)), Equals, 0)
+	c.Assert(len(UnwrapUserError(err4)), Equals, 0)
 }
 
 func (u *UserErrorSuite) TestUserMessagesInErrorNilError(c *C) {
-	c.Assert(UserMessagesInError(nil), IsNil)
+	c.Assert(UnwrapUserError(nil), IsNil)
 }
```

pkg/errors/user_errors.go
denisvmedia: But if the downstream custom error does not implement it, it means they (be it intentional or not) do not provide any means of getting the underlying error. Is there anything we can do about it?

pkg/errors/user_errors.go
ihcsim: This suggestion doesn't change how the caller will call the exported functions, per the unit tests.

pkg/errors/user_errors.go
ihcsim: See my suggested code change for ways to get rid of the `for` loop.

pkg/errors/user_errors.go
denisvmedia: Consider your proposal:

```go
var u *UserError
if errors.As(err, &u) {
    ue.msg += fmt.Sprintf("\n%s", u.Error())
}
```

As there is no loop anymore, what will happen to the potential further underlying user messages?

Also, I'm not sure that gluing/splitting the errors by "\n" is a good choice.

pkg/errors/user_errors_test.go
denisvmedia: I mean "gopkg.in/check.v1" (without a dot in front of it). But if it's a project wide, then it's fine as is.

viveksinghggits: its not a project wide import, just in that test source file. But  I thought we should use dot import just for the sake of consistency. I can change that if you think not having dot import is a better idea.

denisvmedia: > its not a project wide import, just in that test source file.

:) I meant this: if we use dot imports all over the project, then we should keep it here as is. If we don't use dot imports, we better don't use it here. As you say, to be consistent.

viveksinghggits: sorry about that. Yes, since we use dot imports all over the project, I think we should keep dot import here as well.

PR #1449, opened 2022-05-31
Capture Task Pod's Most-Recent Error Logs As Output Artifacts

ihcsim:
## Change Overview

This PR updates `KubeTask` to capture the most-recent logs of a failed task pod as output artifacts in the ActionSet. This ensure that these logs are preserved after the pod is terminated to aid with debugging. The number of log lines to be captured is capped by the tail size parameter, which can be overridden by the new `KANISTER_TASK_LOGS_TAIL_SIZE` environment variable.

Although these logs are currently streamed to the Kanister controller, it may not always be easy to parse the controller's logs for specific task errors.

(The existing `status.error.message` which is normally filled with the **entire** pod spec is untouched. It can use some re-work in a separate PR.)

Test steps:

Deploy this blueprint:

```sh
cat<<EOF |  kubectl apply -f -
kind: Blueprint
metadata:
  name: task-errors
  namespace: kanister
actions:
  awscli:
    phases:
    - func: KubeTask
      name: echo
      args:
        namespace: "{{ .Namespace.Name }}"
        image: ghcr.io/kanisterio/kanister-tools:v9.99.9-dev
        command:
        - sh
        - "-c"
        - |
          aws -h
          exit 1
EOF
```

Deploy this ActionSet:

```sh
cat <<EOF | kubectl create -f -
kind: ActionSet
metadata:
  generateName: task-error-awscli-
  namespace: kanister
spec:
  actions:
  - name: awscli
    blueprint: task-errors
    object:
      kind: Namespace
      name: default
EOF
```

Check the ActionSet for recent logs:

```sh
$ kubectl -n kanister get actionset task-error-awscli-rtknm -ojsonpath='{.status.actions[?(@.name=="awscli")].phases}' | jq .
[
  {
    "name": "echo",
    "output": {
      "recentLogs": "sh: aws: command not found\n"
    },
    "state": "failed"
  }
]
```

Fixes https://github.com/kanisterio/kanister/discussions/1226.

## Pull request type

- Improvement

## Issues

- https://github.com/kanisterio/kanister/discussions/1226

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -460,6 +460,9 @@ func (c *Controller) runAction(ctx context.Context, as *crv1alpha1.ActionSet, aI
 						Message: err.Error(),
 					}
 					ras.Status.Actions[aIDX].Phases[i].State = crv1alpha1.StateFailed
+					if output != nil {
+						ras.Status.Actions[aIDX].Phases[i].Output = output
+					}
 					return nil
 				}
 			} else {
--- a/pkg/function/kube_task.go
+++ b/pkg/function/kube_task.go
@@ -15,7 +15,10 @@
 package function
 
 import (
+	"bytes"
 	"context"
+	"os"
+	"strconv"
 
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
@@ -31,7 +34,10 @@ import (
 )
 
 const (
-	jobPrefix = "kanister-job-"
+	jobPrefix               = "kanister-job-"
+	kubeTaskLogsOutputKey   = "recentLogs"
+	kubeTaskLogsTailSizeEnv = "KANISTER_TASK_LOGS_TAIL_SIZE"
+
 	// KubeTaskFuncName gives the function name
 	KubeTaskFuncName       = "KubeTask"
 	KubeTaskNamespaceArg   = "namespace"
@@ -83,10 +89,53 @@ func kubeTaskPodFunc(cli kubernetes.Interface) func(ctx context.Context, pod *v1
 		}
 		// Wait for pod completion
 		if err := kube.WaitForPodCompletion(ctx, cli, pod.Namespace, pod.Name); err != nil {
-			return nil, errors.Wrapf(err, "Failed while waiting for Pod %s to complete", pod.Name)
+			logs, logerr := recentLogsAsOutput(ctx, cli, pod.Namespace, pod.Name, kubeTaskLogsOutputKey)
+			if logerr != nil {
+				return nil, logerr
+			}
+
+			return map[string]interface{}{
+				logs.Key: logs.Value,
+			}, errors.Wrapf(err, "Failed while waiting for Pod %s to complete", pod.Name)
 		}
 		return out, err
 	}
+}
+
+// recentLogsAsOutput creates an artifact output containing the recent logs of
+// a pod. The output maps 'key' to the log string.
+// The number of log lines to retrieve is determined by either the value of the
+// TaskLogsTailSizeEnv environment variable, or the default
+// kube.TaskLogsTailSizeDefault.
+func recentLogsAsOutput(ctx context.Context, cli kubernetes.Interface, namespace, pod, key string) (*output.Output, error) {
+	var tailSize int64
+	tailSizeEnv, exists := os.LookupEnv(kubeTaskLogsTailSizeEnv)
+	if exists {
+		parsed, err := strconv.ParseInt(tailSizeEnv, 10, 0)
+		if err != nil {
+			return nil, err
+		}
+		tailSize = parsed
+	}
+
+	if tailSize == 0 {
+		tailSize = kube.TaskLogsTailSizeDefault
+	}
+
+	logs, err := kube.GetRecentPodLogs(ctx, cli, namespace, pod, tailSize)
+	if err != nil {
+		return nil, err
+	}
+
+	buf := &bytes.Buffer{}
+	if _, err := buf.ReadFrom(logs); err != nil {
+		return nil, err
+	}
+
+	return &output.Output{
+		Key:   key,
+		Value: buf.String(),
+	}, nil
 }
 
 func (ktf *kubeTaskFunc) Exec(ctx context.Context, tp param.TemplateParams, args map[string]interface{}) (map[string]interface{}, error) {
--- a/pkg/function/kube_task_test.go
+++ b/pkg/function/kube_task_test.go
@@ -61,6 +61,24 @@ func (s *KubeTaskSuite) TearDownSuite(c *C) {
 	}
 }
 
+func errPhase(namespace string) crv1alpha1.BlueprintPhase {
+	return crv1alpha1.BlueprintPhase{
+		Name: "testOutput",
+		Func: KubeTaskFuncName,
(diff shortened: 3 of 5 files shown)


pkg/function/kube_task.go
```diff
@@ -83,12 +89,55 @@ func kubeTaskPodFunc(cli kubernetes.Interface) func(ctx context.Context, pod *v1
...
+	if exists {
+		parsed, err := strconv.ParseInt(tailSizeEnv, 10, 0)
+		if err != nil {
+			return nil, err
+		}
+		tailSize = parsed
+	}
+
+	if tailSize == 0 {
+		tailSize = kube.TaskLogsTailSizeDefault
```
viveksinghggits, line 122: I was thinking are there chances that the error that we are looking for is not actually available in last 15 lines? Do we rely to setting env var in that case?
And if we want to set the env var, I think figuring out how many lines do we want to look for is a bit challenging. How would we decide this number?

ihcsim: We won't know ahead of time. Hence, I am thinking of making it configurable. We definitely don't want to blindly pull down all the logs, because that will get saved into etcd.

viveksinghggits: right. But I was thinking even if its configurable how can someone decide the number of lines we want to tail from the pod. That's a bit challenging in my opinion.

pkg/controller/controller.go
```diff
@@ -460,6 +460,9 @@ func (c *Controller) runAction(ctx context.Context, as *crv1alpha1.ActionSet, aI
 						Message: err.Error(),
 					}
 					ras.Status.Actions[aIDX].Phases[i].State = crv1alpha1.StateFailed
+					if output != nil {
+						ras.Status.Actions[aIDX].Phases[i].Output = output
+					}
```
viveksinghggits, line 465: So if there are n number of phases and in the actionset status and I want to check the actual error log will have I have to get `status.phases[len(phases)-1].output.recentLogs`?

ihcsim: `recentLogs` will only be added for the failed phase. Since the actionset would not continue the moment a phase failed, there should only be one `recentLogs` in the actionset's status..

ihcsim:
@pavannd1 I ran a few more tests with multiline log outputs. The marshalled outputs with multiline strings aren't very consistent. Sometimes, it's able to correctly handle `\n`, but other times, the line breaks are escaped. AFAICT, I don't think our code is modifying the outputs, right?

Tested with mysql:
```yaml
    phases:                                                                                                                                                                                    
    - name: echo-help                                                                                                                                                                          
      output:                                                                                                                                                                                  
        recentLogs: "mysql  Ver 8.0.29 for Linux on x86_64 (MySQL Community Server                                                                                                             
          - GPL)\nCopyright (c) 2000, 2022, Oracle and/or its affiliates.\n\nOracle                                                                                                            
          is a registered trademark of Oracle Corporation and/or its\naffiliates.                                                                                                              
          Other names may be trademarks of their respective\nowners.\n\nUsage: mysql                                                                                                           
          [OPTIONS] [database]\n  -?, --help          Display this help and exit.\n                                                                                                            
          \ -I, --help          Synonym for -?\n  --auto-rehash       Enable automatic                                                                                                         
          rehashing. One doesn't need to use\n                      'rehash' to get                                                                                                            
          table and field completion, but startup\n                      and reconnecting                                                                                                      
          may take a longer time. Disable with\n                      --disable-auto-rehash.\n                                                                                                 
          \                     (Defaults to on; use --skip-auto-rehash to disable.)\n                                                                                                         
          \ -A, --no-auto-rehash \n                      No automatic rehashing. One                                                                                                           
          has to use 'rehash' to get\n                      table and field completion.                                                                                                        
          This gives a quicker start of\n                      mysql and disables                                                                                                              
          rehashing on reconnect.\n  --auto-vertical-output \n                      Automatically                                                                                              
          switch to vertical output mode if the\n                      result is wider                                                                                                         
          than the terminal width.\n  -B, --batch         Don't use history file.               
```

```yaml
    phases:                                                                                                                                                                                    
    - name: echo-help                                                                                                                                                                          
      output:                                                                                                                                                                                  
        recentLogs: |                                                                                                                                                                          
          Enter password: ERROR 2003 (HY000): Can't connect to MySQL server on '127.0.0.1:3306' (111)
```

Test with pgsql:
```yaml
    phases:                                                                                                                                                                                    
    - name: echo-help                                                                                                                                                                          
      output:                                                                                                                                                                                  
        recentLogs: |                                                                                                                                                                          
          sh: 1: pgsql: not found
      state: failed
```

```yaml
  phases:                                                                                                                                                                                    
    - name: echo-help                                                                                                                                                                          
      output:                                                                                                                                                                                  
        recentLogs: |                                                                                                                                                                          
          psql is the PostgreSQL interactive terminal.                                                                                                                                         
                                                                                                                                                                                               
          Usage:                                                                                                                                                                               
            psql [OPTION]... [DBNAME [USERNAME]]                                                                                                                                               
                                                                                                                                                                                               
          General options:                                                                                                                                                                     
            -c, --command=COMMAND    run only single command (SQL or internal) and exit                                                                                                        
            -d, --dbname=DBNAME      database name to connect to (default: "postgres")                                                                                                         
            -f, --file=FILENAME      execute commands from file, then exit                                                                                                                     
            -l, --list               list available databases, then exit                                                                                                                       
            -v, --set=, --variable=NAME=VALUE                                                                                                                                                  
                                     set psql variable NAME to VALUE                                                                                                                           
                                     (e.g., -v ON_ERROR_STOP=1)                                                                                                                                
            -V, --version            output version information, then exit                                                                                                                     
            -X, --no-psqlrc          do not read startup file (~/.psqlrc)                                                                                                                      
            -1 ("one"), --single-transaction                                                                                                                                                   
                                     execute as a single transaction (if non-interactive)                                                                                                      
            -?, --help[=options]     show this help, then exit                                                                                                                                 
                --help=commands      list backslash commands, then exit                                                                                                                        
                --help=variables     list special variables, then exit    
```


```yaml
    phases:                                                                                                                                                                                    
    - name: echo-help                                                                                                                                                                          
      output:                                                                                                                                                                                  
        recentLogs: "psql: error: connection to server on socket \"/tmp/.s.PGSQL.5432\"                                                                                                        
          failed: No such file or directory\n\tIs the server running locally and accepting                                                                                                     
          connections on that socket?\n"
```

ihcsim:
@pavannd1 @viveksinghggits After putting more thought into this, I don't think this is a good implementation. Although it's convenient, the `status` subresource isn't a good interface to retrieve this kind of datapath logs because:

* there is no easy way to determine the correct log size
* logs are now duplicated in two places
* potentially saving sensitive data in the subresource

Feels like we need to:

* separate Kanister's system logs from datapath logs
* provide a sensible ways to drain and consume the datapath logs

viveksinghggits:
Hi @ihcsim,
Yes, I agree with the things that you mentioned there. And I think figuring out the data path logs from kanister system's logs is going to be a bit challenging.
I am assuming by data path log you mean the logs that are generated by commands that are there in phase's command.

ihcsim:
Superceded by #1497.

PR #1457, opened 2022-06-07
Enable grpc for connecting to Kopia API Server

ankitjain235:
## Change Overview

This PR stops setting `DisableGRPC` to `true`, thus allowing Kopa API client to connect to server using GRPC.
Note: The kopia API server also needs to be started with GRPC enabled.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kopia/connect.go
+++ b/pkg/kopia/connect.go
@@ -55,8 +55,6 @@ func ConnectToAPIServer(
 	serverInfo := &repo.APIServerInfo{
 		BaseURL:                             serverAddress,
 		TrustedServerCertificateFingerprint: fingerprint,
-		// TODO(@pavan): Remove once GRPC support is added (kopia 0.8 release)
-		DisableGRPC: true,
 	}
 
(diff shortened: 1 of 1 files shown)


ihcsim:
Is this merge-able? If not, please label it with `hold off merging`. IIRC, we had to revert this due to regression, right?

pavannd1:
Looks like the downstream tests are passing for this change and the issue seen before is not reproducible. But unfortunately, we won't be able to completely say with confidence that it'll work until we release a new version of Kanister and let the tests run for a week or two without disruption.

pavannd1:
We can hold off on this until we have a way to create Kopia server in Kanister 👍🏼

pavannd1:
We can close this and follow-up later 👍🏼

Issue #1455, opened 2022-06-06
Improve Progress Visibility of Long-Running Tasks


ihcsim:
**Is your feature request related to a problem? Please describe.**

When an actionset triggers a long-running task like the `CopyVolumeData` function, the only way to determine any sort of progress is to gain direct access to the logs and events of the data mover pod. This may not always be feasible as the user may not have the appropriate RBAC permissions to access these subresource endpoints in the pod's namespace. If the export operation takes a long time, it's also possible that the pod might be terminated prematurely without leaving behind any traces of useful logs.

Being able to persist progress data will be very helpful for both live reporting of task progress, as well as future retrospection (e.g., the latency of every phase of an action).

**Describe the solution you'd like**

Update the actionset's status subresource to hold overall blueprint and per-phase progress, to enable Kubernetes clients such as `kubectl`, `client-go` etc. to retrieve this information for progress reporting or debugging purposes.

Since progress tracking may not be meaningful for all operations, we should start by identifying a list of long-running operations. Then we will need to devise some mechanism on how to map operation events to numeric progress status.

The progress tracking interface may need to be able to accept different tracking handlers to account for the different mechanisms on how a phase can be executed; e.g. measure progress by number of bytes uploaded vs. remaining, by estimated time to completion, by log events etc.

It's important to note that this progress computation should not compromise the main data protection task's latency nor lead to resource contention. Progress computation will be performed on a best-effort basis, where it may be de-prioritized, skipped or the accuracy of its result not guaranteed, in the event of a resource contention.

**Describe alternatives you've considered**

Let the caller detect and compute the progress status based on task logs returned by Kanister; i.e. no changes on Kanister side.

ihcsim:
Resolved by #1487, #1586.

PR #1461, opened 2022-06-07
RFC: Add design for blueprint and phase progress calculation

ihcsim:
## Change Overview

This PR adds the design proposal for the blueprint and phase progress calculation. Welcome all comments and feedback.

## Pull request type

- RFC

## Issues

- #1455

### DIFF
--- a/design/progress-tracking.md
+++ b/design/progress-tracking.md
@@ -41,7 +41,7 @@ For example,
 ```yaml
 status:
   progress:
-    percentCompleted: 50 # 1 out of 2 phases are completed
+    percentCompleted: 50.00 # 1 out of 2 actions are completed
     lastTransitionTime: 2022-04-06 14:23:34
   actions:
   - blueprint: my-blueprint
@@ -50,19 +50,19 @@ status:
     - name: echo
       state: completed
       progress:
-        percentCompleted: 100
+        percentCompleted: 100.00
         lastTransitionTime: 2022-04-06 14:13:00
     - name: echo
       state: completed
       progress:
-        percentCompleted: 100
+        percentCompleted: 100.00
         lastTransitionTime: 2022-04-06 14:23:34
     name: action-01
     phases:
     - name: echo
       state: pending
       progress:
-        percentCompleted: 30
+        percentCompleted: 30.00
         lastTransitionTime: 2022-04-06 14:30:31
   state: pending
 ```
@@ -85,10 +85,10 @@ which normally are used to invoke long-running operations:
 #### Blueprint Progress
 
 Initially, the blueprint overall progress is computed by checking the number of
-completed phases against the total number of phases:
+completed actions against the total number of actions:
 
 ```
-blueprint_progress_percent = num_completed_phases / total_num_phases * 100
+blueprint_progress_percent = num_completed_actions / total_num_actions * 100
 ```
 
 In subsequent implementation, the computation alogrithm can be updated to assign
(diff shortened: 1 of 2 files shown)


design/progress-tracking.md
```diff
@@ -0,0 +1,244 @@
...
+* `ExportRDSSnapshotToLocation`
+* `RestoreRDSSnapshot`
+
+#### Blueprint Progress
+
+Initially, the blueprint overall progress is computed by checking the number of
+completed phases against the total number of phases:
+
+```
+blueprint_progress_percent = num_completed_phases / total_num_phases * 100
```
viveksinghggits, line 91: should this be `action_progress_percent`?

design/progress-tracking.md
```diff
@@ -0,0 +1,244 @@
...
+* parsing the log outputs for milestones indicator
+
+✍️ Currently, Kanister Functions do not utilize Kopia to perform their
+underlying work. Once the work to integrate Kopia into Kanister is completed,
+we can extract the progress status directly from the log outputs.
+
+Here's a sample log output of the Kopia create snapshot function:
+
+```sh
+$ kopia snapshot create kanister
```
viveksinghggits, line 127: In both the examples (kopia and restic) we are just considering the utilities that are going to upload the data to object store but in some cases there are chances that there are some other commands that are taking some time.
- One example is the utility that we are using to take backup for example `mysqldump`, `esdump` etc. Should we consider these as well or just measuring upload time would be sufficient.

dsu-igeek: Definitely if we can get progress from the tools we should expose that.

dsu-igeek (approved):
I like this.  Thanks for putting this together!

design/progress-tracking.md
ihcsim: @viveksinghggits yes, good catch! To get the blueprint overall progress, we should be assessing the completed actions, not phases.

design/progress-tracking.md
ihcsim: @viveksinghggits Yes, the goal is that downstream users can add more custom progress trackers (at the code level) per their requirements, if the built-in ones don't work. We can consider adding a built-in generic tracker to output to 25%, 50%, 75% and 100% markers based on certain events in the logs.

design/progress-tracking.md
```diff
@@ -0,0 +1,245 @@
...
+### Assumptions And Constraints
+
+* The progress computation should not compromise the main data protection task's
+latency nor lead to resource contention. Progress computation will be performed
+on a best-effort basis, where it may be de-prioritized with no guarantee on the
+accuracy of its result, or skipped entirely in the event of resource contention.
+
+### Changes To API
+
+The blueprint's overall progress will be reported under the new
```
pavannd1, line 35: Similar to Vivek's suggestion, should this say action/actionset's overall progress too?

design/progress-tracking.md
```diff
@@ -0,0 +1,245 @@
...
+* `ExportRDSSnapshotToLocation`
+* `RestoreRDSSnapshot`
+
+#### Blueprint Progress
+
+Initially, the blueprint overall progress is computed by checking the number of
+completed actions against the total number of actions:
+
+```
+blueprint_progress_percent = num_completed_actions / total_num_actions * 100
```
pavannd1, line 91: s/blueprint/actionset here and below?

design/progress-tracking.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Error Handling
+
+If a phase failed, the progress tracking will cease immediately. The last
+reported progress will be retained in the actionset's `status` subresource.
+
+## Test Cases
+
+New unit tests to be added to the new `progress` package to cover blueprint
+progress, and phase progress calculated with a dummy tracker:
```
pavannd1, line 238:
```suggestion
progress, and phase progress calculated with a sample tracker:
```
😯 https://twitter.com/TwitterEng/status/1278733305190342656/photo/1

design/progress-tracking.md
```diff
@@ -0,0 +1,244 @@
...
+    percentCompleted: 50 # 1 out of 2 phases are completed
+    lastTransitionTime: 2022-04-06 14:23:34
+  actions:
+  - blueprint: my-blueprint
+    name: action-00
+    phases:
+    - name: echo
+      state: completed
+      progress:
+        percentCompleted: 100
```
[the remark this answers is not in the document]
ihcsim: Turns out that we may have to use integers after all, due to the possibility of precision loss on round-trip conversion. See https://github.com/kubernetes-sigs/controller-tools/issues/245.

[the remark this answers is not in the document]
ihcsim: If the fractional part is important, we will need to convert the whole float into an integer before persisting them. But the caller will need to convert it back to float.

pavannd1 (approved):

PR #1465, opened 2022-06-08
Expose some Azure utilities functions

leuyentran:
## Change Overview

Expose some Azure blockstorage utilities functions for derivative works

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

### DIFF
--- a/pkg/blockstorage/azure/client.go
+++ b/pkg/blockstorage/azure/client.go
@@ -67,12 +67,12 @@ func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
 		cloudEnvironment = azure.PublicCloud.Name
 	}
 
-	env, err := azure.EnvironmentFromName(cloudEnvironment)
+	env, err := GetEnvironment(cloudEnvironment)
 	if err != nil {
-		return nil, errors.Wrap(err, "Failed to fetch the cloud environment.")
+		return nil, err
 	}
 
-	authorizer, err := getAuthorizer(env, config)
+	authorizer, err := getAuthorizer(*env, config)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/blockstorage/azure/client.go
```diff
@@ -98,9 +93,24 @@ func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
 	}, nil
 }
 
+func GetEnvironment(config map[string]string) (*azure.Environment, string, error) {
```
ihcsim, line 96: It doesn't look like the 2nd returned string is being used.

pkg/blockstorage/azure/client.go
```diff
@@ -63,16 +62,12 @@ func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
 		}
 	}
 
-	if cloudEnvironment, ok = config[blockstorage.AzureCloudEnviornmentID]; !ok || cloudEnvironment == "" {
-		cloudEnvironment = azure.PublicCloud.Name
-	}
-
-	env, err := azure.EnvironmentFromName(cloudEnvironment)
+	env, _, err := GetEnvironment(config)
```
ihcsim, line 65: Not sure about this one because now the reader of this code has to know about reference vs. value types. And unless we update subsequent functions like `getAuthorizer()`, `getCredConfig()` etc. to take `env` to a reference, I am not sure we are gaining much here. 

What's the issue with the previous code?

pkg/blockstorage/azure/client.go
leuyentran: hi @ihcsim , it was going to be used in a downstream caller. However, after some more thought into the design, this approach is more complicated than necessary. I am switching to https://github.com/kanisterio/kanister/pull/1466 instead

pkg/blockstorage/azure/client.go
leuyentran: The goal was to bring some information that is decided at this level back to downstream caller. Specifically, information such as `AzureCloudEnviornmentID` or `AzureResurceMgrEndpoint`. 
However, I now think this approach is unnecessarily complicated to achieve that goal. So I have switched to https://github.com/kanisterio/kanister/pull/1466

leuyentran:
A simpler approach designed at https://github.com/kanisterio/kanister/pull/1466 . This PR will be closed

PR #1482, opened 2022-06-15
Design for integrating kopia with kanister

shlokchaudhari9:
## Change Overview

This PR consist of a design document that discusses ways to integrate Kopia with Kanister

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- #XXX

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/design/kanister-kopia-integration.md
+++ b/design/kanister-kopia-integration.md
@@ -1,0 +1,56 @@
+# Integrating Kopia with Kanister
+
+This document proposes all the changes required within Kanister to encourage the use of Kopia backup/restore tool.
+
+## Introducing Kopia
+
+Kopia is a simple, cross-platform tool for managing encrypted backups in the cloud.
+It provides fast, incremental backups, secure, client-side end-to-end encryption, compression and data deduplication.
+It is one of the fastest backup/restore tool in the field and with a pluggable backend engine, it allows user to decide which storage the backups should go to.
+It is a lock-free system that, unlike other similar tools, allows for concurrent multi-client operations including garbage collection.
+
+To learn more about Kopia and it's features, visit kopia.io
+
+## Goal
+
+Kopia can focus on its core data transformation as part of various Kanister functions.
+While, Kanister takes care of executing these functions from it's application-centric data protection workflows called Blueprints.
+
+##  Proposed Work
+
+### Kanister Functions 
+
+We already have a rich repository of Kanister functions present on path `pkg/function` that enable application-level data protection in various use-cases.
+Few of these functions need to be re-worked to use Kopia. For starters, these functions are;
+
+1. BackupData
+2. BackupDataAll
+3. BackupDataStats
+4. CopyVolumeData
+5. DeleteData
+6. DeleteDataAll
+7. RestoreData
+8. RestoreDataAll
+
+Please note that the functions stated above will only be **refactored** to use Kopia underneath.
+There should not be any changes with respect to the objective of these functions.
+The motive, arguments and usage for each function stays intact as defined on Kanister docs https://docs.kanister.io/functions.html#existing-functions
+
+### Kanister Functions that use Kopia API Server 
+
+We should also create Kanister functions same as above, but leverage the use of [Kopia in it's server mode](https://kopia.io/docs/repository-server/).
+These Kanister functions would act like Kopia clients trying to securely proxy access to the repository storage without exposing sensitive storage credentials.
+
+Following are the names of a few that could be implemented;
+
+1. BackupDataToServer - To perform backup via the KopiaAPIserver
+2. RestoreDataFromServer - To restore from the KopiaAPIServer
+
+### Limitations to using Kopia API Server
+
+Before starting Kopia API Server, it demands a list of usernames and passwords that are allowed access.
+We can make use of the application namespace ID to generate usernames and a random alphanumeric encryption key to create passwords.
+These would be stored in a secret for later use. But;
+- How should we store the server username and server password securely?
+- At which point in Kanister, should Kopia API Server be started?
+- How could this Kopia API Server be disabled in case downstream Kanister users bring their own Kopia API Server?


design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,56 @@
...
+These Kanister functions would act like Kopia clients that securely proxy access repository storage without exposing sensitive storage credentials.
+
+Following is a list of few such functions;
+
+1. BackupDataToServer - To perform backup via the KopiaAPIserver
+2. RestoreDataFromServer - To restore from the KopiaAPIServer
+
+### Limitations to using Kopia API Server
+
+Before starting Kopia API Server, it demands a list of usernames and passwords that are allowed access.
```
julio-lopez, line 51:
```suggestion
Before starting Kopia API Server, it requires the creation of users that are allowed access.
```

julio-lopez, line 51:
```suggestion
Before starting Kopia API Server, it requires the creation of user that are allowed access with their corresponding passwords.
```
```suggestion
Before starting Kopia API Server, it demands a list of usernames and passwords that are allowed access.
```

julio-lopez (commented):
Looking good overall. Minor nits.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,56 @@
...
+We can make use of the application namespace ID to generate usernames and a random alphanumeric encryption key to create passwords.
+These would be stored in a secret for later use. But;
+- How should we store the server username and server password securely?
+- At which point in Kanister, should Kopia API Server be started?
```
ihcsim, line 55: Instead of starting this on-demand, can it be deployed as a separate, required workload? Feels like a lightweight server.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. BackupDataToServer - To perform backup via the KopiaAPIserver
+2. RestoreDataFromServer - To restore from the KopiaAPIServer
+
+### Limitations to using Kopia API Server
+
+Before starting Kopia API Server, it demands a list of usernames and passwords that are allowed access.
+We can make use of the application namespace ID to generate usernames and a random alphanumeric encryption key to create passwords.
+These would be stored in a secret for later use. But;
+- How should we store the server username and server password securely?
```
ihcsim, line 54: Do these secrets have to be generated internally by the server? Or can it be:

* generated by a human user and store that in a `Secret` resource, as part of the installation process (default use case)
* generated and managed by some KMS solutions + [CSI secret store](https://secrets-store-csi-driver.sigs.k8s.io/) (advance, prod-ready use case)

The idea is to use the K8s `Secret` API as the contract between the user, the server and any downstream clients that may need to share these credentials. The responsibility of securing the at-rest storage falls on the user or the KMS solution.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,56 @@
...
+- At which point in Kanister, should Kopia API Server be started?
+- How could this Kopia API Server be disabled in case downstream Kanister consumers bring their own Kopia API Server?
```
ihcsim, line 56: If we can deploy the server as a separate workload, we should be able to add a Helm option to disable the deployment of this workload, and another option to configure the server endpoint that Kanister controller calls. Will that work?

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,56 @@
...
+## Goal
+
+Kopia can focus on its core data transformation as part of various Kanister functions.
+While, Kanister takes care of executing these functions from it's application-centric data protection workflows called Blueprints.
+
+##  Proposed Work
+
+### Kanister Functions 
+
+We already have a rich repository of Kanister functions present on path `pkg/function` that enable application-level data protection in various use-cases.
```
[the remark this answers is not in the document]
julio-lopez:
```suggestion
We already have a rich repository of Kanister functions present on path `pkg/function` that enable application-level data protection in various use cases.
```

design/kanister-kopia-integration.md
julio-lopez: In the first iteration, it may make sense to have the server running as a requirement. Then functionality to launch and manage the server can be added to Kanister.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,56 @@
...
+Kopia is a simple, cross-platform tool for managing encrypted backups in the cloud.
+It provides fast, incremental backups, secure, client-side end-to-end encryption, compression and data deduplication.
+It is one of the fastest backup/restore tool in the field and with a pluggable backend engine, it allows user to decide which storage the backups should go to.
+It is a lock-free system that, unlike other similar tools, allows for concurrent multi-client operations including garbage collection.
+
+To learn more about Kopia and it's features, visit https://kopia.io/
+
+## Goal
+
+Kopia can focus on its core data transformation as part of various Kanister functions.
```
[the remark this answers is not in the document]
shlokchaudhari9: Yes. Also, data export.

design/kanister-kopia-integration.md
shlokchaudhari9: I really like the idea of having a secret store to manage these secrets (2nd use case)

design/kanister-kopia-integration.md
shlokchaudhari9: That should work assuming Kopia API server has the ability to connect and disconnect underlying storage whenever required. cc: @pavannd1

pavannd1: Maybe Kopia documentation has this information but we need to experiment and see if we can connect/disconnect to/from the repository while the server is running. If yes, we can start the server pod/container based on Kanister installation config.

ihcsim: @shlokchaudhari9 sounds like we have some stress test for Kopia that can help determine throughput and latency. This will give us some controlled benchmark info on scalability i.e. how many backup requests can one Kopia pod support, how many backup requests should trigger HPA etc. Good to include this info in the design doc. @pavannd1 knows where to find these tests.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,54 @@
...
+1. BackupData
+2. BackupDataAll
+3. BackupDataStats
+4. CopyVolumeData
+5. DeleteData
+6. DeleteDataAll
+7. RestoreData
+8. RestoreDataAll
+
+Please note that the functions stated above will only be **refactored** to use Kopia underneath.
```
dsu-igeek, line 33: Will we be able to use something besides Kopia or will this make Kopia the only mechanism supporting these functions?

design/kanister-kopia-integration.md
dsu-igeek: Using a server vs local should be configuration driven, not hard-coded into the functions

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Kanister Functions that use Kopia API Server 
+
+We should also create Kanister functions that leverage the use of [Kopia in it's server mode](https://kopia.io/docs/repository-server/).
+These Kanister functions would act like Kopia clients that securely proxy access repository storage without exposing sensitive storage credentials.
+
+Following is a list of few such functions;
+
+1. BackupDataToServer - To perform backup via the KopiaAPIserver
+2. RestoreDataFromServer - To restore from the KopiaAPIServer
```
dsu-igeek, line 45: Using a server versus the local integration should be configuration driven, not be different functions.

design/kanister-kopia-integration.md
shlokchaudhari9: @ihcsim Okay. I'll go through these tests and add details on doc for server scalability.

design/kanister-kopia-integration.md
shlokchaudhari9: We can continue to maintain separate versions of these Kanister Functions. For ex; `v.0.0.0` of BackupData (or any other Function) will continue to follow restic workflow and a version like `v1.0.0-alpha` could be implemented using Kopia.

The reason I mentioned "refactor" is because in a previous discussion with @pavannd1, we thought one of the end goal here could be to completely replace the Restic based Kanister Functions with the one's that use Kopia. But if that's still unclear, we can decide on it at a later point since function versions are easily configurable by downstream users.

design/kanister-kopia-integration.md
shlokchaudhari9: This makes a lot of sense.... Noting it on the design doc 👍🏼

ihcsim: Why do we need both? Why not just use the client/server setup only?

pavannd1: I think the idea is to always use the client/server setup so we can eliminate the direct-repo functions 👍🏼

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,54 @@
+# Integrating Kopia with Kanister
+
+This document proposes all the changes required within Kanister to better support the use of Kopia backup/restore tool.
```
ihcsim, line 3:
```suggestion
This document proposes all the changes required within Kanister to use [Kopia](https://kopia.io/) as the primary backup and restore tool.
```

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Introducing Kopia
+
+[Kopia](https://kopia.io/) is a powerful, cross-platform tool for managing encrypted backups in the cloud.
```
ihcsim, line 7:
```suggestion
Kopia is a powerful, cross-platform tool for managing encrypted backups in the cloud.
```

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,54 @@
...
+It provides fast and secure backups, using compression, data deduplication and client-side end-to-end encryption.
+It supports a variety of backup storage targets, including object stores, which allows users to choose the storage provider that better addresses their needs.
+It is a lock-free system that allows for concurrent multi-client operations including garbage collection.
```
ihcsim, line 10:
```suggestion
It is a lock-free system that allows for concurrent multi-client operations including garbage collection.

To explore other features of Kopia, see its [documentation](https://kopia.io/docs/features/).
```

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Goal
+
+Kopia can focus on its core data export and transformation as part of various Kanister functions.
+While, Kanister continues to execute these functions from it's application-centric data protection workflows called Blueprints.
```
ihcsim, line 15: I am not sure what this means and how this benefits Kanister users and downstream applications. I think this section should describe what problems we are trying to solve, and what is lacking in the current approach.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,54 @@
...
+
+##  Proposed Work
```
ihcsim, line 17: s/Proposed Work/Scope

ihcsim (commented):
This is getting close. Let's work on getting it merged.

Here are some info that I think should be included:

Problem Statement - what problems are we trying to solve? why Kopia? what's wrong with the existing approach? slower, stale dependencies, unencrypted in-flight data, not scalable etc.

UX - how would this affect existing users? what would their upgrade path looks like? E.g., if they just upgrade Kanister and make no changes to their blueprints, will things still work as-is? How would they opt-in to the new Kopia-based functions? Do we have example blueprint, actionset? Do we need to feature flag this? 

Versioning - how are we going to manage the different versions of the Kanister Functions to ensure backward compatibility? what version scheme are we using?

Server setup and failure domain - what steps would a user take to set up and tear down their Kopia server? Will the logs of this server be persisted for debugging purposes? What happens if the server failed while a backup is in progress? Will we retry, fallback to current approach etc.?

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,98 @@
...
+## User Experience
+
+- Existing users and downstream consumers can continue to make use of previous functionality as per the above 'Backward Compatibility' section
+- Which means upgrading the Kanister controller version wouldn't need any changes in existing blueprint workflows. 
+- However, the user experience for new Kanister Functions is expected to change based on the design decisions for Kopia Repository Server workload usage
+- Mainly, the user might have to perform CRUD on the Repository Server workload which would be an added prerequisite to work with Kopia-based Kanister Functions
+- In case, the users switch off the feature flag for server workload and wish to create a server by themselves, 
+  we aim to provide them with detailed steps to create a server workload as part of "bring-your-own-server" BYOS model
+
+## Limitations to using Kopia API Server
```
shlokchaudhari9, line 91: @pavannd1 Should we get rid of the "Limitations" section? Because these are quite vague right now. We will have clearer idea of limitations while working on the Kopia Repo Server CR design. And would be able to document it better on that doc.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,98 @@
...
+## Scope
+
+1. Design the usage of [Kopia Repository Server](https://kopia.io/docs/repository-server/) as an separate on-demand workload
+   - In order to make use of user access control and cloud storage credential abstraction offered in the server-based operations 
+2. Re-work Kanister Functions like BackupData, RestoreData, CopyVolumeData, etc. with  Kopia workflows
+   - And leverage the use of Kopia Repository Server workload within these Functions
+
+### Design the usage of Kopia Repository Server
+
+
```
shlokchaudhari9, line 43: @pavannd1 Might need your help here to briefly state our CR/CRD design approach.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,98 @@
...
+Therefore, they are not ideally recommended to be used for protecting application data.
+
+Instead, we make use of v2 Blueprints whose phases call a `kando` utility to perform data snapshots.
+This tool internally executes `Kopia` commands to manage backups in the cloud.
+
+This approach is again not a robust solution because;
+1. We have to create a separate kopia profile and use it within kando command to pass necessary kopia server info
+2. We have to follow the correct kando command syntax in the Blueprint while backing up and mention `--output-name` in the Action outputArtifacts
+3. Kopia errors are not logged when kando commands fail in the acion phase
```
pavannd1, line 23:
```suggestion
Kanister also provides a command line utility `kando` that can be used to move data to and from object stores.
This tool internally executes `Kopia` commands to move the backup data.
The v2 version of the example Kanister Blueprints supports this. However, there are a few caveats to using these Blueprints.

1. `kando` uses `Kopia` only when a Kanister Profile of type `Kopia` is provided
2. A Kanister Profile of type `Kopia` requires a [Kopia Repository Server](https://kopia.io/docs/repository-server/) running in the same namespace as the Kanister controller
3. A Repository Server requires a [Kopia Repository](https://kopia.io/docs/repositories/) to be initialized on a backend storage target

Kanister currently lacks documentation and automation to use these features.
```

pavannd1 (commented):
Part 1 of my review. Yet to go through the newly created parts. Let me know what you think

design/kanister-kopia-integration.md
shlokchaudhari9: Thanks @pavannd1 :) Your suggestions look better worded 👍🏼

design/kanister-kopia-integration.md
PrasadG193: Can we design the data-mover interface in such a way that in the future, if we want to switch to another tool it will be easier than refactoring all the functions? I understand this could be out of the current scope but just wondering.

design/kanister-kopia-integration.md
pavannd1: This seems more like pre-requisites to using the server rather than limitations to me. We can reword accordingly

pavannd1:
@ihcsim I have updated the documentation to match your suggestions. PTAL
@shlokchaudhari9  I removed the images added as I think they belong with the server design doc and not here. However, I'm in the process of adding a simplified diagram to this design doc.

design/kanister-kopia-integration/kanister-kopia-integration.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Kopia Repository
+
+- As mentioned above, the backup storage location is called a "Repository" in Kopia.
+- A separate repository will be used for each application protected by Kopia-based Blueprints in Kanister.
+- The repository will be initialized when the Repository Server is created the first time. Once initialized, it will continue to exist at the location until cleaned up manually.
+- Accessing the repository requires location and credential information similar to a Kanister Profile CR and a unique password used by Kopia during [encryption](https://kopia.io/docs/features/#end-to-end-zero-knowledge-encryption).
+- In the first iteration, this password will be auto-generated by the Kanister controller. Future iterations will allow users to use a Key Management Service of choice.
+- Only a single repository can exist at a particular backend storage location. To address this, the Kanister controller will generate the repository path using the location information and the UUID of the application namespace. For example, the repository path for `mysql` namespace and S3 bucket called `test-bucket` will be of the form `test-bucket/<UUID of mysql namespace>`.
```
viveksinghggits, line 59: `Only a single repository can exist at a particular backend storage location.`
in the second point here, we are saying that we are going to create kopia repo (repo) for every application and in this sentence we are saying we can only have one repo in the backend storage location.
That seems a bit off? Because it would mean we can just backup one application in one backup storage location.

viveksinghggits: do we really need bucket name in the repo path?
Bucket name is already part of the object URL I think.

design/kanister-kopia-integration/kanister-kopia-integration.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Kopia Repository Server
+
+- A Kopia Repository Server allows Kopia clients proxy access to the backend storage location through it.
+- A separate server will be used for each repository initialized in Kanister.
```
viveksinghggits, line 64: a separate server would be used just for initialising the repo or for every action?

design/kanister-kopia-integration/kanister-kopia-integration.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Automate the initialization of a Kopia Repository for an application backed up by Kanister.
+2. Design and automate the lifecycle of the required Kopia Repository Server.
+3. Add new versions of Kanister Data Functions like BackupData, RestoreData, etc. with Kopia as the primary data mover tool.
+
+## User Experience
+
+- All the new features mentioned in this document will be opt-in only. Existing users will not see any changes in the Kanister controller's behavior.
+- Users will be able to continue using their current Blueprints, switch to the v2 version of the example Blueprints, or use Blueprints with the new version of the Kanister Data Functions.
+- Users opting to use the v2 Blueprints and Blueprints with Kopia-based Kanister Data Functions will be required to follow instructions to set up the required Kopia Repository Server before executing the actions.
```
ihcsim, line 47: WDYT of adding a start-up option like `--storage=kopia`, `--feature-gate=storage=kopia` etc., which will enable the new CRD and auto-provision the default single kopia server?

design/kanister-kopia-integration/kanister-kopia-integration.md
pavannd1, line 59:
```suggestion
- Only a single repository can exist at a particular path in the backend storage location. To address this, the Kanister controller will generate the repository path using the location information and the UUID of the application namespace. For example, the repository path for `mysql` namespace and S3 bucket called `test-bucket` will be of the form `test-bucket/<UUID of mysql namespace>`.
```

design/kanister-kopia-integration/kanister-kopia-integration.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Kopia Repository
+
+- As mentioned above, the backup storage location is called a "Repository" in Kopia.
+- A separate repository will be used for each application protected by Kopia-based Blueprints in Kanister.
+- The repository will be initialized when the Repository Server is created the first time. Once initialized, it will continue to exist at the location until cleaned up manually.
+- Accessing the repository requires location and credential information similar to a Kanister Profile CR and a unique password used by Kopia during [encryption](https://kopia.io/docs/features/#end-to-end-zero-knowledge-encryption).
+- In the first iteration, this password will be auto-generated by the Kanister controller. Future iterations will allow users to use a Key Management Service of choice.
+- Only a single repository can exist at a particular path in the backend storage location. To address this, the Kanister controller will generate the repository path using the location information and the UUID of the application namespace. For example, the repository path for `mysql` namespace and S3 bucket called `test-bucket` will be of the form `test-bucket/<UUID of mysql namespace>`.
```
ihcsim, line 59: Doesn't that mean backups of multiple applications in the same namespace will end up in the same repository at `test-bucket/<UUID of mysql namespace>`? I thought the model is one repository per application.

design/kanister-kopia-integration/kanister-kopia-integration.md
pavannd1: I have updated it to clarify the limitation. Yes, it is true that only a single repository can exist at a particular path which is why we need to add a unique string in the path to allow creation of multiple repos in the same bucket.

What do you mean by object URL? How can that be used in a generated string?

design/kanister-kopia-integration/kanister-kopia-integration.md
```diff
@@ -0,0 +1,91 @@
...
+4. Reduced memory consumption
+5. Increased variety of backend storage target for backups
+
+These improvements can be achieved by using `Kopia` as the primary data movement tool in these Kanister Functions.
+
+Kanister also provides a command line utility `kando` that can be used to move data to and from object stores.
+This tool internally executes `Kopia` commands to move the backup data.
+The v2 version of the example Kanister Blueprints supports this. However, there are a few caveats to using these Blueprints.
+1. `kando` uses `Kopia` only when a Kanister Profile of type `Kopia` is provided
+2. A Kanister Profile of type `Kopia` requires a [Kopia Repository Server](https://kopia.io/docs/repository-server/) running in the same namespace as the Kanister controller
```
ihcsim, line 23: Do you have an example of what the new server CRD looks like? I am wondering instead of duplicating profile, we can couple/reference the new CRD to a profile?

design/kanister-kopia-integration/kanister-kopia-integration.md
viveksinghggits: so let's say we have an object in the bucket  `infracloud.kanister.io` at the path `abc/def/a.txt`, the path to this object is going to look like 
```
s3://infracloud.kanister.io/abc/def/a.txt
```
but if we decide that the file is going to be present at path `<bucket>/abc/def/a.txt` in that case the path to object is going to look like
```
s3://<bucket>/<bucket>/abc/def/a.txt
```
I am not sure but I think is what going to happen.

pavannd1: That's not right. If the path provided by the user is `s3://<bucket>/` then the path to the repository will be `s3://<bucket>/<application namespace UID>`

design/kanister-kopia-integration/kanister-kopia-integration.md
pavannd1: There will be a long-running server for each repository and not for every action. Let me know if you want me to modify the statement to clarify this.

design/kanister-kopia-integration/kanister-kopia-integration.md
pavannd1: Do you mean the start-up of the controller? Sure, we can have a feature gate to enable the new CRD.
But note that the Blueprints are different for Kopia-based and stow/restic-based functions. If folks want to turn off the feature and use kopia blueprints, it won't be possible.

design/kanister-kopia-integration/kanister-kopia-integration.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Detailed Design
+
+### Kopia Repository
+
+- As mentioned above, the backup storage location is called a "Repository" in Kopia.
+- A separate repository will be used for each application protected by Kopia-based Blueprints in Kanister.
+- The repository will be initialized when the Repository Server is created the first time. Once initialized, it will continue to exist at the location until cleaned up manually.
+- Accessing the repository requires location and credential information similar to a Kanister Profile CR and a unique password used by Kopia during [encryption](https://kopia.io/docs/features/#end-to-end-zero-knowledge-encryption).
+- In the first iteration, this password will be auto-generated by the Kanister controller. Future iterations will allow users to use a Key Management Service of choice.
```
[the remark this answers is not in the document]
pavannd1: Ah yes! I think I remember this discussion.
We can have users supply a password for each repository. However, it is essential that the users store this password safely since there is no way to recover the repository if the password is lost.

design/kanister-kopia-integration/kanister-kopia-integration.md
pavannd1: That's right!
I was thinking of the namespace as the deduplication domain in Kopia. This would mean multiple applications in the same namespace share the repository and the backup data from all these applications will be deduped and stored together. There are pros and cons to this approach. We can change the granularity if required 👍🏼

design/kanister-kopia-integration/kanister-kopia-integration.md
pavannd1: @shlokchaudhari9 could you please post a CRD example when you get a chance?
I sort of agree with @julio-lopez's suggestion of keeping the server CRD independent of the profile CR. This would allow us to slowly deprecate the profiles altogether. There are security benefits as well as some drawbacks around the NFS backend that would be resolved with this approach. But to improve the experience for existing users, the tooling we provide around the creation of the CRD could include options to receive the information from a profile reference.

design/kanister-kopia-integration/kanister-kopia-integration.md
viveksinghggits: I think it's ok.

design/kopia.md
```diff
@@ -0,0 +1,461 @@
...
+  --config-file=/run/kopia/repo.config \
+  --server-cert-fingerprint=<redacted> \
+  --password=<redacted>
+```
+
+The `<service-name>` is the Kopia server's `Service` resource name. By
+convention, it includes the repository's `targetedNamespace`. E.g., the
+`Service` fronting the Kopia server of a repository serving the
+`monitoring` namespace can be `repo-monitoring`. Then its FQDN is
+`repo-monitoring.kanister.svc.cluster.local`.
```
shlokchaudhari9, line 219: What happens when the user wants to connect to a different repository on the same back-end location and operate on same application (`targetNamespace`)? 

If a new `Repository` CR is expected to be created and used here, how would the new CR differentiate between `Service` resources that include same `targetNamespace`? I think the resultant FQDN would be a point of conflict between such multiple `Repository` CRs.

ihcsim: Can they use a different `<service-name>` e.g., `repo2-monitoring`?

design/kopia.md
```diff
@@ -0,0 +1,461 @@
...
+
+The access username used to authenticate with the Kopia server is added to the
+`/run/kopia/repo.config` configuration file, following the
+[`repo.ClientOptions`][5] structure.
+
+The content of the configuration file, server certificate fingerprint, and
+and access password are exported to the remote data mover clients via the
+`KubeExec` functionality over TLS.
```
shlokchaudhari9, line 227: Not sure if I completely understand this. Would we use `outputArtifacts` to pass this information to all other `BackupData`, `RestoreData`, etc. actions defined in BP?

ihcsim: Yeah, this part still needs to be fleshed out. With `KubeTask`, we have more control over what goes into the container before it's started. With `KubeExec`, current thought is to `POST` these info to the `exec` subresource of the data mover pod via its `stdin`.

design/kopia.md
shlokchaudhari9: Yes. The `Service` resources can have some prefix or suffix to differentiate. Probably, we can use generate name in the `repository-controller` with the `<service-name>` to do that. But then, we would need to devise a strategy for passing the resultant `<service-name>` (or entire FQDN) to the data mover clients.

design/kopia.md
```diff
@@ -0,0 +1,471 @@
...
+  namespace: kanister
+  labels:
+    repo.kanister.io/target-namespace: monitoring
+spec:
+  # immutable
+  targetNamespace: monitoring
+  location:
+    # required
+    secretName: repository-monitoring-location
+  server:
```
shlokchaudhari9, line 48: What about Repository path or `repoPath` field? Are we going to internally create it using a pre-defined format and the location? @pavannd1

ihcsim: Depends if we want to give user the option to change it. To keep it simple, I think the controller should generate it, to avoid having the handle scenarios like what if user changed the path.

design/kanister-kopia-integration/kanister-kopia-integration.md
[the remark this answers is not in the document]
pavannd1: I'll be updating this to say that the user has to supply the password in a K8s secret.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,471 @@
...
+```yaml
+apiVersion: cr.kanister.io/v1alpha1
+kind: RepositoryServer
+metadata:
+   name: repository-monitoring
+   namespace: kanister
+   labels:
+      repo.kanister.io/target-namespace: monitoring
+spec:
+   location:
```
shlokchaudhari9, line 127: As mentioned in [this comment](https://github.com/kanisterio/kanister/pull/1645#discussion_r979895726), can we rename the `location` field here?

The Location struct for Profile CR [in pkg/apis/cr/v1alpha1/types.go](https://github.com/kanisterio/kanister/blob/dab83e87b16e7eeb73645ed0768450fd978ee2b7/pkg/apis/cr/v1alpha1/types.go#L255) is conflicting with this location field here

pavannd1: @shlokchaudhari9 Yes, I think `storage` might work here as well. I'll change it 👍🏼

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,471 @@
...
+      # required: storage location info
+      secret:
+         name: location
+         namespace: kanister
+      # required: creds to access the location
+      credentialSecret:
+         name: loc-creds
+         namespace: kanister
```
ihcsim, line 135: We should use the `secretRef` suffix, following the K8s API convention. E.g., see [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#csipersistentvolumesource-v1-core). In the code, this should be back by the [`corev1.SecretReference`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#secretreference-v1-core) struct.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,471 @@
...
+      # required: TLS certificate required for secure communication between the Kopia client and server
+      tlsSecret:
+         name: repository-monitoring-server-tls
+         namespace: kanister
+      # required: repository users list
+      userAccessSecret:
+         name: repository-monitoring-server-access
+         namespace: kanister
+      # required: selector for network policy required to enable cross-namespace access to the server
+      labelSelector:
```
ihcsim, line 164: Can we be more explicit about this being selectors for network policy, and not e.g., `Service`? How about, following the example in the network policy [doc](https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource), something like:

```yaml
server:
    networkPolicy:
        namespaceSelector:
            matchLabels:
                project: myproject
        podSelector:
            matchLabels:
                role: frontend
```

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,471 @@
...
+kopia repository connect s3 \
+   --bucket=my-bucket \
+   --access-key=<redacted> \
+   --secret-access-key=<redacted> \
+   [--endpoint=https://foo.example.com \]
+   [--prefix=my-prefix \]
+   [--region=us-west-1 \]
+   --password=<redacted> \
+   --override-hostname=<hostname> \
+   --override-username=<username>
```
ihcsim, line 354: Do we know if this is still necessary or is `kopia server start` smart enough to figure it out if we add all the repo info into the `repo.config` file?

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,471 @@
...
+   bucket: my-bucket
+   # optional: specified in case of S3-compatible stores
+   endpoint: https://foo.example.com
+   # optional: used as a sub path in the bucket for all backups
+   path: kanister/backups
+   # required, if supported by the provider
+   region: us-west-1
+   # optional: if set to true, do not verify SSL cert.
+   # Default, when omitted, is false
+   skipSSLVerify: false
```
ihcsim, line 205: We should document all the supported keys. to help the user to put this secret together. AFAICT, the only place where this is documented now is in the `kopia` CLI [code](https://github.com/kopia/kopia/blob/05e729a7858a6e86cb48ba29fb53cb6045efce2b/cli/storage_s3.go#L19-L27).

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,471 @@
...
+      rootPath: /repo/monitoring/
+      # required: password to access the repository
+      passwordSecret:
+         name: repository-monitoring-password
+         namespace: kanister
+      # required: these values will be used by the repository server controller
+      # for the connection from the server to the repository. Kopia will use
+      # these values to track and control operations
+      username: kanisterAdmin
+      hostname: monitoring.app
```
ihcsim, line 149: When and how do these get created? They can take any values, right?

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,471 @@
...
+   --config-file=/run/kopia/repo.config \
+   --server-cert-fingerprint=<redacted> \
+   --override-username=<username> \
+   --override-hostname=<hostname> \
+   --password=<redacted>
+```
+
+The `<service-name>` is the Kopia server's `Service` resource name.
+This will be auto-generated by the repository controller and provided via the `status` subresource of the `RepositoryServer` resource.
+The `server-cert-fingerprint` is derived from the TLS certificates provided during the creation of the server resource.
```
ihcsim, line 423: It's ok to add the fingerprint to the `status` too, since it's public.

ihcsim (commented):
I added some more questions and thoughts. Also, in general, I find that wrapping the lines to about 80 chars width (in text, markdown docs) makes it easier to add review comments, and improve readability.

design/kanister-kopia-integration.md
shlokchaudhari9: I agree with @ihcsim here. @pavannd1 we should make the `labelSelector` field more descriptive. Users might get confused about the type of these labels. They might question whether the labels here are pod-selectors or namespace-selectors. Better we explicitly ask them.

shlokchaudhari9: Also makes it easier for controller to define the ingress rule in `NetworkPolicy`. Rather than making assumptions about the kind of selector these labels would be, controller would directly use these selectors while creating the `NetworkPolicy`.

design/kanister-kopia-integration.md
pavannd1: A connect operation is required even if the configuration is added to a file.
We are exploring ways to utilize the `kopia repository connect from-config` option provided by Kopia to supply all the information via a single config file.

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,594 @@
...
+   storage:
+      # required: storage location info
+      secretRef:
+         name: location
+         namespace: kanister
+      # required: creds to access the location
+      # optional when using location type file-store
+      credentialSecretRef:
+         name: loc-creds
+         namespace: kanister
```
viveksinghggits, line 207: Is it safe to assume that these secrets are always going to be in the kanister (controller's) namespace. If yes, do we need the namespace field?

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,594 @@
...
+      # required: selector for network policy required to enable cross-namespace access to the server
+      networkPolicy:
+         namespaceSelector:
+            matchLabels:
+               app: monitoring
+         podSelector:
+            matchLabels:
+               pod: kopia-client
+               app: monitoring
+status:
```
shlokc9, line 245: @pavannd1 We also discussed about adding a new string field `isReady` in the status subresource that would convey the users if RepositoryServer is ready to use or not
1. Value would either be "True" if ready
2. Or would be "False" if an error occurred in the controller

design/kanister-kopia-integration.md
shlokc9: We have discussed this over a call with @pavannd1. This was intentionally added because we want the CR-controller to allow fetching secrets user create in their own namespace. This would be in-sync with the discussions to allow Blueprints or Profile CRs to be created in user/app namespace and not the controller(kanister) namespace

design/kanister-kopia-integration.md
```diff
@@ -0,0 +1,594 @@
...
+   version `v1.0.0-alpha`:
+
+   1. BackupData
+   2. BackupDataAll
+   3. BackupDataStats
+   4. CopyVolumeData
+   5. DeleteData
+   6. DeleteDataAll
+   7. RestoreData
+   8. RestoreDataAll
```
PrasadG193, line 116: As discussed, instead of adding versions, we would add new functions with the names having`*WithKopia` suffix.

PrasadG193: @shlokc9 can you add a note about this is yet to decide or TBD.

PrasadG193 (approved):

design/kanister-kopia-integration.md
shlokc9: Sure, I'll add a note about this. 👍🏼 
Maybe a separate design would be needed for handling kopia clients inside Kanister Functions....

PR #1539, opened 2022-07-04
Update ubi-minimal base image to ubi-minimal:8.6-854

infraq:
Update ubi-minimal base image to ubi-minimal:8.6-854


viveksinghggits:
Is Kanister integration test sufficient to test these upgrades are not going to break anything?

PR #1555, opened 2022-07-15
Vsphere: list snapshot by description

AKhoria:
## Change Overview

Add alternative way to list snapshots in Fcd provider by using description field
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/blockstorage/vmware/vmware.go
```diff
@@ -253,7 +255,8 @@ func (p *FcdProvider) SnapshotCreate(ctx context.Context, volume blockstorage.Vo
 	var res types.AnyType
 	err := wait.PollImmediate(time.Second, defaultRetryLimit, func() (bool, error) {
 		log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
-		task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), noDescription)
+		description := generateDescription(tags)
+		task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), description)
```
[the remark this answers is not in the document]
AKhoria: Thanks, @carlbraganza 
Yes, I'm currently working on this part. Here is a commit, I'm going to test it today
https://github.com/kanisterio/kanister/pull/1555/commits/8c7cc05000f94fdd8ba2e5a8415b7e7b1b472d15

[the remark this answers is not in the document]
ihcsim: Instead of adding multiple checks to see if the snapshot is truly created in `wait.PollImmediate`, can we split it up then snapshot creation and error handling parts? Feels like separating them makes the error handling part easier to reason about, and more testable.

So something like this:
```diff
diff --git a/pkg/blockstorage/vmware/vmware.go b/pkg/blockstorage/vmware/vmware.go
index 9952c040..4beee14b 100644
--- a/pkg/blockstorage/vmware/vmware.go
+++ b/pkg/blockstorage/vmware/vmware.go
@@ -255,48 +255,52 @@ func (p *FcdProvider) SnapshotCreate(ctx context.Context, volume blockstorage.Vo
 	var res types.AnyType
 	description := generateDescription(tags)
 	err := wait.PollImmediate(time.Second, defaultRetryLimit, func() (bool, error) {
-		log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
-		task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), description)
-		if lerr != nil {
-			res := p.getCreatedSnapshot(ctx, tags, volume.ID, time.Now().Add(-1*vmWareTimeout))
-			if res != nil {
-				log.Error().WithError(lerr).Print("Wait failed but snapshot was created")
-				return true, nil
+		createErr := func() error {
+			log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
+			task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), description)
+			if lerr != nil {
+				return errors.Wrap(lerr, "CreateSnapshot task creation failure")
 			}
-			return false, errors.Wrap(lerr, "CreateSnapshot task creation failure")
-		}
-		log.Debug().Print("Started CreateSnapshot task", field.M{"VolumeID": volume.ID})
-		res, lerr = task.Wait(ctx, vmWareTimeout)
-		if lerr != nil {
-			res := p.getCreatedSnapshot(ctx, tags, volume.ID, time.Now().Add(-1*vmWareTimeout))
-			if res != nil {
-				log.Error().WithError(lerr).Print("Wait failed but snapshot was created")
-				return true, nil
+			log.Debug().Print("Started CreateSnapshot task", field.M{"VolumeID": volume.ID})
+			res, lerr = task.Wait(ctx, vmWareTimeout)
+			if lerr != nil {
+				return lerr
 			}
-			// check if snapshot was created
-			if soap.IsVimFault(lerr) {
-				switch soap.ToVimFault(lerr).(type) {
-				case *types.InvalidState:
-					log.Error().WithError(lerr).Print("There is some operation, other than this CreateSnapshot invocation, on the VM attached still being protected by its VM state. Will retry", field.M{"VolumeID": volume.ID})
-					return false, nil
-				case *vslmtypes.VslmSyncFault: // potentially can leak snapshots
-					if vFault, ok := soap.ToVimFault(lerr).(*vslmtypes.VslmSyncFault); ok {
-						log.Error().Print(fmt.Sprintf("VslmSyncFault: %#v", vFault))
-					}
-					if !(govmomiError{lerr}).Matches(reVslmSyncFaultFatal) {
-						log.Error().Print(fmt.Sprintf("CreateSnapshot failed with VslmSyncFault. Will retry: %s", (govmomiError{lerr}).Format()), field.M{"VolumeID": volume.ID})
-						return false, nil
-					}
-					return false, errors.Wrap(lerr, "CreateSnapshot failed with VslmSyncFault. A snapshot may have been created by this failed operation")
-				case *types.NotFound:
-					log.Error().WithError(lerr).Print("CreateSnapshot failed with NotFound error. Will retry", field.M{"VolumeID": volume.ID})
+			return nil
+		}()
+
+		if createErr == nil {
+			return true, nil
+		}
+
+		// it's possible that snapshot was created despite of SOAP errors
+		res := p.getCreatedSnapshot(ctx, tags, volume.ID, time.Now().Add(-1*vmWareTimeout))
+		if res != nil {
+			log.Error().WithError(createErr).Print("snapshot created with errors")
+			return true, nil
+		}
+
+		// snapshot wasn't created, handle the different SOAP errors then retry
+		if soap.IsVimFault(createErr) {
+			switch soap.ToVimFault(createErr).(type) {
+			case *types.InvalidState:
+				log.Error().WithError(createErr).Print("There is some operation, other than this CreateSnapshot invocation, on the VM attached still being protected by its VM state. Will retry", field.M{"VolumeID": volume.ID})
+				return false, nil
+			case *vslmtypes.VslmSyncFault: // potentially can leak snapshots
+				if vFault, ok := soap.ToVimFault(createErr).(*vslmtypes.VslmSyncFault); ok {
+					log.Error().Print(fmt.Sprintf("VslmSyncFault: %#v", vFault))
+				}
+				if !(govmomiError{createErr}).Matches(reVslmSyncFaultFatal) {
+					log.Error().Print(fmt.Sprintf("CreateSnapshot failed with VslmSyncFault. Will retry: %s", (govmomiError{createErr}).Format()), field.M{"VolumeID": volume.ID})
 					return false, nil
 				}
+				return false, errors.Wrap(createErr, "CreateSnapshot failed with VslmSyncFault. A snapshot may have been created by this failed operation")
+			case *types.NotFound:
+				log.Error().WithError(createErr).Print("CreateSnapshot failed with NotFound error. Will retry", field.M{"VolumeID": volume.ID})
+				return false, nil
 			}
-			return false, errors.Wrap(lerr, "Failed to wait on task")
 		}
-		log.Debug().Print("CreateSnapshot task complete", field.M{"VolumeID": volume.ID})
-		return true, nil
+		return false, errors.Wrap(createErr, "Failed to wait on task")
 	})
 	if err != nil {
 		log.Error().WithError(err).Print(fmt.Sprintf("Failed to create snapshot for FCD %s: %s", volume.ID, govmomiError{err}.Format()))
```
Feels like this is also more testable, if we extract them out into separate functions.

Note: just a take-it-or-leave-it suggestion.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -51,11 +51,13 @@ const (
...
 	VSphereIsParaVirtualizedKey = "VSphereIsParaVirtualizedKey"
 
-	noDescription     = ""
 	defaultWaitTime   = 60 * time.Minute
 	defaultRetryLimit = 30 * time.Minute
 
 	vmWareTimeoutMinEnv = "VMWARE_GOM_TIMEOUT_MIN"
+
+	DescriptionTag = "kanister-8d6b8dc9ceb74bb4b47b3c0a408658c4"
+	VolumeIdTag    = "kanister-volume-8d6b8dc9ceb74bb4b47b3c0a408658c4"
```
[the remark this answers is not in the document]
AKhoria: I tried to avoid possibility to overlap with user defined variables. Generated GUID seemed to be the best way. I think we should never change them so user could rely on those const when use kanister

pkg/blockstorage/vmware/vmware.go
```diff
@@ -310,6 +324,40 @@ func (p *FcdProvider) SnapshotCreate(ctx context.Context, volume blockstorage.Vo
...
+		if strings.HasPrefix(name, DescriptionTag) {
+			tagsForFilter[name] = value
+		}
+	}
+	if len(tagsForFilter) == 0 {
+		return nil
+	}
+	tagsForFilter[VolumeIdTag] = volID
+	sns, err := p.SnapshotsList(ctx, tagsForFilter)
+	if len(sns) == 1 && err == nil {
```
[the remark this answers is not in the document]
AKhoria: I'm trying to get snapshot which was created in that session, so yes we assume there will be only one snapshot for this volume + tag + timeframe. if we found more than one - something is wrong(maybe two creation with the same tags are run in parallel) and we don't know which spapshot we can reuse so we shouldn't reuse them at all

pkg/blockstorage/vmware/vmware.go
```diff
@@ -570,6 +618,46 @@ func (p *FcdProvider) getSnapshotTags(ctx context.Context, fullSnapshotID string
...
+	var result []*blockstorage.Snapshot
+
+	filterStr := generateDescription(descriptionTags)
+	for _, volID := range volumeIDs {
+		snapshots, _ := p.Gom.RetrieveSnapshotInfo(ctx, vimID(volID))
+		for _, snapshot := range snapshots {
+			if snapshot.Description == filterStr {
+				sn, err := convertFromObjectToSnapshot(&snapshot, volID)
+				if err != nil {
+					return nil, errors.Wrap(err, "Failed to convert object to snapshot")
```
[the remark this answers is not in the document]
AKhoria: But if the found snapshot has this description it is the one we are looking for.
Also, that's exactly our current logic, pls check the snapshotsListByTag function: line 681 ->  line 438

pkg/blockstorage/vmware/vmware.go
```diff
@@ -570,6 +618,46 @@ func (p *FcdProvider) getSnapshotTags(ctx context.Context, fullSnapshotID string
...
+	var volumeIDs []string
+	var descriptionTags = map[string]string{}
+	for name, value := range tags {
+		if strings.HasPrefix(name, DescriptionTag) {
+			descriptionTags[name] = value
+		}
+		if name == VolumeIdTag {
+			volumeIDs = strings.Split(value, ",")
+		}
+	}
```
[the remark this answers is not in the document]
AKhoria: That's a very good point, I'll fix this

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
AKhoria: Yes, I've also noticed that I have a duplicates but I didn't find any suitable solution. Your proposal looks good I'll take it

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
ihcsim: Generated GUID sounds good to me. Are these constants also shared with downstream? Just want to make sure they don't go out-of-sync, if used in different projects.

ihcsim:
Code change LGTM. @carlbraganza @bathina2 PTAL and see if it's mergeable. Thanks.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -251,38 +253,54 @@ var reVslmSyncFaultFatal = regexp.MustCompile("Change tracking invalid or disk i
...
-					if !(govmomiError{lerr}).Matches(reVslmSyncFaultFatal) {
-						log.Error().Print(fmt.Sprintf("CreateSnapshot failed with VslmSyncFault. Will retry: %s", (govmomiError{lerr}).Format()), field.M{"VolumeID": volume.ID})
-						return false, nil
-					}
-					return false, errors.Wrap(lerr, "CreateSnapshot failed with VslmSyncFault. A snapshot may have been created by this failed operation")
-				case *types.NotFound:
-					log.Error().WithError(lerr).Print("CreateSnapshot failed with NotFound error. Will retry", field.M{"VolumeID": volume.ID})
+
+		// it's possible that snapshot was created despite of SOAP errors
+		res = p.getCreatedSnapshot(ctx, tags, volume.ID, time.Now().Add(-1*vmWareTimeout))
```
carlbraganza, line 277: Shouldn't you just save the time `epoch` when the function was called?

pkg/blockstorage/vmware/vmware.go
```diff
@@ -251,38 +253,54 @@ var reVslmSyncFaultFatal = regexp.MustCompile("Change tracking invalid or disk i
...
+			task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), description)
+			if lerr != nil {
+				return errors.Wrap(lerr, "CreateSnapshot task creation failure")
+			}
+			log.Debug().Print("Started CreateSnapshot task", field.M{"VolumeID": volume.ID})
+			res, lerr = task.Wait(ctx, vmWareTimeout)
+			if lerr != nil {
+				return lerr
+			}
+			return nil
```
carlbraganza, line 269: Suggestion:
```
return lerr
```

pkg/blockstorage/vmware/vmware.go
```diff
@@ -251,38 +253,54 @@ var reVslmSyncFaultFatal = regexp.MustCompile("Change tracking invalid or disk i
...
 	var res types.AnyType
+	description := generateDescription(tags)
 	err := wait.PollImmediate(time.Second, defaultRetryLimit, func() (bool, error) {
-		log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
-		task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), noDescription)
-		if lerr != nil {
-			return false, errors.Wrap(lerr, "CreateSnapshot task creation failure")
+		createErr := func() error {
+			log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
+			task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), description)
```
carlbraganza, line 260: Why `lerr` and not, say, `cerr` or `cErr` or `createErrror`? It was inexplicable in the original code; no need to carry bad names forward...

pkg/blockstorage/vmware/vmware.go
```diff
@@ -251,38 +253,54 @@ var reVslmSyncFaultFatal = regexp.MustCompile("Change tracking invalid or disk i
 // SnapshotCreate is part of blockstorage.Provider
 func (p *FcdProvider) SnapshotCreate(ctx context.Context, volume blockstorage.Volume, tags map[string]string) (*blockstorage.Snapshot, error) {
 	var res types.AnyType
+	description := generateDescription(tags)
 	err := wait.PollImmediate(time.Second, defaultRetryLimit, func() (bool, error) {
-		log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
-		task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), noDescription)
-		if lerr != nil {
-			return false, errors.Wrap(lerr, "CreateSnapshot task creation failure")
+		createErr := func() error {
```
carlbraganza, line 258: This is admittedly subjective, but I don't like the style of declaring functions within functions where not strictly necessary. Here the outer wait loop will be evaluating the function creation code each pass (though the compiler may optimize things).
I think it may be clearer if you were to define a set of helper functions before the wait loop and invoke them with arguments (instead of relying on lexical scoping)?  Perhaps even make these helpers methods on `FcdProvider` - could aid in unit testing!

pkg/blockstorage/vmware/vmware.go
```diff
@@ -310,6 +328,47 @@ func (p *FcdProvider) SnapshotCreate(ctx context.Context, volume blockstorage.Vo
...
+	}
+
+	if len(filteredSns) == 1 && err == nil {
+		return filteredSns[0]
+	}
+	if err != nil {
+		log.Error().WithError(err).Print("Failed to list when checking failed creation")
+	}
+	if len(filteredSns) > 1 {
+		log.Error().Print("More than one snapshot was found")
```
carlbraganza, line 357: Log the IDs - this would be very useful in the field.

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
carlbraganza: I don't think you need to get paranoid over this as long as the string is distinct enough. However, easier on the eye than a GUID would be a pseudo-DNS like name: e.g. `kanisterio.kanister/description` and `kanisterio.kanister/volume-id`

pkg/blockstorage/vmware/vmware.go
```diff
@@ -310,6 +328,47 @@ func (p *FcdProvider) SnapshotCreate(ctx context.Context, volume blockstorage.Vo
...
+	}
+	return nil
+}
+
+func generateDescription(tags map[string]string) string {
```
carlbraganza, line 362: Suggestion: `generateDescriptionTag`

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
carlbraganza: BTW,  from the code it appears that `VolumeIdTag` should be `VolumeIdListTag`.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -570,6 +629,36 @@ func (p *FcdProvider) getSnapshotTags(ctx context.Context, fullSnapshotID string
 
 // SnapshotsList is part of blockstorage.Provider
 func (p *FcdProvider) SnapshotsList(ctx context.Context, tags map[string]string) ([]*blockstorage.Snapshot, error) {
+	filterStr := generateDescription(tags)
+	if filterStr == "" {
+		return p.snapshotsListByTag(ctx, tags)
+	}
```
carlbraganza, line 635: I think the function would flow smoother if the "true" part of the check invoked a description specific method. e..g
```
func (p *FcdProvider) SnapshotsList(ctx context.Context, tags map[string]string) ([]*blockstorage.Snapshot, error) {
  if snapDesc := generateDescription(tags); snapDesc != "" {
    return p.snapshotsListByDescription(ctx, snapDesc, tags)
  }

  return p.snapshotListByTag(ctx, tags)
}

File: pkg/blockstorage/vmware/vmware.go [hunk 13]
```diff
@@ -570,6 +629,36 @@ func (p *FcdProvider) getSnapshotTags(ctx context.Context, fullSnapshotID string
...
+
+	volumeIDs := strings.Split(tags[VolumeIdTag], ",")
```
carlbraganza, line 637: Suggestion: `VolumeIdListTag`

pkg/blockstorage/vmware/vmware.go
```diff
@@ -570,6 +629,36 @@ func (p *FcdProvider) getSnapshotTags(ctx context.Context, fullSnapshotID string
...
+	if len(volumeIDs) == 0 {
+		log.Debug().Print("vSphere can't list by description without list of volumes. Cannot list snapshots")
+		return nil, nil
```
carlbraganza, line 640: This message would not be visible in K10 which discards debug.  I suggest defining a specific error and returning it.

pkg/blockstorage/vmware/vmware.go
carlbraganza: Why pass in `tags`? Shouldn't the arguments just be:

- context
- description string
- volume ID
- epoch

pkg/blockstorage/vmware/vmware.go
carlbraganza: The `p.snapshotListByDescription` function could be used directly by `getCreatedSnapshot`

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
ihcsim: Mainly because I want to understand the scope and side-effect these values have on downstream. Hard to tell at the moment without comments in the code, or searching through PRs in other projects. Hope the question is reasonable and not overly paranoia.

pkg/blockstorage/vmware/vmware.go
AKhoria: We have a similar approach in the previous ListSnapshot function, are you sure we need to introduce breaking changes here?

pkg/blockstorage/vmware/vmware.go
AKhoria: By some reason this name is used everywhere in the code, so I thought it has some meaning

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
AKhoria: Yes, this const should be use from other projects, so they always will be synchronised even if we change the name in future(but if we change it, snapshots created in older version becomes inaccessible) 
@carlbraganza if you think that dns-like naming is enough I'll change this

pkg/blockstorage/vmware/vmware.go
AKhoria: I don't know, honestly, on the other hand having function inside the loop helps understand what is happening in the code, you should jump to the function. Let's extract it to the helper and ask @ihcsim what does he think

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
carlbraganza: Yes, DNS like naming is sufficiently unique.

pkg/blockstorage/vmware/vmware.go
ihcsim: I am afraid I don't have much more thoughts re: unit tests. We should avoid conflating the scope of this PR. As previously [mentioned](https://github.com/kanisterio/kanister/pull/1555#discussion_r926078367), I proposed this refactoring because I felt it would make the code easier to read, and possibly more testable; didn't mean to imply that we add more tests in this PR. What you got still LGTM :+1:.

Also, I did it that way to make the diff easier to see. Other times, I used embedded functions or closures when I don't want to share the function with other callers, or avoid introducing a function with long list arguments. It's just another tool in the toolkit.

pkg/blockstorage/vmware/vmware.go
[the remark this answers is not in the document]
ihcsim: > but if we change it, snapshots created in older version becomes inaccessible

Thanks for the clarification. I'd think this is worth mentioning at least mention in a code comment, to cautious others who may stumble across this code in the future, to avoid breaking other projects.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -251,38 +256,43 @@ var reVslmSyncFaultFatal = regexp.MustCompile("Change tracking invalid or disk i
 // SnapshotCreate is part of blockstorage.Provider
 func (p *FcdProvider) SnapshotCreate(ctx context.Context, volume blockstorage.Volume, tags map[string]string) (*blockstorage.Snapshot, error) {
 	var res types.AnyType
+	description := generateSnapshotDescription(tags)
 	err := wait.PollImmediate(time.Second, defaultRetryLimit, func() (bool, error) {
-		log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
-		task, lerr := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), noDescription)
-		if lerr != nil {
-			return false, errors.Wrap(lerr, "CreateSnapshot task creation failure")
+		timeOfCreateSnapshotCall := time.Now()
```
carlbraganza, line 261: This should be set before the call to `wait.PollImmediate` function unless you want it reset within the implicit polling loop.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
...
+	task, err := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), description)
+	if err != nil {
+		return nil, errors.Wrap(err, "CreateSnapshot task creation failure")
+	}
+	log.Debug().Print("Started CreateSnapshot task", field.M{"VolumeID": volume.ID})
+	res, err := task.Wait(ctx, vmWareTimeout)
+	return res, err
+}
+
+func (p *FcdProvider) getCreatedSnapshot(ctx context.Context, description string, volID string, notEarlierThan time.Time) interface{} {
```
carlbraganza, line 501: Why a return type of `interface{}`?  You are treating the returned value as a `bool` so use that type!

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
...
+
+func (p *FcdProvider) createSnapshotAndWaitForCompletion(volume blockstorage.Volume, ctx context.Context, description string) (types.AnyType, error) {
+	log.Debug().Print("CreateSnapshot", field.M{"VolumeID": volume.ID})
+	task, err := p.Gom.CreateSnapshot(ctx, vimID(volume.ID), description)
+	if err != nil {
+		return nil, errors.Wrap(err, "CreateSnapshot task creation failure")
+	}
+	log.Debug().Print("Started CreateSnapshot task", field.M{"VolumeID": volume.ID})
+	res, err := task.Wait(ctx, vmWareTimeout)
+	return res, err
```
carlbraganza, line 498: nit:
```
return task.Wait(ctx, vmWareTimeout)
```

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
...
+	var tagsAsStr []string
+	for name, value := range tags {
+		if strings.HasPrefix(name, DescriptionTag) {
+			tagsAsStr = append(tagsAsStr, fmt.Sprintf("%s:%s", name, value))
+		}
+	}
+	return strings.Join(tagsAsStr, ",")
+}
+
+func getNotDscriptionTags(tags map[string]string) map[string]string {
```
carlbraganza, line 532: suggestion: `getTagsWithoutDescription`

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
...
+	if err != nil {
+		return nil, errors.Wrap(err, "CreateSnapshot task creation failure")
+	}
+	log.Debug().Print("Started CreateSnapshot task", field.M{"VolumeID": volume.ID})
+	res, err := task.Wait(ctx, vmWareTimeout)
+	return res, err
+}
+
+func (p *FcdProvider) getCreatedSnapshot(ctx context.Context, description string, volID string, notEarlierThan time.Time) interface{} {
+	var filteredSns []blockstorage.Snapshot
```
carlbraganza, line 502: Why would you not declare this to be `[]*bockstorage.Snapshot`? Why copy the datastructure?  On line 506 it is safe to copy the value of the range variable to some other data structure, if that was what you were concerned about.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
 	return cat.ID, nil
 }
 
+// SnapshotsList is part of blockstorage.Provider
+func (p *FcdProvider) SnapshotsList(ctx context.Context, tags map[string]string) ([]*blockstorage.Snapshot, error) {
+	if filterStr := generateSnapshotDescription(tags); filterStr != "" {
+		volumeIDs := strings.Split(tags[VolumeIdListTag], ",")
+		if len(volumeIDs) == 0 {
+			return nil, errors.New("vSphere can't list by description without list of volumes. Cannot list snapshots")
+		}
```
carlbraganza, line 484: Or you could just default to `p.snapshotsListByTag`...

carlbraganza (commented):
Looking pretty good. I still have a question or two...

pkg/blockstorage/vmware/vmware.go
AKhoria: I put it inside PollImmediate because we  will try to create a new snapshot on the next run, so there will be leakage on kanister level anyway. So we only re-try to get snapshot created in the current loop. The higher level logic  (that's why we adding a more reliable way to list snapshots)

pkg/blockstorage/vmware/vmware.go
```diff
@@ -51,11 +51,16 @@ const (
...
 	defaultWaitTime   = 60 * time.Minute
 	defaultRetryLimit = 30 * time.Minute
 
 	vmWareTimeoutMinEnv = "VMWARE_GOM_TIMEOUT_MIN"
+
+	// DescriptionTag is the prefix of the tags that should be placed in the snapshot description.
+	// This constant must be used by clients, so changing this field may make already created snapshots inaccessible.
+	DescriptionTag = "kanister.fcd.description"
+	// VolumeIdListTag is the predefined name of the tag which contains volume ids separated by comma
+	VolumeIdListTag = "kanister.fcd.volume-id"
```
denisvmedia, line 63: nit
```suggestion
	VolumeIDListTag = "kanister.fcd.volume-id"
```

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
...
+		return nil, errors.Wrap(err, "CreateSnapshot task creation failure")
+	}
+	log.Debug().Print("Started CreateSnapshot task", field.M{"VolumeID": volume.ID})
+	res, err := task.Wait(ctx, vmWareTimeout)
+	return res, err
+}
+
+func (p *FcdProvider) getCreatedSnapshot(ctx context.Context, description string, volID string, notEarlierThan time.Time) interface{} {
+	var filteredSns []blockstorage.Snapshot
+	sns, err := p.snapshotsListByDescription(ctx, []string{volID}, description)
```
denisvmedia, line 503: Shouldn't err be processed before iterating over sns?

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
...
+	for name, value := range tags {
+		if strings.HasPrefix(name, DescriptionTag) {
+			tagsAsStr = append(tagsAsStr, fmt.Sprintf("%s:%s", name, value))
+		}
+	}
+	return strings.Join(tagsAsStr, ",")
+}
+
+func getNotDscriptionTags(tags map[string]string) map[string]string {
+	result := make(map[string]string)
```
denisvmedia, line 533: IMO, this would be a little bit nicer in terms of allocation:

```suggestion
	result := make(map[string]string, len(tags))
```

It will allocate a little bit more than required (depending on the HasPrefix evaluation), but it will ensure no multiple allocations will occur.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -465,6 +475,78 @@ func (p *FcdProvider) GetCategoryID(ctx context.Context, categoryName string) (s
...
+	for name, value := range tags {
+		if !strings.HasPrefix(name, DescriptionTag) {
+			result[name] = value
+		}
+	}
+	return result
+}
+
+func getSnapshotsIDs(snapshots []blockstorage.Snapshot) []string {
+	var result []string
```
denisvmedia, line 543: For the same reasoning as above.

```suggestion
	result := make([]string, 0, len(snapshots))
```

pkg/blockstorage/vmware/vmware_manual_test.go
```diff
@@ -0,0 +1,54 @@
+package vmware
+
+import (
+	"context"
+
+	"github.com/gofrs/uuid"
+	"gopkg.in/check.v1"
+	. "gopkg.in/check.v1"
```
denisvmedia, line 8: These two lines look suspicious when imported at once.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -251,38 +256,43 @@ var reVslmSyncFaultFatal = regexp.MustCompile("Change tracking invalid or disk i
...
+
+		// it's possible that snapshot was created despite of SOAP errors
+		res = p.getCreatedSnapshot(ctx, description, volume.ID, timeOfCreateSnapshotCall)
+		if res != nil {
+			log.Error().WithError(createErr).Print("snapshot created with errors")
+			return true, nil
+		}
+
+		// snapshot wasn't created, handle the different SOAP errors then retry
+		if soap.IsVimFault(createErr) {
```
denisvmedia, line 276: If we invert the condition, we can reduce the level of indentation.

pkg/blockstorage/vmware/vmware.go
```diff
@@ -251,38 +256,43 @@ var reVslmSyncFaultFatal = regexp.MustCompile("Change tracking invalid or disk i
...
+			switch soap.ToVimFault(createErr).(type) {
```
denisvmedia, line 277: By assigning the type assertion call, we can avoid double assertion in the switch cases.

```suggestion
			switch t := soap.ToVimFault(createErr).(type) {
```

pkg/blockstorage/vmware/vmware.go
AKhoria: We are treating its as Types.ID, changed the code to returnung Types.ID explicitly

pkg/blockstorage/vmware/vmware.go
AKhoria: looks like overoptimization for me in this case, but OK, will change it

carlbraganza (approved):

PR #1570, opened 2022-07-27
Allow GCP buckets in other regions

KastenMike:
This PR lets you create buckets in specific regions. Currently default args are used (StorageClass: "MULTI_REGIONAL", Location: "US"); Location default is mentioned [here](https://pkg.go.dev/cloud.google.com/go/storage@v0.38.0#BucketAttrs), StorageClass through testing. Settings are passed via `config.Region` like S3 does in `s3Config` func.

Example test (placed inside a new dir+file mike/mike_test.go in my case):

```
func TestRegionalBucket(t *testing.T) {
	fmt.Println("TestRegionalBucket......")
	projectId := "projectId goes here"
	configJSON := "configJSON goes here"
	bucket := "mimckayd-euro"
	location := "europe-west8"
	providerType := objectstore.ProviderTypeGCS

	ctx := context.Background()
	pc := objectstore.ProviderConfig{
		Type:     providerType,
		Region:   location,
	}
	secret := &objectstore.Secret{}
	secret.Type = objectstore.SecretTypeGcpServiceAccountKey
	secret.Gcp = &objectstore.SecretGcp{
		ProjectID:  projectId,
		ServiceKey: configJSON,
	}
	p, err := objectstore.NewProvider(ctx, pc, secret)
	if err != nil {
		t.Errorf("NO NewProvider! %v", err)
	}

	b, err := p.CreateBucket(ctx, bucket)
	if err != nil {
		t.Errorf("NO BUCKET! %v", err)
	}

	fmt.Println("bucket created...", b.String())
}
```

I wanted to add some kind of test to `objectstore_test` like this:
```
func (s *ObjectStoreProviderSuite) TestGCSRegions(c *C) {
	if s.osType != ProviderTypeGCS {
		c.Skip("Test only applicable to GCS")
	}
	gcsRegions := []string{"us-west1", "europe-west8", "asia-south2"}
	ctx := context.Background()
	for _, region := range gcsRegions {
		s.initProvider(c, region)
		bucketName := s.createBucketName(c)
		_, err := s.provider.CreateBucket(ctx, bucketName)
		c.Check(err, IsNil)
		// try to find region from d~
		s.provider.DeleteBucket(ctx, bucketName)
	}
}
```

But at this point it isn't possible; it requires another update to stow to support it (exporting the real region), in particular: https://github.com/kastenhq/stow/pull/27

Can be added in a future PR after that goes in


pkg/objectstore/objectstore.go
```diff
@@ -145,15 +145,20 @@ func s3Config(ctx context.Context, config ProviderConfig, secret *Secret) (stowK
...
+	cm := stow.ConfigMap{}
 	if secret != nil {
 		if secret.Type != SecretTypeGcpServiceAccountKey {
 			return "", nil, errors.Errorf("invalid secret type %s", secret.Type)
 		}
 		configJSON = secret.Gcp.ServiceKey
 		projectID = secret.Gcp.ProjectID
+		if config.Region != "" {
+			cm[stowgcs.ConfigLocation] = config.Region
+			cm[stowgcs.ConfigStorageClass] = REGIONAL
```
KastenMike, line 160: Alternatively `config.StorageClass` could be added but it would only be relevant for GCS

ankitjain235 (approved):
LGTM.

Issue #1576, opened 2022-07-29
[BUG] Kanister Controller cannot update Status on Kubernetes versions < v1.20


pavannd1:
**Describe the bug**
Kanister Controller fails to add `Status` to ActionSets on Kubernetes Clusters of version < v1.20. 

**To Reproduce**
1. Build controller image from `master` branch
2. Install it on a cluster running K8s version 1.19 or lower
3. Create a Blueprint and an ActionSet to execute actions
4. Kanister controller fails to update the status
```
  Warning  Update Failed  5s    Kanister Controller  Could not update ActionSet: ActionSet.cr.kanister.io "hookbp-wmwr2" is invalid: status.progress.lastTransitionTime: Invalid value: "null": status.progress.lastTransitionTime in body must be of type string: "null"
```

**Expected behavior**
ActionSet should successfully run all the actions.

**Screenshots**
Example Blueprint:
```
apiVersion: cr.kanister.io/v1alpha1
kind: Blueprint
metadata:
  name: hook-blueprint
  namespace: kanister
actions:
  post-export:
    kind: Namespace
    phases:
    - args:
        command:
        - /bin/sh
        - -c
        - |
          echo "This is a test phase"
        image: alpine:latest
        podOverride:
          serviceAccountName: myrelease-kanister-operator
      func: KubeTask
      name: hookPhase
```

Example ActionSet:
```
apiVersion: cr.kanister.io/v1alpha1
kind: ActionSet
metadata:
  generateName: hookbp-
  namespace: kanister
spec:
  actions:
  - name: post-export
    blueprint: hook-blueprint
    object:
      kind: Namespace
      name: default
      namespace: default
```

Describe ActionSet after creation:
```
Name:         hookbp-wmwr2
Namespace:    kanister
Labels:       <none>
Annotations:  <none>
API Version:  cr.kanister.io/v1alpha1
Kind:         ActionSet
Metadata:
  Creation Timestamp:  2022-07-29T22:46:40Z
  Generate Name:       hookbp-
  Generation:          1
  Managed Fields:
    API Version:  cr.kanister.io/v1alpha1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:generateName:
      f:spec:
        .:
        f:actions:
    Manager:         kubectl-create
    Operation:       Update
    Time:            2022-07-29T22:46:40Z
  Resource Version:  57845
[...]

PR #1487, opened 2022-06-16
Update Controller To Track Action Progress

ihcsim:
## Change Overview

This PR updates the controller to track the progress of an actionset, by assessing the phase states. The progress tracker runs in its own goroutine with a 2-second ticker. The progress computation assigns weight to a phase, per the underlying Kanister Function the supports the phase. If a Kanister Function is marked as a long-running one, its phase will be assigned the `heavy` weight. Otherwise, the phase has the `normal` weight.

The actionset progress always starts at 10%, to distinguish started/running phases from idle ones, and ends at 100%.

The computation is also done on a best-effort basis, with no elaborated retries on update errors, to ensure the progress goroutine doesn't contend for resources nor contribute to performance latency. In other words, it's possible that the progress goes from 10% to 100% with no immediate updates.

Design doc : https://github.com/kanisterio/kanister/blob/master/design/progress-tracking.md

Out-of-scope: Progress computation of each phase based on log outputs are slated for future PRs.

Release note: The `actionSet` CRD must be upgraded manually, as Helm doesn't upgrade CRD.

## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test

## Issues

- #1455

## Test Plan

Deploy this multi-actions blueprint, where each phase has different completion time:

```sh
cat <<EOF | kubectl apply -f -
apiVersion: cr.kanister.io/v1alpha1
kind: Blueprint
metadata:
  name: actionset-progress
  namespace: kanister
actions:
  action0:
    phases:
    - func: KubeTask
      name: action0-task0
      args:
        namespace: "{{ .Namespace.Name }}"
        image: ghcr.io/kanisterio/kanister-tools:v9.99.9-dev
        command:
        - sh
        - -c
        - |
         sleep 1
    - func: KubeTask
      name: action0-task1
      args:
        namespace: "{{ .Namespace.Name }}"
        image: ghcr.io/kanisterio/kanister-tools:v9.99.9-dev
        command:
        - sh
        - -c
        - |
         sleep 1
    - func: KubeTask
      name: action0-task2
      args:
        namespace: "{{ .Namespace.Name }}"
        image: ghcr.io/kanisterio/kanister-tools:v9.99.9-dev
        command:
        - sh
[...]

PR #1577, opened 2022-07-29
Revert "Update Controller To Track Action Progress (#1487)"

pavannd1:
Reverts kanisterio/kanister#1487

Refer to #1576 

Reverting to allow downstream users to continue using Kanister on older K8s versions while we find a fix.


onkarbhat (approved):
LGTM

miquella (approved):

PR #1579, opened 2022-08-02
Update ubi-minimal base image to ubi-minimal:8.6-902

infraq:
Update ubi-minimal base image to ubi-minimal:8.6-902


pavannd1:
Duplicate

PR #1605, opened 2022-08-22
Add GCS bucket creation test coverage + Region method

KastenMike:
Adds test coverage for GCS bucket creation + Region method. This was 1 approach I was considering. The other is https://github.com/kanisterio/kanister/pull/1604

The main thing putting me off this approach is the strange Directory/Bucket intermingling & the fact I had to add a func on the Directory

### DIFF
--- a/pkg/objectstore/objectstore.go
+++ b/pkg/objectstore/objectstore.go
@@ -50,7 +50,6 @@ type Provider interface {
 // Bucket abstracts the object store of different cloud providers
 type Bucket interface {
 	Directory
-	GetRegion() string
 }
 
 // Directory operations
--- a/pkg/objectstore/objectstore.go
+++ b/pkg/objectstore/objectstore.go
@@ -50,6 +50,7 @@ type Provider interface {
 // Bucket abstracts the object store of different cloud providers
 type Bucket interface {
 	Directory
+	GetRegion() string
 }
 
 // Directory operations
@@ -90,6 +91,8 @@ type Directory interface {
 
 	// Serialize directory
 	String() string
+
+	GetRegion() string
 }
 
 // NewProvider creates a new Provider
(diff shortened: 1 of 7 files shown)


pkg/objectstore/objectstore.go
```diff
@@ -90,6 +90,8 @@ type Directory interface {
 
 	// Serialize directory
 	String() string
+
+	GetRegion() string
```
ihcsim, line 94: Can this be added to the `objectstore.Bucket` interface and let the `bucket` struct implements it?

pkg/objectstore/objectstore_test.go
```diff
@@ -471,6 +471,26 @@ func (s *ObjectStoreProviderSuite) TestBucketGetRegions(c *C) {
...
+	gcsRegions := []string{"US-WEST1", "EUROPE-WEST8", "ASIA-SOUTH2"}
+	ctx := context.Background()
+	for _, region := range gcsRegions {
+		s.initProvider(c, region)
+		bucketName := s.createBucketName(c)
+		_, err := s.provider.CreateBucket(ctx, bucketName)
+		c.Check(err, IsNil)
+		b, getErr := s.provider.GetBucket(ctx, bucketName)
+		err = s.provider.DeleteBucket(ctx, bucketName)
+		c.Check(getErr, IsNil)
```
ihcsim, line 488: Might wanna check `getErr` before attempting to delete the bucket.

KastenMike: But then in the rare chance that create passes & get fails then you don't attempt bucket cleanup. Defer is prohibited or not encouraged inside a for loop

pkg/objectstore/objectstore.go
KastenMike: I tried this initially but because of the weird relationship between bucket & directory it made me implement in directory also; then I realised I can remove the bucket one and have it only in directory

KastenMike: or I should say, iirc, because we return Directory interface rather than Bucket from some funcs...will need to look again Tomorrow to find the funcs causing the issues

pkg/objectstore/objectstore_test.go
ihcsim: I see what you mean. You can use `defer` inside a go routine or closure within the `for` loop. It's just a take-it-or-leave-it suggestion.

PR #1606, opened 2022-08-23
executing controller with custom config

AKhoria:
## Change Overview

Add an option to use custom config in controller(useful for local debug)

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kancontroller/kancontroller.go
+++ b/pkg/kancontroller/kancontroller.go
@@ -46,13 +46,24 @@ const (
 )
 
 func Execute() {
+	execute(nil)
+}
+
+func ExecuteWithCustomConfig(config *rest.Config) {
+	execute(config)
+}
+
+func execute(config *rest.Config) {
 	ctx := context.Background()
 	// Initialize the clients.
 	log.Print("Getting kubernetes context")
-	config, err := rest.InClusterConfig()
-	if err != nil {
-		log.WithError(err).Print("Failed to get k8s config")
-		return
+	if config == nil {
+		var err error
+		config, err = rest.InClusterConfig()
+		if err != nil {
+			log.WithError(err).Print("Failed to get k8s config")
+			return
+		}
 	}
 
 	// Run HTTPS webhook server if webhook certificates are mounted in the pod


pkg/kancontroller/kancontroller.go
```diff
@@ -46,13 +46,24 @@ const (
 )
 
 func Execute() {
+	execute(nil)
```
KastenMike, line 49: personally I would move the `config, err = rest.InClusterConfig()` into here and pass the result instead of nil rather than having nested if's in `execute`

viveksinghggits: the other option that we can consider is using the way kubectl or other tools do this.

ihcsim:
@AKhoria what problem are you trying to solve? Can we use the [`controller-runtime.GetConfig()`](https://pkg.go.dev/sigs.k8s.io/controller-runtime#pkg-variables) function, which creates a `rest.Config` from one of (in this order):

* --kubeconfig flag pointing at a file
* KUBECONFIG environment variable pointing at a file
* In-cluster config if running in cluster
* $HOME/.kube/config if exists.

pkg/kancontroller/kancontroller.go
ihcsim: The `k8s.io/cli-runtime/pkg/genericclioptions` will include [options](k8s.io/cli-runtime/pkg/genericclioptions) that are suitable for CLI, but not controller i think.

AKhoria:
> @AKhoria what problem are you trying to solve? Can we use the [`controller-runtime.GetConfig()`](https://pkg.go.dev/sigs.k8s.io/controller-runtime#pkg-variables) function, which creates a `rest.Config` from one of (in this order):
> 
> * --kubeconfig flag pointing at a file
> * KUBECONFIG environment variable pointing at a file
> * In-cluster config if running in cluster
> * $HOME/.kube/config if exists.

@ihcsim Sometimes it might be useful to run a controller locally for debugging purposes, from the code using kanister.

viveksinghggits:
> @AKhoria what problem are you trying to solve? Can we use the [`controller-runtime.GetConfig()`](https://pkg.go.dev/sigs.k8s.io/controller-runtime#pkg-variables) function, which creates a `rest.Config` from one of (in this order):
> 
> * --kubeconfig flag pointing at a file
> * KUBECONFIG environment variable pointing at a file
> * In-cluster config if running in cluster
> * $HOME/.kube/config if exists.

this is how I wanted to recommend this.

viveksinghggits:
cc @AKhoria

PR #1610, opened 2022-08-26
Update ubi-minimal base image to ubi-minimal:-

infraq:
Update ubi-minimal base image to ubi-minimal:-


viveksinghggits:
Is there something wrong, it looks like the tag was not formed correctly in the title and desc etc.

PrasadG193:
`8.6-902` is still the latest version of [ubi-minimal](https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8) image

PR #1611, opened 2022-08-30
Add workflow to build and consume kopia binary in kanister-tools

ankitjain235:
Signed-off-by: Ankit Jain <ankitjain.meone@gmail.com>

## Change Overview

This PR adds functionality to build and push kopia image to ghcr.io/kanisterio.
When a PR merges, we update the image.

Current commit corresponds to v0.11.3 commit upstream.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


Makefile
```diff
@@ -41,6 +41,10 @@ DOCKER_CONFIG ?= "$(HOME)/.docker"
 
 # Mention the vm-driver that should be used to install OpenShift
 vm-driver ?= "kvm"
+
+COMMIT_ID ?= "317cc36"
```
PrasadG193, line 45: why hardcoded? IMO the default value should be the latest commit at that time.

build/update_kopia_image.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Copyright 2021 The Kanister Authors.
```
PrasadG193, line 3:
```suggestion
# Copyright 2022 The Kanister Authors.
```

docker/tools/Dockerfile
```diff
@@ -10,8 +10,7 @@ LABEL name="kanister-tools" \
 
 COPY --from=restic/restic:0.11.0 /usr/bin/restic /usr/local/bin/restic
 # kastenhq/kopia alpine-sha-7664c8c image
-COPY --from=ghcr.io/kastenhq/kopia@sha256:540e05631dc48428725dc938ce62d7e4c1e698ba1fc6956ad36dd349ad076a20 \
-  /kopia/kopia /usr/local/bin/kopia
+COPY --from=ghcr.io/kanisterio/kopia:alpine-317cc36 /kopia/kopia /usr/local/bin/kopia
```
PrasadG193, line 13: Does one have to update and commit this file each time they want to update kopia binary in tools image?

Makefile
pavannd1: Is there a way we can find the commit ID from a release version? We shouldn't have the latest because there's no guarantee it will be stable.

ankitjain235: We can get commit ID from a release version but for the case now and few other cases as well, we don't necessarily want to pick the commit from released tag, so left it open to use any commit we want to pick.

docker/tools/Dockerfile
ankitjain235: Yes, since we want a specific version of the kopia binary in kanister-tools, we should always specify it here and update whenever needed.

ankitjain235:
Working on a follow-up PR to automate the process : https://github.com/kanisterio/kanister/pull/1620

.github/workflows/main.yaml
```diff
@@ -87,5 +87,6 @@ jobs:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
     - run: tar -xvf ./src.tar.gz
+    - run: make update-kopia-image
```
pavannd1, line 90: Is this going to build an image on each Kanister commit into master?

ankitjain235: It would build but since the sah256 and tag are same, image won't be updated.

viveksinghggits (approved):

build/update_kopia_image.sh
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+
+IMAGE_REGISTRY="ghcr.io/kanisterio"
+
+set -o errexit
+set -o nounset
```
PrasadG193, line 23: duplicate code. These options are already declared on line no 17 and 18

PrasadG193: Please set all the options in the header of the file

akankshakumari393 (approved):

docker/kopia-build/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR /
+
+RUN git clone https://github.com/${kopiaRepoOrg}/kopia.git
+WORKDIR /kopia
+
+RUN git checkout ${kopiaBuildCommit}
+
+# Build static executable
+RUN GO111MODULE=on CGO_ENABLED=0 GO_EXTLINK_ENABLED=0 go build \
+  -ldflags="-s -w -X github.com/kopia/kopia/repo.BuildVersion=${kopiaBuildCommit}" \
```
PrasadG193, line 17: @ankitjain235 since kopia users goreleaser, can't we use goreleaser command to build the binary?

ankitjain235: There also we would have to build and push the docker image only right? Since we are only consuming the kopia binary in kanister-tools, we can't publish it as an artifact as well, and just cloning and building the binary from a specific commit seems better through a Dockerfile instead of a script that would do that.

PrasadG193: You mean goreleaser config in Kopia doesn't create a binary?

ankitjain235: It does, but for building the binary from kanister, we would have to clone kopia before building, and that won't be possible unless we do this through a script right?

PrasadG193: I did not get that.
So in this dockerfile, we are cloning kopia github repo and executing `go build` to generate binary. My question is, instead of `go build` command, can't we use `goreleaser` in kopia to build kopia binary?

PrasadG193: that will reduce maintenance efforts to keep the build command used in goreleaser config in sync with the one we are using in Dockerfile.

docker/kopia-build/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+ENV GITHUB_REPOSITORY=https://github.com/${kopiaRepoOrg}/kopia
+
+WORKDIR /kopia
+
+# Build kopia binary from specific commit
+RUN git checkout ${kopiaBuildCommit} && \
+  wget https://github.com/goreleaser/goreleaser/releases/download/v1.11.2/goreleaser_1.11.2_x86_64.apk && \
+  apk add ./goreleaser_1.11.2_x86_64.apk --allow-untrusted && \
+  rm -f goreleaser_1.11.2_x86_64.apk && \
```
PrasadG193, line 20: @ankitjain235 we can copy kopia binary from the goreleaser docker image - https://hub.docker.com/r/goreleaser/goreleaser

.github/workflows/main.yaml
pavannd1: But do we build and push a new image with unchanged binary every time?

ankitjain235: Yes, for now I haven't handled such scenario to check for existing image with same commit.

docker/kopia-build/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+COPY --from=docker.io/goreleaser/goreleaser@sha256:b13418f20019fffb29797aff3d03f5e9ca84cea93634f4169a7ed603a95ab198 \
+  /usr/bin/goreleaser /usr/bin/goreleaser
+RUN goreleaser build --output=. --rm-dist --single-target
+
+RUN adduser -D kopia && addgroup kopia kopia
+USER kopia:kopia
+
+COPY --chown=kopia . /kopia
+
+FROM alpine:3.14
```
pavannd1, line 28: Why not 3.16?

docker/kopia-build/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM golang:1.17-alpine3.14 AS builder
```
pavannd1, line 1: Looks like the new image is available now - [golang:1.19.1-alpine3.16](https://hub.docker.com/layers/library/golang/1.19.1-alpine3.16/images/sha256-8074e76a21f9e52ceec00387bdf357cc1caac37f1e31f1e312dc103e9f5e45d3?context=explore)

docker/kopia-build/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN git clone https://github.com/${kopiaRepoOrg}/kopia.git
+
+ENV GITHUB_REPOSITORY=https://github.com/${kopiaRepoOrg}/kopia
+
+WORKDIR /kopia
+
+# Build kopia binary from specific commit
+RUN git checkout ${kopiaBuildCommit}
+# Refers to goreleaser/goreleaser:v1.11.2
+COPY --from=docker.io/goreleaser/goreleaser@sha256:b13418f20019fffb29797aff3d03f5e9ca84cea93634f4169a7ed603a95ab198 \
```
pavannd1, line 19: Is this the same goreleaser version used during release?

.github/workflows/main.yaml
pavannd1: As discussed offline, as a follow-up, we should make this a manual workflow. This should be updated only when we bump the commit ID.

docker/kopia-build/Dockerfile
ankitjain235: In kanister build, we are using `goreleaser/goreleaser:v1.2.4`. It's about 8 months old when we last updated it.

pavannd1 (approved):

docker/kopia-build/Dockerfile
julio-lopez: @ankitjain235 (@pavannd1) Can this be simplified?

For example, does `goreleaser` need to be used here?

It is possible to build kopia directly. For example, see https://github.com/kastenhq/kopia/blob/publish/docker/kopia-alpine/Dockerfile

A few questions:
- is is necessary to build a container image for kopia? or would it suffice to build the binary, say as part of building the tools image?
- is it necessary to build the image or binary every time? or can it be cached and reused, and only rebuilt when the kopia version / commit id changes?

julio-lopez (commented):
🥇Thanks for doing this.

Couple of questions to iterate and refine

docker/kopia-build/Dockerfile
ankitjain235: @julio-lopez 

We used goreleaser just so that we can build the kopia binary the way it is done upstream and we won't have to worry about changing the command if it needs to be changed any time as per change upstream.

In a follow-up PR(https://github.com/kanisterio/kanister/pull/1620) that is WIP, I am adding an action that can be manually triggered whenever we want to build kopia. I kept the process to build a container image for kopia only for this use case. That way, we would only build Kopia image when it's required. Making it as a part of kanister-tools image would result in re-building the binary in scenarios I tried out.

pavannd1:
@julio-lopez good questions 👍🏼 We have discussed those improvements and will iterate on this. It was crucial to get this in to unblock the upgrades 💯

PR #1615, opened 2022-08-31
Add azure managed identity to azure package in blockstorage

leuyentran:
## Change Overview

This PR add : 

1.  Support for Azure Manage Identity by adding MSI authorizer
2. Add an Authenticate() interface to check validity of Azure creds.  Azure Managed Identity is  the first type to be supported for this interface .  Similar to other Azure interfaces, this interface is not united tested because creating a new Azure provider involves calling `Instance Azure Metadata` which will results in `context timeout `
4. Add utilities functions to provide appropriate abstractions levels for point 1 and 2 + unit tests

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/blockstorage/azure/azuredisk.go
```diff
@@ -44,6 +44,22 @@ func (s *AdStorage) Type() blockstorage.Type {
 	return blockstorage.TypeAD
 }
 
+// Authenticate check Azure creds if the credType is supported
+func (s *AdStorage) Authenticate(ctx context.Context, credType string, creds map[string]string) (CredsValidity, error) {
```
leuyentran, line 48: This interface is not united tested because creating a new Azure provider involves calling `Azure Instance Metadata` which will results in `context timeout`

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,84 @@
...
+
+// internal interface to authenticate with different Azure credentials type
+type authenticator interface {
+	authenticate(creds map[string]string) (CredsValidity, error)
+}
+
+// authenticate with MSI creds
+type msiAuthenticator struct{}
+
+func (m *msiAuthenticator) authenticate(creds map[string]string) (CredsValidity, error) {
```
leuyentran, line 65: This interface is not united tested because it involves network calls which will results in `context timeout`

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+// currently avaialble types: https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization
+// to be available with azidentity: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-credential-types
+// to faciliate future migration, only the common ones between two SDKs are listed
+const (
+	CredTypeManagedIdentity   string = "cred-type-msi"
+	CredTypeClientSecret      string = "cred-type-client-secret"
+	CredTypeClientCertificate string = "cred-type-client-certificate"
+)
+
+var availableCredsType = sets.NewString(CredTypeManagedIdentity)
```
bathina2, line 22: Is this setting MSI as default?

bathina2: Not sure why a set is being used. Where are we building the set with available types?

pkg/blockstorage/azure/client.go
```diff
@@ -98,15 +98,37 @@ func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
...
+	ba, ok := a.(*autorest.BearerAuthorizer)
+	if !ok {
+		return nil, errors.New("Failed to get Azure authorizer")
+	}
+	return ba, nil
+}
 
+func getMSIsAuthorizer(config map[string]string) (*autorest.BearerAuthorizer, error) {
+	msiConfig := auth.NewMSIConfig()
+	msiConfig.ClientID = config[blockstorage.AzureTenantID]
```
onkarbhat, line 127: Is there a typo here ? Should this be using blockstorage.AzureCientID instead of blockstorage.AzureTenantID ?

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func isClientCredsAvailable(config map[string]string) bool {
+	if config[blockstorage.AzureTenantID] != "" &&
+		config[blockstorage.AzureCientID] != "" &&
+		config[blockstorage.AzureClentSecret] != "" {
+		return true
+	} else {
+		return false
+	}
```
onkarbhat, line 44: Lets simplify this to 

`return  config[blockstorage.AzureTenantID]!= "" && config[blockstorage.AzureCientID] != "" && config[blockstorage.AzureClentSecret] != ""`

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func isMSICredsAvailable(config map[string]string) bool {
+	if config[blockstorage.AzureTenantID] == "" &&
+		config[blockstorage.AzureCientID] != "" &&
+		config[blockstorage.AzureClentSecret] == "" {
+		return true
+	} else {
+		return false
+	}
```
onkarbhat, line 54: Similar to the comment above lets simplify this to 

`return config[blockstorage.AzureTenantID] == "" && config[blockstorage.AzureClientID] != " && config[blockstorage.AzureClientSecret] == "" `

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+		config[blockstorage.AzureClentSecret] != "" {
+		return true
+	} else {
+		return false
+	}
+}
+
+func isMSICredsAvailable(config map[string]string) bool {
+	if config[blockstorage.AzureTenantID] == "" &&
+		config[blockstorage.AzureCientID] != "" &&
```
onkarbhat, line 49: typo AzureCientID -> AzureClientID

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+		config[blockstorage.AzureClentSecret] == "" {
```
onkarbhat, line 50: typo AzureClentSecret -> AzureClientSecret

pkg/blockstorage/azure/client.go
```diff
@@ -98,15 +98,37 @@ func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
 
 // nolint:unparam
 func getAuthorizer(env azure.Environment, config map[string]string) (*autorest.BearerAuthorizer, error) {
+	if isClientCredsAvailable(config) {
+		return getClientCredsAuthorizer(env, config)
+	} else if isMSICredsAvailable(config) {
+		return getMSIsAuthorizer(config)
+	}
+	return &autorest.BearerAuthorizer{}, errors.New("Missing credentials, or credential type not supported")
```
onkarbhat, line 106: Please return `nil, errors.New.....` .

pkg/blockstorage/azure/auth.go
onkarbhat, line 22: Why does this have only the type for ManagedIdentity ?  I noticed that you have a function named getClientCredsAuthorizer which gives the impression that we do support something other than ManagedIdentity. But then  this function isCredTypeSupported only returns true if the type is CredTypeManagedIdentity .

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// currently avaialble types: https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization
+// to be available with azidentity: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-credential-types
+// to faciliate future migration, only the common ones between two SDKs are listed
+const (
+	CredTypeManagedIdentity   string = "cred-type-msi"
+	CredTypeClientSecret      string = "cred-type-client-secret"
+	CredTypeClientCertificate string = "cred-type-client-certificate"
```
onkarbhat, line 19: I dont see these two constants used in this file. Lets not include it if they are not used.

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+		config[blockstorage.AzureClentSecret] == "" {
+		return true
+	} else {
+		return false
+	}
+}
+
+// internal interface to authenticate with different Azure credentials type
+type authenticator interface {
+	authenticate(creds map[string]string) (CredsValidity, error)
```
onkarbhat, line 59: Curious why you need another field of type CredsValidity when the error would have been sufficient ? My recommendation is to create new errors and export them in case the caller wants to look for specific errors/conditions.  With the current model of using CredsValidity, the caller would have to check and use both the validity field and the error field. It is simpler to have the caller deal with only an error field. 
```
var ErrMSIUnsupported = errors.New("MSI Endpoint is not supported")
var ErrServicePrincipalTokenCreateFailed = ...
var ErrRefreshTokenFailed = ....

```

and use them in msIAuthenticator's authenticate . A nil error would indicate that authentication was successful. Only when error is not nil, the caller could check the exported errors from this package.

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,94 @@
...
+type msiAuthenticator struct{}
+
+func (m *msiAuthenticator) authenticate(creds map[string]string) (CredsValidity, error) {
+	// check if MSI endpoint is available
+	if !adal.MSIAvailable(context.Background(), nil) {
+		return CredsTypeNotSupported, errors.New("MSI endpoint is not supported")
+	}
+	// create a service principal token
+	msiConfig := auth.NewMSIConfig()
+	msiConfig.ClientID = creds[blockstorage.AzureTenantID]
```
onkarbhat, line 72: Is there a typo here ? Should we use blockstorage.ClientID instead of tenant ID ?

pkg/blockstorage/azure/auth.go
leuyentran: It's a typo from source: https://github.com/kanisterio/kanister/blob/71b59c968301913f440cb81950fd3966938be310/pkg/blockstorage/helpers.go#L33

pkg/blockstorage/azure/auth.go
leuyentran: It's a typo from source: 
https://github.com/kanisterio/kanister/blob/71b59c968301913f440cb81950fd3966938be310/pkg/blockstorage/helpers.go#L34

pkg/blockstorage/azure/auth.go
leuyentran: Indeed. Thank you

pkg/blockstorage/azure/client.go
leuyentran: indeed. Thank you

pkg/blockstorage/azure/auth.go
leuyentran: I can add a refactor commit

pkg/blockstorage/azure/auth.go
leuyentran: I can add a refactor commit

pkg/blockstorage/azure/auth.go
leuyentran: I am not able to think of a design at the moment with other priorities. Between the caller's ability to look for specific errors/conditions and returning errors from Azure APIs, I think it makes sense to prioritize the latter to give end-user precise information on failures. I will : 
1. Remove  `CredsValidity`
2. Keep error returning as is 
Therefore, a successful authentication from the caller side is `error == nil `, any error otherwise will be piped back up the stack

What do you think @onkarbhat

pkg/blockstorage/azure/auth.go
leuyentran: Because the interface **Authenticated()** only supported ManagedIdentity at that point and that was PR's intention. However, that would be too confusing and thus I had added `ClientCredsAuthorizer`

leuyentran: It does not set MSI as default, just a list of 1 since `Authenticated()` was only added for MSI. However I think it's confusing to only have MSI and not ClientCreds, so ClientCreds is added:
```go
var availableCredsType = sets.NewString(CredTypeManagedIdentity, CredTypeClientSecret)
```

a `set` is used so I can use an interface such as `Has()` to quickly search for supported credential types. When more credentials type is added in the future, only the `set` initialization needs to be extended

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,121 @@
+package azure
+
+import (
+	"context"
+
+	"github.com/Azure/go-autorest/autorest/adal"
```
ihcsim, line 6: Looks like this package will be [deprecated next year](https://pkg.go.dev/github.com/Azure/go-autorest/autorest/adal@v0.9.21#section-readme). Can we look into using [`azidentity`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity) instead?

leuyentran: The purpose of interface `Authenticate()` is to validate Azure creds as a one-of action. These creds are then passed on to an `Authorizer()` that later authenticates and return a token for every Azure API calls that interact with Azure resources. 
And for `Authorizer()`, we are still using `adal` indirectly as part of  `github.com/Azure/go-autorest/autorest/azure/auth` 

Furthermore, `azidentity` and `adal` differs slightly in their credential definitions and workflows. So I'd prefer not to use `azidentity` for `Authenticate()` and `adal` for `Authorizer()` in case we don't completely understand the nuances in expectations.  

Instead, I had abstracted out the logics as much as I can, which should make APIs swapping easier during migration :)

ihcsim (commented):
Code change LGTM. I will leave it to @onkarbhat and @bathina2 to approve.

bathina2:
The changes look good. However, I do have some design opinions. From the caller side I would expect something like - 
```
authenticator, err := GetAuthenticator
if err != nil {
    //spit out err
}
err = authenticator.Authenticate()
```
Let me know if you have time to discuss tomorrow

pkg/blockstorage/azure/auth_tests.go
```diff
@@ -0,0 +1,74 @@
...
+type AuthSuite struct{}
+
+var _ = Suite(&AuthSuite{})
+
+func (s *AuthSuite) SetUpSuite(c *C) {
+}
+
+func (s *AuthSuite) TestIsCredTypeSupported(c *C) {
+	credType := "cred-type-msi"
+	c.Assert(isCredTypeSupported(credType), Equals, true)
```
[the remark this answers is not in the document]
leuyentran: This test is meant for testing the following definition and function. The definition of supported is if credential type is part of `availableCredsType` - which can be extended in the future

Test for `CredTypeClientSecret` is added 
```go
var availableCredsType = sets.NewString(CredTypeManagedIdentity, CredTypeClientSecret)

func isCredTypeSupported(credType string) bool {
	return availableCredsType.Has(credType)
}
```

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Failed to create a service principal token")
+	}
+	// network call to check for token
+	err = spt.Refresh()
+	if err != nil {
+		return errors.Wrap(err, "Failed to refresh token")
+	}
+	// creds passed authentication
+	return nil
```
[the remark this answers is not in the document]
leuyentran: The caller, i.e. `Authenticate()` directly return results of these methods. Therefore,  the explicitly here is meant for code readability for future developers to understand: `err==nil` is considered passing authentication.

pkg/blockstorage/azure/auth_tests.go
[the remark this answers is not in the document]
leuyentran: After discussion with @bathina2 , this function is now removed

pkg/blockstorage/azure/auth.go
leuyentran: https://github.com/kanisterio/kanister/issues/1621

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,108 @@
...
+		return &msiAuthenticator{}, nil
+	case isClientCredsAvailable(config):
+		return &clientSecretAuthenticator{}, nil
+	default:
+		return nil, errors.New("Fail to get an authenticator for provided creds combination")
+	}
+}
+
+// authenticate with MSI creds
+type msiAuthenticator struct{}
```
bathina2, line 45: nit: consider renaming to msiAzureAuthenticator

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,108 @@
...
+	err = spt.Refresh()
+	if err != nil {
+		return errors.Wrap(err, "Failed to refresh token")
+	}
+	// creds passed authentication
+	return nil
+}
+
+// authenticate with client secret creds
+type clientSecretAuthenticator struct{}
```
bathina2, line 69: nit: consider renaming to clientSecretAzureAuthenitcator.

bathina2 (approved):
Looks good, Minor nits about naming. Might be good to be more explicit, but not a big deal

pkg/blockstorage/azure/auth.go
leuyentran: Since the package name is `azure' , I think it is okay to leave the naming as is. If they need to be public for some reason in the future, it would still be `azure.ClientSecretAuthenticator`

pkg/blockstorage/azure/auth.go
onkarbhat: Sounds good to me. Thank you Le

pkg/blockstorage/azure/auth.go
```diff
@@ -0,0 +1,108 @@
...
+		config[blockstorage.AzureCientID] != "" &&
+		config[blockstorage.AzureClentSecret] == "")
+}
+
+// Public interface to authenticate with different Azure credentials type
+type AzureAuthenticator interface {
+	Authenticate(creds map[string]string) error
+}
+
+func NewAzureAutheticator(config map[string]string) (AzureAuthenticator, error) {
```
onkarbhat, line 33: Please add comments for exported methods/structs in this PR.

onkarbhat (commented):
LGTM . Posting it after merge.

PR #1636, opened 2022-09-13
Update ubi-minimal base image to ubi-minimal:8.6-941

infraq:
Update ubi-minimal base image to ubi-minimal:8.6-941


pavannd1 (approved):
LGTM

PR #1633, opened 2022-09-12
Add ability to configure MysqlDB app via Options

alexvbg:
## Change Overview

This PR add ability to configure MysqlDB app via options + added function that help us configure chart values for that app.

## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [x] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

PR #1637, opened 2022-09-14
Add ability to modify chart in MongoDB and MysqlDB structs

alexvbg:
## Change Overview

- added interface HelmApp that helps us to extend current App chart configuration;
- HelmApp implemented in MysqlDB and MongoDB.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/app/mongodb.go
+++ b/pkg/app/mongodb.go
@@ -41,6 +41,8 @@ type IsMasterOutput struct {
 	Ismaster bool `json:"ismaster"`
 }
 
+var _ HelmApp = &MongoDB{}
+
 type MongoDB struct {
 	cli       kubernetes.Interface
 	namespace string
@@ -49,19 +51,9 @@ type MongoDB struct {
 	chart     helm.ChartInfo
 }
 
-type MongoDBOptions func(*MongoDB)
-
-func MongoDBChartValues(values map[string]string) MongoDBOptions {
-	return func(db *MongoDB) {
-		for k, v := range values {
-			db.chart.Values[k] = v
-		}
-	}
-}
-
 // Last tested working version "9.0.0"
-func NewMongoDB(name string, opts ...MongoDBOptions) App {
-	m := MongoDB{
+func NewMongoDB(name string) HelmApp {
+	return &MongoDB{
 		username: "root",
 		name:     name,
 		chart: helm.ChartInfo{
@@ -78,10 +70,14 @@ func NewMongoDB(name string, opts ...MongoDBOptions) App {
 			},
 		},
 	}
-	for _, opt := range opts {
-		opt(&m)
-	}
-	return &m
+}
+
+func (mongo *MongoDB) Chart() *helm.ChartInfo {
+	return &mongo.chart
+}
+
+func (mongo *MongoDB) SetChart(chart helm.ChartInfo) {
+	mongo.chart = chart
 }
 
 func (mongo *MongoDB) Init(ctx context.Context) error {
--- a/pkg/app/mongodb.go
+++ b/pkg/app/mongodb.go
@@ -49,9 +49,19 @@ type MongoDB struct {
 	chart     helm.ChartInfo
 }
 
+type MongoDBOptions func(*MongoDB)
+
+func MongoDBChartValues(values map[string]string) MongoDBOptions {
+	return func(db *MongoDB) {
+		for k, v := range values {
+			db.chart.Values[k] = v
+		}
+	}
+}
+
 // Last tested working version "9.0.0"
-func NewMongoDB(name string) App {
-	return &MongoDB{
+func NewMongoDB(name string, opts ...MongoDBOptions) App {
+	m := MongoDB{
 		username: "root",
 		name:     name,
 		chart: helm.ChartInfo{
@@ -68,6 +78,10 @@ func NewMongoDB(name string) App {
 			},
 		},
 	}
+	for _, opt := range opts {
+		opt(&m)
+	}
+	return &m
 }
 
 func (mongo *MongoDB) Init(ctx context.Context) error {
--- a/pkg/app/helm_app.go
+++ b/pkg/app/helm_app.go
@@ -1,0 +1,11 @@
+package app
+
+import "github.com/kanisterio/kanister/pkg/helm"
+
+type HelmApp interface {
+	App
+	// Chart returns the chart of this Helm app.
+	Chart() *helm.ChartInfo
+	// SetChart sets the chart of this Helm app.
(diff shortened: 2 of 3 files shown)


pkg/app/mongodb.go
```diff
@@ -68,6 +78,10 @@ func NewMongoDB(name string) App {
 			},
 		},
 	}
+	for _, opt := range opts {
+		opt(&m)
+	}
+	return &m
```
ihcsim, line 84: Any reasons why you can't just passed in the map of values directly? What do we gain by introducing `MonggoDBOptions`? Same question for the MySQL changes in #1633. E.g.,

```diff
diff --git a/pkg/app/mongodb.go b/pkg/app/mongodb.go
index 627e50fa..4dc14b70 100644
--- a/pkg/app/mongodb.go
+++ b/pkg/app/mongodb.go
@@ -49,18 +49,8 @@ type MongoDB struct {
 	chart     helm.ChartInfo
 }
 
-type MongoDBOptions func(*MongoDB)
-
-func MongoDBChartValues(values map[string]string) MongoDBOptions {
-	return func(db *MongoDB) {
-		for k, v := range values {
-			db.chart.Values[k] = v
-		}
-	}
-}
-
 // Last tested working version "9.0.0"
-func NewMongoDB(name string, opts ...MongoDBOptions) App {
+func NewMongoDB(name string, values map[string]string) App {
 	m := MongoDB{
 		username: "root",
 		name:     name,
@@ -78,8 +68,8 @@ func NewMongoDB(name string, opts ...MongoDBOptions) App {
 			},
 		},
 	}
-	for _, opt := range opts {
-		opt(&m)
+	for k, v := range values {
+		m.chart.Values[k] = v
 	}
 	return &m
 }
```

alexvbg: Hi! The reason is, I don't wanna change contract for constructor, so `MongoDB` struct can be created both by `NewMongoDB("mongo")` or `NewMongoDB("mongo", SomeOptionalFunc)`. Also, there is can be not only chart configuration, for example, if we need to change chart version we can create optional func for this.

ihcsim: You can make `values` an optional argument, just like what you did with `opts`. 

If there are other chart configuration to be changed, can we come up with a more generic way to do it to benefit other apps too? The current approach requires future changes to introduce one new `Option` type and multiple functions, to other apps. 

WDYT of this approach of making the app expose its `chart` so that you have more control over what to do with it in the client code? Will it do what you want?


```diff
diff --git a/pkg/app/app.go b/pkg/app/app.go
index 57c8cab3..5484d514 100644
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	crv1alpha1 "github.com/kanisterio/kanister/pkg/apis/cr/v1alpha1"
+	"github.com/kanisterio/kanister/pkg/helm"
 )
 
 // App represents an application we can install into a namespace.
@@ -43,6 +44,16 @@ type App interface {
 	GetClusterScopedResources(context.Context) []crv1alpha1.ObjectReference
 }
 
+type HelmApp interface {
+	App
+
+	// Chart returns the chart of this Helm app.
+	Chart() *helm.ChartInfo
+
+	// SetChart sets the chart of this Helm app.
+	SetChart(chart helm.ChartInfo)
+}
+
 // DatabaseApp inherits methods from App, but also includes method to add, read
 // and remove entries stored byt the App.
 type DatabaseApp interface {
diff --git a/pkg/app/mongodb.go b/pkg/app/mongodb.go
index 627e50fa..a592de1f 100644
--- a/pkg/app/mongodb.go
+++ b/pkg/app/mongodb.go
@@ -41,6 +41,8 @@ type IsMasterOutput struct {
 	Ismaster bool `json:"ismaster"`
 }
 
+var _ HelmApp = &MongoDB{}
+
 type MongoDB struct {
 	cli       kubernetes.Interface
 	namespace string
@@ -49,18 +51,8 @@ type MongoDB struct {
 	chart     helm.ChartInfo
 }
 
-type MongoDBOptions func(*MongoDB)
-
-func MongoDBChartValues(values map[string]string) MongoDBOptions {
-	return func(db *MongoDB) {
-		for k, v := range values {
-			db.chart.Values[k] = v
-		}
-	}
-}
-
 // Last tested working version "9.0.0"
-func NewMongoDB(name string, opts ...MongoDBOptions) App {
+func NewMongoDB(name string) HelmApp {
 	m := MongoDB{
 		username: "root",
 		name:     name,
@@ -78,12 +70,17 @@ func NewMongoDB(name string, opts ...MongoDBOptions) App {
 			},
 		},
 	}
-	for _, opt := range opts {
-		opt(&m)
-	}
 	return &m
 }
 
+func (mongo *MongoDB) Chart() *helm.ChartInfo {
+	return &mongo.chart
+}
+
+func (mongo *MongoDB) SetChart(chart helm.ChartInfo) {
+	mongo.chart = chart
+}
+
 func (mongo *MongoDB) Init(ctx context.Context) error {
 	// Instantiate Client SDKs
 	cfg, err := kube.LoadConfig()
```

alexvbg: My concern here is that there is some dependencies with App interface in k10. Yes, sure, we can do it more generic way, but it looks like for my purpose it will only be need to have changes in mysql and mongo only, so i don't want to overengineer here. But you are right, it good to comply with the DRY. If we gonna change App interface, we have to retest all things related to this in k10.

ihcsim: We are not changing the `App` interface - just changing Mongo and MySQL to a `HelmApp` which is still an `App`. As far as k10 is concerned, they should still work like an `App`, but now you should have access to the chart struct on K10 side. If down the road, others follow the same pattern in this PR, there will be more code for the maintainers to maintain.

alexvbg: Oh, yea, I got it. Ok, will change it this way. Thank you for review!

viveksinghggits: thank you Ivan.

alexvbg: Would you mind if I do the same for mysql in this PR? Or its better in different?

alexvbg: I really like how it looks now, pushed it, fixed the description.

ihcsim (approved):
LGTM :+1: - thanks for the change.

Issue #1342, opened 2022-03-30
Auto-generate CRD Documentation


ihcsim:
Use a tool like https://github.com/ahmetb/gen-crd-api-reference-docs/ to auto-generate the [CRD documentation](https://docs.kanister.io/architecture.html#custom-resources), to avoid deviation between the code and the doc. Update the Makefile `docs` target to include the new command.

ihcsim:
The styling of the current rst docs isn't compatible with that generated by the api-reference-docs tool. To be revisit after we move to GitHub Pages.

Kartik-Garg:
@ihcsim I am willing to contribute, can I take it up?

PrasadG193:
Hey @Kartik-Garg,
Thanks for showing interest. Please feel free to reach out to us if you need help.

Kartik-Garg:
Hey @PrasadG193 , I have created the PR : https://github.com/kanisterio/kanister/pull/1720
regarding this issue, can you please have a look at it? 
thank you!

Kartik-Garg:
Can we get this issue reopened please, work on PR is completed and it's updated

PR #1639, opened 2022-09-15
Add Design Doc For Versioning Functions

ihcsim:
## Change Overview

This PR adds a design document that proposes a versioning scheme to manage different versions of a Kanister Function. This version management scheme will help to ensure a smooth rollout and upgrade path for the upcoming Kopia integration work, where some Kanister Functions will have different versions of implementation based on different data moving mechanism.

Specifically, these Kanister Functions needs to retain its original `restic` implementation to ensure short term backward compatibility, while giving users a reasonable time window to upgrade their operational workflows to work with the Kopia server-based approach.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/design/func-version.md
+++ b/design/func-version.md
@@ -87,10 +87,14 @@ func init() {
 var  _ kanister.VersionedFunc = (*kubeExecFunc)(nil)
 
 func (*kubeExecFunc) StableVersion() string {
+  // to be real from functions.yaml config file
+  // hard-coded for prototype purposes
   return "v0.1.0"
 }
 
-func (*kubeExecFunc) Versions() map[string]kanister.Func {
+func (*kubeExecFunc) Versions() (map[string]kanister.Func, error) {
+  // to be real from functions.yaml config file
+  // hard-coded for prototype purposes
   return map[string]kanister.Func{
     kanister.DefaultVersion: &kubeExecFunc{},
     "v0.1.0": &kubeExecFuncV010{
@@ -351,15 +355,41 @@ posterity.
 
 ### Versions Discovery
 
-The versions of all Kanister Functions can be stored in a version YAML file
-that maps the Function name to the supported versions. Special annotation will
-be used to denote all stable versions. This file can be [embedded][4] in
-the Kanister controller binary during build. The controller uses this file to
-determine the stable version of each Function, whenever it needs that
-information.
+The versions of all Kanister Functions can be stored in a YAML file that maps
+the Function name to the supported versions, allowing users and maintainers to
+easily browse the code repository to determine all supported function versions
+of a particular release.
 
-To facilitate easy version discovery, the controller exposes an endpoint that
-publishes its supported Kanister Functions versions.
+Special annotations will be used to denote available, stable and deprecated
+versions of all functions. This file can be [embedded][4] in the Kanister
+controller binary during build. Each Function reads its version information
+from this file during the initial Functions registration process, and the later
+runtime phase execution.
+
+This is an example of what the version YAML file can look like:
+
+```yaml
+functions:
+  kubeExec:
+    versions:
+      served:
+      - 0.1.0
+      - 0.2.0
+      stable: 0.1.0
+      deprecated:
+      - 0.0.0
+  kubeTask:
+    versions:
+      served:
+      - 0.0.0
+      - 0.1.0
+      - 0.1.1
+      - 0.1.2
+      stable: 0.1.2
+```
+
+To facilitate easy version discovery, the controller can expose an HTTP endpoint
+that publishes its supported Kanister Functions versions.
 
 ## Test Cases
 
--- a/design/func-version.md
+++ b/design/func-version.md
@@ -1,0 +1,364 @@
+# Kanister Functions Version Management
+
+<!-- toc -->
+<!-- /toc -->
+
+## Problem Statement
+
+As Kanister continues to grow and evolve to meet the needs of its community,
+it's inevitable that eventually, some of the existing built-in [Kanister
+Functions][0] will be updated with breaking changes, deprecated or even
+removed entirely.
+
+The lack of a proper strategy to manage and rollout such changes in a
+responsible backward compatible manner can lead to costly operational overhead
+for Kanister users. In addition, it also becomes difficult for Kanister
+maintainers to assess the blast radius of the change, determine the rollout
+timing, provide timely support etc.
+
+This proposal outlines a version management scheme that can be used to manage
+the lifecycle of Kanister Functions as they evolve from inception to retirement.
+
+## Proposed Solution
+
+### Summary
+
+The proposed solution uses a version management scheme to communicate the
+stability, availability, promotion and deprecation of Kanister Functions.
+
+The design extends the existing implementation of the action-level
+[`PreferredVersion` mechanism][1] to the phase-level by enabling users to
+specify the desired  Kanister Function version within the blueprint. It also
+proposes changes to the relevant interfaces with code examples demonstrating the
+implementation of multiple versions of a Kanister Function.
+
+### Assumptions And Constraints
+
+* The proposed solution is compatible with existing downstream custom Kanister
+Function implementation, without any code change.
+
+### Function Versioning
+
+The proposed solution introduces a new Go interface named `VersionedFunc`. The
+`VersionedFunc` type embeds the `Func` type to include information about the
+multiple versions of implementation of the underlying Kanister Function.
+
+```go
+// VersionedFunc extends the Func type to include information about the multiple
+// versions of the underlying Kanister Function.
+type VersionedFunc interface {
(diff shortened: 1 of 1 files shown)


shlokchaudhari9 (commented):
In a future scenario where Kanister maintainers (or users) implement multiple versions of BackupData and RestoreData Functions, we should think of a mechanism to avoid version conflicts while running such dependent Functions from a Blueprint.

For example; if user runs an action consisting of BackupData v0.0.0 and then tries to restore using RestoreData v0.1.0 that would certainly lead to failures because the backend backup-restore workflow will be inconsistent between versions.

We may be able to handle this to some extent by;
1. adding version validation checks for such type of Functions, or
2. documenting example Blueprints to depict correct usage

But, we should definitely discuss a good solution for such cases.

design/func-version.md
```diff
@@ -0,0 +1,378 @@
...
+  // Func is the v0.0.0 version of the function. For existing functions, this
+  // maps to their existing `Func` implementation.
+  Func
+
+  // StableVersion returns the stable version of the function.
+  StableVersion() string
+
+  // Versions maps the different versions of the function to their respective
+  // implementations.
+  Versions() map[string]Func
```
viveksinghggits, line 69: We obviously need this, but I am not sure this function should be part of this interface. More comments below here this is implemented.

design/func-version.md
```diff
@@ -0,0 +1,378 @@
...
+  return map[string]kanister.Func{
+    kanister.DefaultVersion: &kubeExecFunc{},
+    "v0.1.0": &kubeExecFuncV010{
+      parent: &kubeExecFunc{},
+    },
+    "v0.2.0": &kubeExecFuncV020{
+      parent: &kubeExecFunc{},
+    },
+  }
+}
```
viveksinghggits, line 103: It looks like this implementation depends on the values that are there in the map where we register functions. That is the reason I was thinking if having the method as behaviour of functions would be a good idea.
Also, if my understanding is right, should we not read the values from that map, instead of just had coding here.

viveksinghggits: or maybe I am not able to understand the exact use case of this method.

design/func-version.md
```diff
@@ -0,0 +1,378 @@
...
+error and halt the execution. The version number will still be preserved for
+posterity.
+
+### Versions Discovery
+
+The versions of all Kanister Functions can be stored in a version YAML file
+that maps the Function name to the supported versions. Special annotation will
+be used to denote all stable versions. This file can be [embedded][4] in
+the Kanister controller binary during build. The controller uses this file to
+determine the stable version of each Function, whenever it needs that
```
viveksinghggits, line 358: Controller can also call `StableVersion()` method from the interface to get the stable version right?

design/func-version.md
```diff
@@ -0,0 +1,378 @@
...
+information.
+
+To facilitate easy version discovery, the controller exposes an endpoint that
+publishes its supported Kanister Functions versions.
+
+## Test Cases
```
viveksinghggits, line 364: Should we also include the documentation changes? If multiple versions of the functions have different arguments, how the document is going to look like?

design/func-version.md
```diff
@@ -0,0 +1,378 @@
...
+the lifecycle of Kanister Functions as they evolve from inception to retirement.
+
+## Proposed Solution
+
+### Summary
+
+The proposed solution uses a version management scheme to communicate the
+stability, availability, promotion and deprecation of Kanister Functions.
+
+The design extends the existing implementation of the action-level
```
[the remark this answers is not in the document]
ihcsim: Function-level settings take precedence.

design/func-version.md
ihcsim: Yes, all supported versions will be included in the documentation. Similar to https://github.com/kanisterio/kanister/issues/1342, we should use a script to automate the functions doc generation.

design/func-version.md
ihcsim: If possible, I don't want the new caller code to directly depend on the global map going forward. Sorry for the confusion - I need to fix up this part. I imagine the actual implementation of `StableVersion()` and `Versions()` would involve returning the versions information read from the in-memory config file discussed below.

design/func-version.md
ihcsim: `StableVersion()` would read from this file.

design/func-version.md
ihcsim: So during `init()`, the [global registration function](https://github.com/kanisterio/kanister/pull/1639/files#diff-8a18494585b5ed63b67bb4b0ed2e7cee05065091750638787ad5bd7b57213fe6R160-R169) asks a Kanister Function for all its versions, via the `Versions()` method.

The `Versions()` method gets the versions info from the [`embed`ded](https://pkg.go.dev/embed) in-memory config file, which looks something like:

```yaml
functions:
  kubeExec:
    versions:
      served:
      - 0.1.0
      - 0.2.0
      stable: 0.1.0
      deprecated:
      - 0.0.0
  kubeTask:
    versions:
      served:
      - 0.0.0
      - 0.1.0
      - 0.1.1
      - 0.1.2
      stable: 0.1.2
```

During phase execution, we will follow the [existing version detection logic](https://github.com/kanisterio/kanister/blob/8b5eb2026220d4ef9c1c6d0ba07c395acec2c784/pkg/phase.go#L123-L148) to find the requested version from the `funcs` global map. The `StableVersion()` function let us to deduce the default version, in case the blueprint doesn't specify a function version. We can also add a `IsDeprecated(version string) bool` string to the `kanister.VersionedFunc` interface to determine if a requested version is deprecated.

WDYT?

design/func-version.md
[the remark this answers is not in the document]
viveksinghggits: I somehow think that we should try to simplify this, if not as part of this PR then maybe as part of other PR. Because there seem to be a lot of options now using which we can specify the function version which might become confusing to users as well as developers. Or do you think it's ok?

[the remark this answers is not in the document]
ihcsim: Removing field from the API requires a bit more planning. Personally, I want to remove the actionset level option but I think it's beyond the scope of this design.

design/func-version.md
viveksinghggits: ok, so IIUC we don't want to access the global map that is the only reason we are trying to persist the supported versions in this file?
Also, we are slightly changing (`RegisterVersionedFunc`) how new function are going to register themselves? Because until now functions used to register themselves in their respective source files.
I am not sure if we are going to get into problems if we keep using that global map but it looks like we now have two places where we maintain the function versions, that global map and this file.

design/func-version.md
viveksinghggits: this name is a bit confusing :wink:

ihcsim: How so? It's meant to be a getter, without adding the `Get` prefix, per Go's convention.

design/func-version.md
ihcsim: We use this file to populate the global map. The phase execution code will still look up the global map for the function and version that the phase wants. Ideally, I want to get rid of the global map, but that will break many things.

Whatever approach we take, there will be two places with the function versions anyway. We need to load them from a source into the global map. Do we rather hard-code all the versions of all functions?

The main reason why I want to store these versions in a file is so that we can look at the repo and easily determine the states of all the function versions, at each commit. I added this info in the "Versions Discovery" section.

mlavi:
In our discussion today, as I was coming up to speed (and still am), I mentioned Feature Flags for conditional execution of different function versions. The most common use case for Feature Flags are boolean toggles of functionality via an external control plane dependency (SDK and network dependencies for implementation). https://featureflags.io/multivariate-feature-flags/ could suffice for saying something like: io.kanester.function.foo.stable = 0.4.3

One of my colleagues works with https://www.getunleash.io for local, on-prem deployment, which is mentioned on https://featureflags.io/go-feature-flags/

It seems like overkill and places a new management, deployment, configuration, and security burden to externalize this need. We need something simpler, but It should be food for through in the future.  I recall discussion that even a configmap wouldn't be desirable, because it would be too external? So the solution needs to be much more local to the codebase? And there still is a need for multiple function versions to be solved, but going back to feature flags, it suggests that binary toggles of functions are simpler than multiple versions.

Suggest we ask around and model any other working code! There was mention of K8s snapshot API, any pointers?

ihcsim:
@viveksinghggits @pavannd1 Per offline discussion, feel free to propose alternatives here for refactoring the existing `funcs` global map to represent the function versions stability status. We also discussed the scope of updating all the existing functions if we changed the `Func` interface. Please include the findings here. Thanks.

PR #1650, opened 2022-09-29
ActionSet handling perf tweaks

### LINKED ISSUE #1646: [BUG] Poor performance of ActionSet processing
**Describe the bug**
When I create multiple action sets simultaneously, they are processed with delays on Kanister side.

**To Reproduce**
1. Add debug messages into `onAdd` event handler in Kanister's `pkg/controller/controller.go` file.
2. Create multiple action sets simultaneously
3. Observe in Kanister log that `onAdd` debug messages appear with delays.

**Expected behavior**
I suppose that processing of multiple action sets should be faster. At the moment this looks like a performance bottleneck for applications using Kanister.

**Environment**
Kubernetes Version/Provider: 1.21
Storage Provider: GKE
Cluster Size (#nodes): 3

**Additional context**
As an experiment I modified `pkg/controller/controller.go` with the following log calls:
```
+func objToString(obj interface{}) string {
+       switch v := obj.(type) {
+       case *crv1alpha1.ActionSet:
+               return fmt.Sprintf("ActionSet{\"name\": \"%s\", \"uuid\": \"%s\"}", v.Name, v.UID)
+       case *crv1alpha1.Blueprint:
+               return fmt.Sprintf("Blueprint{\"name\": \"%s\", \"uuid\": \"%s\"}", v.Name, v.UID)
+       default:
+               return fmt.Sprintf("UnknownObject{\"type\": \"%T\"}", obj)
+       }
+}
+
 func (c *Controller) onAdd(obj interface{}) {
+       log.Info().Print("Controller::onAdd", field.M{"obj": objToString(obj)})
        o, ok := obj.(runtime.Object)
        if !ok {
                objType := fmt.Sprintf("%T", obj)
@@ -176,6 +188,7 @@ func (c *Controller) onUpdate(oldObj, newObj interface{}) {
 }
 
 func (c *Controller) onDelete(obj interface{}) {
+       log.Info().Print("Controller::onDelete", field.M{"obj": objToString(obj)})
        switch v := obj.(type) {
        case *crv1alpha1.ActionSet:
                if err := c.onDeleteActionSet(v); err != nil {
```

Then in a test application I created 10 action sets at once via golang k8s api, waited for their completion and deleted all of them. Observed the following in logs of a test application: 
```
"Time: 20220922-16:07:35.351Z, Message Created ActionSet pvc-6-ngir57wj" <---- Such messages are logged right after calling ActionSet.Create()
"Time: 20220922-16:07:36.123Z, Message Created ActionSet pvc-1-ngil6thx" 
"Time: 20220922-16:07:39.113Z, Message Created ActionSet pvc-9-ngikbwr6"
"Time: 20220922-16:07:39.917Z, Message Created ActionSet pvc-3-ngixrv48"
"Time: 20220922-16:07:40.708Z, Message Created ActionSet pvc-5-ngipkztn"
"Time: 20220922-16:07:42.316Z, Message Created ActionSet pvc-8-ngizglzm"
"Time: 20220922-16:07:42.520Z, Message Created ActionSet pvc-10-ngt67pv"
"Time: 20220922-16:07:43.711Z, Message Created ActionSet pvc-2-ngifjqwn"
"Time: 20220922-16:07:44.509Z, Message Created ActionSet pvc-4-ngix545n"
"Time: 20220922-16:07:44.716Z, Message Created ActionSet pvc-7-ngiz5nrt"
"Time: 20220922-16:07:52.116Z, Message Deleted ActionSet pvc-6-ngir57wj" <---- Such messages are logged right after ActionSet.Delete()
"Time: 20220922-16:07:54.313Z, Message Deleted ActionSet pvc-1-ngil6thx"
"Time: 20220922-16:07:58.332Z, Message Deleted ActionSet pvc-9-ngikbwr6"
"Time: 20220922-16:08:00.142Z, Message Deleted ActionSet pvc-3-ngixrv48"
"Time: 20220922-16:08:10.946Z, Message Deleted ActionSet pvc-5-ngipkztn"
"Time: 20220922-16:08:21.973Z, Message Deleted ActionSet pvc-8-ngizglzm"
"Time: 20220922-16:08:32.215Z, Message Deleted ActionSet pvc-10-ngt67pv"
"Time: 20220922-16:08:52.594Z, Message Deleted ActionSet pvc-2-ngifjqwn"
"Time: 20220922-16:08:53.194Z, Message Deleted ActionSet pvc-4-ngix545n"
"Time: 20220922-16:08:53.782Z, Message Deleted ActionSet pvc-7-ngiz5nrt"
```
and on Kanister side:
```
"Time: 2022-09-22T16:07:35.3500057Z, Message Controller::onAdd pvc-6-ngir57wj"
"Time: 2022-09-22T16:07:36.12121531Z, Message Controller::onAdd pvc-1-ngil6thx"
"Time: 2022-09-22T16:07:39.110795426Z, Message Controller::onAdd pvc-9-ngikbwr6"
"Time: 2022-09-22T16:07:40.167924955Z, Message Controller::onAdd pvc-3-ngixrv48"
"Time: 2022-09-22T16:07:45.775161887Z, Message Controller::onAdd pvc-5-ngipkztn"
"Time: 2022-09-22T16:07:55.769729013Z, Message Controller::onAdd pvc-8-ngizglzm"
"Time: 2022-09-22T16:08:02.167332988Z, Message Controller::onAdd pvc-10-ngt67pv"
"Time: 2022-09-22T16:08:06.766619434Z, Message Controller::onAdd pvc-2-ngifjqwn"
"Time: 2022-09-22T16:08:13.173679056Z, Message Controller::onAdd pvc-4-ngix545n"
"Time: 2022-09-22T16:08:19.769290363Z, Message Controller::onAdd pvc-7-ngiz5nrt"
"Time: 2022-09-22T16:08:26.167965139Z, Message Controller::onDelete pvc-6-ngir57wj"
"Time: 2022-09-22T16:08:26.169936496Z, Message Controller::onDelete pvc-1-ngil6thx"
"Time: 2022-09-22T16:08:26.172442064Z, Message Controller::onDelete pvc-9-ngikbwr6"
"Time: 2022-09-22T16:08:26.177037769Z, Message Controller::onDelete pvc-3-ngixrv48"
"Time: 2022-09-22T16:08:26.187949371Z, Message Controller::onDelete pvc-5-ngipkztn"
"Time: 2022-09-22T16:08:26.195837893Z, Message Controller::onDelete pvc-8-ngizglzm"
"Time: 2022-09-22T16:08:32.210891539Z, Message Controller::onDelete pvc-10-ngt67pv"
"Time: 2022-09-22T16:08:52.594266832Z, Message Controller::onDelete pvc-2-ngifjqwn"
"Time: 2022-09-22T16:08:53.193679498Z, Message Controller::onDelete pvc-4-ngix545n"
"Time: 2022-09-22T16:08:53.779738608Z, Message Controller::onDelete pvc-7-ngiz5nrt"
```

Here we can see that there's 44 sec delay between moments of adding pvc-6-ngir57wj and pvc-7-ngiz5nrt.

ed-shilo:
## Change Overview

The proposed set of changes improves perf of Add ActionSet workflow handling in Kanister controller by fixing the following issues:
1)  Currently controller's `onAdd `event handler starts an ActionSet execution goroutines after prevalidation stage, which contains several API calls to K8S apiserver. These calls take several seconds before spawning a goroutine, and considering the fact that go-client cache calls `onAdd` sequentially this adds significant time delay before starting ActionSets if multiple ones are submitted simultaneously.
2) If multiple ActionSets are handled in parallel, this makes a load on k8s api servers and makes it less responsive. It was measured that for 10 simultaneous ActionSet executions each k8s API call may take up to 1 sec during execution. To reduce server load it was decided to reduce number of k8s API calls where it's possible. For example, all calls to get ActionSet blueprint were replaced with a single call to Blueprint's List method with further lookup through received Blueprints.

As a proof of perf improvement I measured execution time of several ActionSets submitted simultaneously depending on their count and phase count in a corresponding blueprint. A test ActionSet just performed single KubeExec call to minimize phase execution impact.  

Results: 

| Phases per Blueprint | Parallel ActionSets | Old execution time | New execution time |
|----------------------|---------------------|--------------------|--------------------|
| 1                    | 10                  | 35 sec             | 24 sec             |
| 1                    | 20                  | 1 min 12 sec       | 1 min 1 sec        |
| 1                    | 30                  | 1 min 50 sec       | 1 min 16 sec       |
| 3                    | 10                  | 56 sec             | 42 sec             |
| 3                    | 20                  | 1 min 53 sec       | 1 min 22 sec       |
| 3                    | 30                  | 2 min 51 sec       | 2 min 4 sec        |
| 5                    | 10                  | 1 min 12 sec       | 1 min 2 sec        |
| 5                    | 20                  | 2 min 26 sec       | 2 min 4 sec        |
| 5                    | 30                  | 3 min 26 sec       | 3 min 0 sec        |

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #1646 

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -27,7 +27,7 @@ import (
 	"strings"
 	"sync"
 
-	customresource "github.com/kanisterio/kanister/pkg/customresource"
+	"github.com/kanisterio/kanister/pkg/customresource"
 	"github.com/pkg/errors"
 	"gopkg.in/tomb.v2"
 	corev1 "k8s.io/api/core/v1"
@@ -75,6 +75,8 @@ func New(c *rest.Config) *Controller {
 
 // StartWatch watches for instances of ActionSets and Blueprints acts on them.
 func (c *Controller) StartWatch(ctx context.Context, namespace string) error {
+	log.Info().Print(fmt.Sprintf("Config QPS is %v", c.config.QPS))
+
 	crClient, err := versioned.NewForConfig(c.config)
 	if err != nil {
 		return errors.Wrap(err, "failed to get a CustomResource client")
@@ -146,9 +148,14 @@ func (c *Controller) onAdd(obj interface{}) {
 	o = o.DeepCopyObject()
 	switch v := o.(type) {
 	case *crv1alpha1.ActionSet:
-		if err := c.onAddActionSet(v); err != nil {
-			log.Error().WithError(err).Print("Callback onAddActionSet() failed")
-		}
+		t, ctx := c.LoadOrStoreTomb(context.Background(), v.Name)
+		t.Go(func() error {
+			if err := c.onAddActionSet(ctx, t, v); err != nil {
+				log.Error().WithError(err).Print("Callback onAddActionSet() failed")
+			}
+			return nil
+		})
+
 	case *crv1alpha1.Blueprint:
 		c.onAddBlueprint(v)
 	default:
@@ -191,25 +198,29 @@ func (c *Controller) onDelete(obj interface{}) {
 	}
 }
 
-func (c *Controller) onAddActionSet(as *crv1alpha1.ActionSet) error {
-	as, err := c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Get(context.TODO(), as.GetName(), v1.GetOptions{})
+func (c *Controller) onAddActionSet(ctx context.Context, t *tomb.Tomb, as *crv1alpha1.ActionSet) error {
+	as, err := c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Get(ctx, as.GetName(), v1.GetOptions{})
 	if err != nil {
 		return errors.WithStack(err)
 	}
 	if err := validate.ActionSet(as); err != nil {
 		return err
+	}
+	bps, err := initBpCache(ctx, c.crClient, as.GetNamespace())
+	if err != nil {
+		return errors.WithStack(err)
 	}
 	if as.Status == nil {
-		c.initActionSetStatus(as)
+		c.initActionSetStatus(ctx, as, bps)
 	}
-	as, err = c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Get(context.TODO(), as.GetName(), v1.GetOptions{})
+	as, err = c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Get(ctx, as.GetName(), v1.GetOptions{})
 	if err != nil {
 		return errors.WithStack(err)
 	}
 	if err := validate.ActionSet(as); err != nil {
 		return err
 	}
-	return c.handleActionSet(as)
+	return c.handleActionSet(ctx, t, as, bps)
 }
 
 func (c *Controller) onAddBlueprint(bp *crv1alpha1.Blueprint) {
@@ -283,8 +294,7 @@ func (c *Controller) onDeleteBlueprint(bp *crv1alpha1.Blueprint) {
 	log.Print("Deleted Blueprint ", field.M{"BlueprintName": bp.GetName()})
 }
 
-func (c *Controller) initActionSetStatus(as *crv1alpha1.ActionSet) {
-	ctx := context.Background()
+func (c *Controller) initActionSetStatus(ctx context.Context, as *crv1alpha1.ActionSet, bps *bpCache) {
 	ctx = field.Context(ctx, consts.ActionsetNameKey, as.GetName())
 	if as.Spec == nil {
 		log.Error().WithContext(ctx).Print("Cannot initialize an ActionSet without a spec.")
@@ -294,12 +304,10 @@ func (c *Controller) initActionSetStatus(as *crv1alpha1.ActionSet) {
 	actions := make([]crv1alpha1.ActionStatus, 0, len(as.Spec.Actions))
 	var err error
 	for _, a := range as.Spec.Actions {
-		var actionStatus *crv1alpha1.ActionStatus
-		actionStatus, err = c.initialActionStatus(as.GetNamespace(), a)
+		actionStatus, err := c.initialActionStatus(a, bps)
 		if err != nil {
-			bp, _ := c.crClient.CrV1alpha1().Blueprints(as.GetNamespace()).Get(ctx, a.Blueprint, v1.GetOptions{})
 			reason := fmt.Sprintf("ActionSetFailed Action: %s", a.Name)
-			c.logAndErrorEvent(ctx, "Could not get initial action:", reason, err, as, bp)
+			c.logAndErrorEvent(ctx, "Could not get initial action:", reason, err, as, bps.Get(a.Blueprint))
 			break
 		}
 		actions = append(actions, *actionStatus)
@@ -318,14 +326,14 @@ func (c *Controller) initActionSetStatus(as *crv1alpha1.ActionSet) {
 	}
 }
 
-func (c *Controller) initialActionStatus(namespace string, a crv1alpha1.ActionSpec) (*crv1alpha1.ActionStatus, error) {
+func (c *Controller) initialActionStatus(a crv1alpha1.ActionSpec, bps *bpCache) (*crv1alpha1.ActionStatus, error) {
 	if a.Blueprint == "" {
 		// TODO: If no blueprint is specified, we should consider a default.
 		return nil, errors.New("Blueprint not specified")
 	}
-	bp, err := c.crClient.CrV1alpha1().Blueprints(namespace).Get(context.TODO(), a.Blueprint, v1.GetOptions{})
-	if err != nil {
-		return nil, errors.Wrap(err, "Failed to query blueprint")
+	bp := bps.Get(a.Blueprint)
+	if bp == nil {
+		return nil, errors.New("Failed to retrieve blueprint " + a.Blueprint)
 	}
 	bpa, ok := bp.Actions[a.Name]
 	if !ok {
@@ -357,7 +365,7 @@ func (c *Controller) initialActionStatus(namespace string, a crv1alpha1.ActionSp
 	return actionStatus, nil
 }
 
-func (c *Controller) handleActionSet(as *crv1alpha1.ActionSet) (err error) {
+func (c *Controller) handleActionSet(ctx context.Context, t *tomb.Tomb, as *crv1alpha1.ActionSet, bps *bpCache) (err error) {
 	if as.Status == nil {
 		return errors.New("ActionSet was not initialized")
 	}
@@ -365,10 +373,9 @@ func (c *Controller) handleActionSet(as *crv1alpha1.ActionSet) (err error) {
 		return nil
 	}
 	as.Status.State = crv1alpha1.StateRunning
-	if as, err = c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Update(context.TODO(), as, v1.UpdateOptions{}); err != nil {
+	if as, err = c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Update(ctx, as, v1.UpdateOptions{}); err != nil {
 		return errors.WithStack(err)
 	}
-	ctx := context.Background()
 	ctx = field.Context(ctx, consts.ActionsetNameKey, as.GetName())
 	// adding labels with prefix "kanister.io/" in the context as field for better logging
 	for key, value := range as.GetLabels() {
@@ -386,11 +393,10 @@ func (c *Controller) handleActionSet(as *crv1alpha1.ActionSet) (err error) {
 	}()
 
 	for i := range as.Status.Actions {
-		if err = c.runAction(ctx, as, i); err != nil {
+		bp := bps.Get(as.Spec.Actions[i].Blueprint)
+		if err = c.runAction(ctx, t, as, i, bp); err != nil {
 			// If runAction returns an error, it is a failure in the synchronous
 			// part of running the action.
-			bpName := as.Spec.Actions[i].Blueprint
-			bp, _ := c.crClient.CrV1alpha1().Blueprints(as.GetNamespace()).Get(ctx, bpName, v1.GetOptions{})
 			reason := fmt.Sprintf("ActionSetFailed Action: %s", as.Status.Actions[i].Name)
 			c.logAndErrorEvent(ctx, fmt.Sprintf("Failed to launch Action %s:", as.GetName()), reason, err, as, bp)
 			as.Status.State = crv1alpha1.StateFailed
@@ -406,15 +412,21 @@ func (c *Controller) handleActionSet(as *crv1alpha1.ActionSet) (err error) {
 	return nil
 }
 
+func (c *Controller) LoadOrStoreTomb(ctx context.Context, asName string) (*tomb.Tomb, context.Context) {
+	var t *tomb.Tomb
+	if v, ok := c.actionSetTombMap.Load(asName); ok {
+		t = v.(*tomb.Tomb)
+	} else {
+		t, ctx = tomb.WithContext(ctx)
+		c.actionSetTombMap.Store(asName, t)
+	}
+	return t, ctx
+}
+
 // nolint:gocognit
-func (c *Controller) runAction(ctx context.Context, as *crv1alpha1.ActionSet, aIDX int) error {
+func (c *Controller) runAction(ctx context.Context, t *tomb.Tomb, as *crv1alpha1.ActionSet, aIDX int, bp *crv1alpha1.Blueprint) error {
 	action := as.Spec.Actions[aIDX]
 	c.logAndSuccessEvent(ctx, fmt.Sprintf("Executing action %s", action.Name), "Started Action", as)
-	bpName := as.Spec.Actions[aIDX].Blueprint
-	bp, err := c.crClient.CrV1alpha1().Blueprints(as.GetNamespace()).Get(ctx, bpName, v1.GetOptions{})
-	if err != nil {
-		return errors.WithStack(err)
-	}
 	tp, err := param.New(ctx, c.clientset, c.dynClient, c.crClient, c.osClient, action)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


viveksinghggits (commented):
LGTM. I will let someone else have a look as well. thank you, good find :100:

ed-shilo:
@ihcsim Could you please take a look at these changes too?

pkg/param/param.go
```diff
@@ -513,7 +513,7 @@ func UpdatePhaseParams(ctx context.Context, tp *TemplateParams, phaseName string
 
 // UpdateDeferPhaseParams updates the TemplateParams deferPhase output with passed output
 // This output would be generated/passed by execution of the phase
-func UpdateDeferPhaseParams(ctx context.Context, tp *TemplateParams, output map[string]interface{}) {
```
denisvmedia, line 516: This change probably shouldn't belong to this PR.

ed-shilo: Just removed unused parameter

denisvmedia: Yes, but this makes the PR harder to read. Also, this is an exported function. And may be used by those who use kanister as a library. I'd better not change it at all. But even if it has to be changed, I think it deserves its own PR.

ed-shilo: Ok, reverted it back

pkg/controller/controller.go
```diff
@@ -294,12 +302,20 @@ func (c *Controller) initActionSetStatus(as *crv1alpha1.ActionSet) {
 	actions := make([]crv1alpha1.ActionStatus, 0, len(as.Spec.Actions))
 	var err error
 	for _, a := range as.Spec.Actions {
-		var actionStatus *crv1alpha1.ActionStatus
-		actionStatus, err = c.initialActionStatus(as.GetNamespace(), a)
+		if a.Blueprint == "" {
+			// TODO: If no blueprint is specified, we should consider a default.
+			err = errors.New("Blueprint not specified")
```
pavannd1, line 307: Should we call `c.logAndErrorEvent` here?

pkg/controller/controller.go
```diff
@@ -294,12 +302,20 @@ func (c *Controller) initActionSetStatus(as *crv1alpha1.ActionSet) {
...
+			break
+		}
+		bp := bps.Get(a.Blueprint)
+		if bp == nil {
+			err = errors.New("Failed to retrieve blueprint " + a.Blueprint)
```
pavannd1, line 312: Same as above

pkg/controller/controller.go
```diff
@@ -406,15 +414,21 @@ func (c *Controller) handleActionSet(as *crv1alpha1.ActionSet) (err error) {
...
+func (c *Controller) LoadOrStoreTomb(ctx context.Context, asName string) (*tomb.Tomb, context.Context) {
+	var t *tomb.Tomb
+	if v, ok := c.actionSetTombMap.Load(asName); ok {
+		t = v.(*tomb.Tomb)
+	} else {
+		t, ctx = tomb.WithContext(ctx)
+		c.actionSetTombMap.Store(asName, t)
+	}
+	return t, ctx
+}
```
denisvmedia, line 426: nit (to avoid using `else`):

```suggestion
	var t *tomb.Tomb
	if v, ok := c.actionSetTombMap.Load(asName); ok {
		t = v.(*tomb.Tomb)
		return t, ctx
	t, ctx = tomb.WithContext(ctx)
	c.actionSetTombMap.Store(asName, t)
	return t, ctx
}
```

pkg/controller/controller.go
ed-shilo: Fixed

pkg/controller/controller.go
ed-shilo: Fixed

pkg/controller/controller.go
ed-shilo: Fixed

pkg/controller/controller.go
```diff
@@ -671,3 +681,25 @@ func setObjectKind(obj runtime.Object) {
 	}
 	ok.SetGroupVersionKind(gvk)
 }
+
+type bpCache struct {
+	bpMap map[string]*crv1alpha1.Blueprint
+}
+
+func initBpCache(ctx context.Context, cli versioned.Interface, ns string) (*bpCache, error) {
```
denisvmedia, line 689: I'm wondering why `initBpCache` is a separate function? Did you consider it to be a part of `Controller`?

pkg/controller/controller.go
```diff
@@ -386,11 +395,10 @@ func (c *Controller) handleActionSet(as *crv1alpha1.ActionSet) (err error) {
 	}()
 
 	for i := range as.Status.Actions {
-		if err = c.runAction(ctx, as, i); err != nil {
+		bp := bps.Get(as.Spec.Actions[i].Blueprint)
```
denisvmedia, line 398: What if bp == nil? (not found in the cache)

pkg/controller/controller.go
ed-shilo: Not, it's assumed to be a part of `onAdd` function context.

pkg/controller/controller.go
ed-shilo: It's checked that bp is in cache during init action set stage, so no need to check it here one more time

denisvmedia: Cannot it happen that between you init the cache and you make a call to bps.Get something will change? In other words, do we have any kind of lock that guarantees the atomicity of the operations?

pkg/controller/controller.go
ed-shilo: Fixed

pkg/controller/controller.go
ed-shilo: Currently, all information required for phase execution is stored inside of Blueprint object. With introducing bp cache the code became more atomic than it was before, because each onAdd has a snapshot of blueprint information required for phases. Unfortunately, the whole code is not atomic, but this is out of the task scope.

ed-shilo: Moved bp cache initialization function to controller

denisvmedia: > Unfortunately, the whole code is still not atomic, but this is out of the task scope.

I think, if you just check the bp value (ok in your current approach the value of `ok`), then it will be safer than just hope.

pkg/controller/controller.go
```diff
@@ -66,6 +66,8 @@ type Controller struct {
 	actionSetTombMap sync.Map
 }
 
+type bpCache map[string]*crv1alpha1.Blueprint
```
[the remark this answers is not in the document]
ed-shilo: It looks simpler among function args, don't it?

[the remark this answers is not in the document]
denisvmedia: To me it just hides the information instead. But I might be biased or subjective. Do we have any kind of a similar approach in Kanister?

pkg/controller/controller.go
ed-shilo: Ok, will add this check here. But the cache is not shared among onAdd event handlers, so nothing will happen with it after initialization.

pkg/controller/controller.go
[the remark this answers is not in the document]
ed-shilo: No, I don't see that this is widespread in Kanister. Ok, will fix

pkg/controller/controller.go
ed-shilo: Fixed

pkg/controller/controller.go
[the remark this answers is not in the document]
ed-shilo: Fixed

denisvmedia (approved):
LGTM

pavannd1 (approved):
Thank you @ed-shilo 💯

pkg/controller/controller.go
```diff
@@ -191,25 +196,43 @@ func (c *Controller) onDelete(obj interface{}) {
...
+
+func (c *Controller) onAddActionSet(ctx context.Context, t *tomb.Tomb, as *crv1alpha1.ActionSet) error {
+	as, err := c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Get(ctx, as.GetName(), v1.GetOptions{})
 	if err != nil {
 		return errors.WithStack(err)
 	}
 	if err := validate.ActionSet(as); err != nil {
 		return err
 	}
+	bps, err := c.createBpCache(ctx, as.GetNamespace())
```
ihcsim, line 221: This will get expensive during a controller restart or if the actionset get re-queued. See https://github.com/kanisterio/kanister/pull/1267. IIRC, on restart, the controller will call `onAddActionSet()` on all existing actionset resources. With this change, the controller will create a map of _all_ blueprints for _every_  known actionset.

ihcsim (commented):
Thanks for working on this. I left some questions and comments. If needed, we can address them in a follow-up PR.

pkg/controller/controller.go
```diff
@@ -191,25 +196,43 @@ func (c *Controller) onDelete(obj interface{}) {
...
+func (c *Controller) createBpCache(ctx context.Context, ns string) (map[string]*crv1alpha1.Blueprint, error) {
+	bps, err := c.crClient.CrV1alpha1().Blueprints(ns).List(ctx, v1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	bpMap := make(map[string]*crv1alpha1.Blueprint)
+	for _, bp := range bps.Items {
+		bpMap[bp.Name] = bp
+	}
```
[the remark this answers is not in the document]
ed-shilo: I don't see where this is validated, but if I'm not mistaken according to the code actions in an actionset may have associations with different blueprints.
Using an informer will add need to wait after adding a blueprint and before adding an actions due to cache resync. Do we want such behavior?

pkg/controller/controller.go
```diff
@@ -191,25 +196,43 @@ func (c *Controller) onDelete(obj interface{}) {
...
+
+	return bpMap, nil
+}
+
+func (c *Controller) onAddActionSet(ctx context.Context, t *tomb.Tomb, as *crv1alpha1.ActionSet) error {
+	as, err := c.crClient.CrV1alpha1().ActionSets(as.GetNamespace()).Get(ctx, as.GetName(), v1.GetOptions{})
```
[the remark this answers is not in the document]
ed-shilo: Seems like it's unnecessary, need to remove

pkg/controller/controller.go
```diff
@@ -294,10 +316,20 @@ func (c *Controller) initActionSetStatus(as *crv1alpha1.ActionSet) {
...
 	for _, a := range as.Spec.Actions {
-		var actionStatus *crv1alpha1.ActionStatus
-		actionStatus, err = c.initialActionStatus(as.GetNamespace(), a)
+		if a.Blueprint == "" {
+			// TODO: If no blueprint is specified, we should consider a default.
+			err = errors.New("Blueprint is not specified for action")
+			c.logAndErrorEvent(ctx, "Could not get blueprint:", "Blueprint not specified", err, as)
+			break
+		}
+		bp, ok := bps[a.Blueprint]
```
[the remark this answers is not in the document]
ed-shilo: Same questions about adding need to wait between bp and as creation.

pkg/controller/controller.go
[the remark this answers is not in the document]
viveksinghggits: not really @ed-shilo. An actionset can point to just one blueprint.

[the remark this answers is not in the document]
viveksinghggits: created blueprints not being available immediately *might* be a concern.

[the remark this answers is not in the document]
ihcsim: The blueprint informer watches for new blueprints and updates its indexer accordingly. Try add a new blueprint, the blueprint informer should respond immediately. I am not sure why we are using the low level `cache` directly [here](https://github.com/kanisterio/kanister/blob/a83a548072c420d676c47e1daac833cbe46cfcb4/pkg/controller/controller.go#L105-L122), but it looks like we are already using client-go caching mechanism.

Re-sync applies to _existing_ resources, when client-go tries to reconcile any differences between the cache and the data in its store.

@ed-shilo You are right about the 1:n mapping between actionset and blueprint. It looks like an actionset can have [multiple actions](https://github.com/kanisterio/kanister/blob/a83a548072c420d676c47e1daac833cbe46cfcb4/pkg/apis/cr/v1alpha1/types.go#L77-L80), each of which can point to whichever blueprint it wants, even though downstream normally associate one actionset to one blueprint. Not sure why the API is designed that way.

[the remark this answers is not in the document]
ed-shilo: As far as I see, each action inside of actionset has separate blueprint field. Is it formally assumed that it should have the same value in case of multiple actions?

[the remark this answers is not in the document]
ed-shilo: @ihcsim tried to use lister for getting blueprints, but as I see a newly created blueprint goes into cache after some small time. In tests this is observed in failing of attempts to add an action sets linked with an added blueprint due to cache miss. Are we OK with this?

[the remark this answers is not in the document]
viveksinghggits: You are right, even though we don't use different blueprints for separate actions in an actionset, it does look like the code allows it.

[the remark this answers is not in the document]
ihcsim: > Is it formally assumed that it should have the same value in case of multiple actions?

No, we can't assume that anymore. If the API let user do multiple blueprints, we have to support it.

Re: missing cache, did you use the informers lister or the [cache lister](https://pkg.go.dev/k8s.io/client-go/tools/cache#Lister). Looks like our code is using the cache one.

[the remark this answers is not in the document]
ed-shilo: @ihcsim I used the lister from [the link](https://github.com/kanisterio/kanister/blob/c2a8e32e36a80a26d424f99699422d972667ec98/pkg/client/informers/externalversions/cr/v1alpha1/blueprint.go#L37-L40) you proposed previosly. As far as I see, it uses caching.

pkg/controller/controller.go
```diff
@@ -146,9 +146,14 @@ func (c *Controller) onAdd(obj interface{}) {
 	o = o.DeepCopyObject()
 	switch v := o.(type) {
 	case *crv1alpha1.ActionSet:
-		if err := c.onAddActionSet(v); err != nil {
-			log.Error().WithError(err).Print("Callback onAddActionSet() failed")
-		}
+		t, ctx := c.LoadOrStoreTomb(context.Background(), v.Name)
```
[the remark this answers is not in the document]
ed-shilo: As I understand, originally the tomb was used for graceful action set deletion. In fact each submitted goroutine doesn't listen for tomb signals, but I suppose it's better to leave them wrapped with the tomb.

ed-shilo:
@ihcsim @viveksinghggits @pavannd1 I've opened the follow-up PR, which addresses the rest of comments left after merge. I decided to return back direct retrievals of blueprints from the API server, as it seems that using caches brings much more problems in comparison with its contribution to overall workflow runtime.

PR #1666, opened 2022-10-09
initializeClients function to return struct of the clients. 

### LINKED ISSUE #531: CleanUp the kanctl code to return a struct of clients
In current implementation the [initializeClients](https://github.com/kanisterio/kanister/blob/master/pkg/kanctl/kanctl.go#L80)  function returns four variable instead of that we should rather return a struct of the clients.

harshitasao:
Signed-off-by: harshitasao <harshitasao@gmail.com>

## Change Overview

Made the initializeClients function return a struct of clients so that when we try to return any other client, the list does not continue to grow.
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #531 

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kanctl/kanctl.go
```diff
@@ -73,24 +79,30 @@ func resolveNamespace(cmd *cobra.Command) (string, error) {
 	return kube.ConfigNamespace()
 }
 
-func initializeClients() (kubernetes.Interface, versioned.Interface, osversioned.Interface, error) {
+func initializeClients() (*Clients, error) {
+	nilClient := Clients{}
 	config, err := kube.LoadConfig()
 	if err != nil {
-		return nil, nil, nil, err
+		return &nilClient, err
```
ihcsim, line 86: Why not just `return nil, err`? Same for line 90, 95, 100.

pkg/kanctl/kanctl.go
```diff
@@ -39,6 +39,12 @@ var (
 	Verbose bool
 )
 
+type Clients struct {
+	KubeClient kubernetes.Interface
+	CrdClient  versioned.Interface
```
ihcsim, line 44: Can we name this `KanisterClient`, since it calls the Kanister API?

pkg/kanctl/kanctl.go
```diff
@@ -39,6 +39,12 @@ var (
...
+	OsClient   osversioned.Interface
```
ihcsim, line 45: s/OsClient/OSClient

following the Golang convention on initialism: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

ihcsim:
@viveksinghggits PTAL.

pkg/kanctl/actionset.go
```diff
@@ -105,24 +105,24 @@ func newActionSetCmd() *cobra.Command {
 }
 
 func initializeAndPerform(cmd *cobra.Command, args []string) error {
-	cli, crCli, osCli, err := initializeClients()
+	clients, err := initializeClients()
 	if err != nil {
 		return err
 	}
-	params, err := extractPerformParams(cmd, args, cli, osCli)
+	params, err := extractPerformParams(cmd, args, clients.KubeClient, clients.OsClient)
```
viveksinghggits, line 112: @ihcsim would it be a good idea if we directly pass `clients` here?

pkg/kanctl/kanctl.go
```diff
@@ -39,6 +39,12 @@ var (
 	Verbose bool
 )
 
+type Clients struct {
+	KanisterClient kubernetes.Interface
+	CrdClient      versioned.Interface
+	OSClient       osversioned.Interface
+}
```
PrasadG193, line 46:
```suggestion
type Clients struct {
	KubeClient kubernetes.Interface
	KanisterClient      versioned.Interface
	OSClient       osversioned.Interface
}
```
Let's rename CrdClient to KanisterClient

PrasadG193: https://github.com/kanisterio/kanister/pull/1666#discussion_r992614443
I think there is confusion. In this comment, Ivan suggested renaming CrdClient not KubeClient :)

harshitasao: Understood, thanks!

viveksinghggits:
LGTM overall. @harshitasao will be it possible for you to add any tests that you have performed after the changes. I would recommend, building the kanctl binary and then creating the profile/actionset object would be good idea. You can follow any example from here after building the binary https://github.com/kanisterio/kanister/tree/master/examples/mysql.

PR #1673, opened 2022-10-11
Update kopia dockerfile for goboring

bathina2:
## Change Overview

This PR builds the kopia image using GO_BORING flags instead of using upstream goreleaser.
It also changes the the base image to buster since alpine uses the musl-libc library which is not present in the UBI-minimal image.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


Makefile
```diff
@@ -263,7 +265,7 @@ release-snapshot:
 	@$(MAKE) run CMD='-c "GORELEASER_CURRENT_TAG=v9.99.9-dev goreleaser --debug release --rm-dist --snapshot"'
 
 update-kopia-image:
-	@/bin/bash ./build/update_kopia_image.sh $(KOPIA_COMMIT_ID) $(KOPIA_REPO)
+	@/bin/bash ./build/update_kopia_image.sh -c $(KOPIA_COMMIT_ID) -r $(KOPIA_REPO) -b $(GOBORING)
```
julio-lopez, line 268: How does this expand when invoked as `make update-kopia-image`? `-b` or `b ""`
Does it matter?

docker/kopia-build/Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM golang:1.19.1-alpine3.16 AS builder
+FROM golang:1.19.2-buster AS builder
```
julio-lopez, line 1: Wondering whether this should be `golang:1.19-buster` instead?
I theory, this could make the build not fully reproducible, but OTOH it allows _automatically_ getting the Golang updates.
Let's get @miquella input here.

Makefile
bathina2: Doesn't matter, will set it to empty.`-b ""`

docker/kopia-build/Dockerfile
```diff
@@ -15,24 +16,24 @@ WORKDIR /kopia
 
 # Build kopia binary from specific commit
 RUN git checkout ${kopiaBuildCommit}
-# Refers to goreleaser/goreleaser:v1.11.2
-COPY --from=docker.io/goreleaser/goreleaser@sha256:b13418f20019fffb29797aff3d03f5e9ca84cea93634f4169a7ed603a95ab198 \
-  /usr/bin/goreleaser /usr/bin/goreleaser
-RUN goreleaser build --output=. --rm-dist --single-target
+RUN GO111MODULE=on GOOS=linux GOARCH=amd64 go build -o kopia \
```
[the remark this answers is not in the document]
bathina2: Yea the previous line is `WORKDIR /kopia`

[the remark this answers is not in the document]
julio-lopez: I see. I missed that because it was not included in the default diff view, duh!

docker/kopia-build/Dockerfile
miquella: @julio-lopez: I chatted with @bathina2 about this and we believe that we should change to `golang:1.19-buster`, as you suggest.

Although there is a bit of a downside in that it's not explicit which version of 1.19 was being used for a previously built image, it does avoid a lot of the toil in upgrading the version of Go every ~month or so. The Go team is very careful to not release breaking changes in minor versions, so the risk of breaking things is fairly low.

julio-lopez (approved):
LGTM  ✅
🚢it :shipit: 🚀

PR #1695, opened 2022-10-20
Add Name to PodOptions

kale-amruta:
## Change Overview

Add an option to have fixed name for a pod

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod.go
+++ b/pkg/kube/pod.go
@@ -122,6 +122,7 @@ func CreatePod(ctx context.Context, cli kubernetes.Interface, opts *PodOptions) 
 
 	pod := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: opts.GenerateName,
 			Labels: map[string]string{
 				consts.LabelKeyCreatedBy: consts.LabelValueKanister,
 			},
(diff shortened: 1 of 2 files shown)


akankshakumari393:
LGTM

pkg/kube/pod.go
```diff
@@ -129,6 +130,11 @@ func CreatePod(ctx context.Context, cli kubernetes.Interface, opts *PodOptions)
 		Spec: patchedSpecs,
 	}
 
+	// Add fixed name to a pod
+	if opts.Name != "" {
```
ankitjain235, line 134: Do we need to not set `GenerateName` if `Name` is set? Or would the `Name` be chosen as precedence always? Should we fail pod creation if both are set?

ankitjain235 (commented):
Minor question, looks good otherwise.

pkg/kube/pod.go
kale-amruta: kubernetes give Name as precedence over GenerateName always. Hence it doesnt matter even if the field is set

pkg/kube/pod.go
```diff
@@ -129,6 +130,11 @@ func CreatePod(ctx context.Context, cli kubernetes.Interface, opts *PodOptions)
 		Spec: patchedSpecs,
 	}
 
+	// Add fixed name to a pod
```
pavannd1, line 133:
```suggestion
	// Override `GenerateName` if `Name` option is provided
```

pavannd1 (approved):
Minor comment update. LG otherwise

PR #1711, opened 2022-10-26
Build kopia binary as part of kanister-tools

ankitjain235:
## Change Overview

This PR updates workflow to build Kopia binary as a part of Kanister-tools image build instead of a separate image.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


docker/tools/Dockerfile
```diff
@@ -1,3 +1,49 @@
+# Build Kopia binary
+FROM golang:1.19-bullseye AS builder
+
+ARG kopiaBuildCommit
+ARG kopiaRepoOrg
```
PrasadG193, line 5: I think best practice is to use `snake_case` instead of camelCase for ARG and ENV

PrasadG193: If arg is not set, does it fail the build?

ankitjain235: We do expect the arg to be always set.

julio-lopez: Ankit: it is expected to be set when called via `goreleaser`. What happens if someone tries to build the image directly? what happens if the argument is unset? do things fails?

julio-lopez:
This is the right direction, there is no need for building kopia in a separate image and pushing it to a registry.

docker/tools/Dockerfile
ankitjain235: I have kept default values in Dockerfile, allowing them to be passed using build-arg if required.

docker/tools/Dockerfile
```diff
@@ -1,3 +1,50 @@
+# Build Kopia binary
+FROM golang:1.19-bullseye AS builder
+
+# Refers to https://github.com/kopia/kopia/commit/317cc36892707ab9bdc5f6e4dea567d1e638a070
+ARG kopiaBuildCommit=317cc36
```
PrasadG193, line 5: Should we use the latest commit by default? Maybe a master branch?

ankitjain235: Since we are cloning, it would be latest commit always so we can leave it empty as well. Setting master might confuse others if it should be branch instead of commit.

PrasadG193: sounds good

docker/tools/Dockerfile
```diff
@@ -1,3 +1,46 @@
...
+
+# Add CA certs
+RUN apt-get update && apt-get -y install ca-certificates && \
+  rm -rf /var/cache/apk/* && \
+  adduser kopia && addgroup kopia kopia && \
+  chown kopia /kopia
+
+USER kopia:kopia
+
+ENTRYPOINT [ "/kopia/kopia" ]
```
pavannd1, line 41: Why is this required?

pavannd1 (approved):
LGTM

PR #1749, opened 2022-11-20
gofmt issue fixes as per goreportcard (#1274)

### LINKED ISSUE #1274: Fix linter issues per Go Report
[Go report](https://goreportcard.com/report/github.com/kanisterio/kanister) is reporting many linter and formatting issues. We can use some help to clean them up. CI should also be updated to pick up these issues.

Bhargav-InfraCloud:
Signed-off-by: Bhargav Ravuri <bhargav.ravuri@infracloud.io>

## Change Overview
- Fix the go formatting issues reported by [goreportcardare][1].
- Add `gofmt` in [golint.sh](golint.sh) to warn for future commits.
- No functional or test logic changes made. This is only for go formatting.

## Pull request type

## Issues 

- fixes #1274

## Test Plan
- [x] :muscle: Manual

1. Run `make golint`.
2. Install [goreportcard][1] and run `goreportcard-cli -v`. It should print stats similar to: 
```
Grade .......... A+ 100.0%
Files .............. 31757
Issues ................. 0
gofmt ............... 100%
go_vet .............. 100%
ineffassign ......... 100%
gocyclo ............. 100%
license ............. 100%
misspell ............ 100%
```
`Note:` % of `gofmt` should be higher than what it was before this commit.

## Message for maintainers
- [golint][2] is archived and so it was disabled from [goreportcard][1] in [commit][3].
- Also, [goreportcard][1] uses [gometalinter][4] which is archived. The replacement is [golangci-lint][5] which is being used by Kanister already. However, most of the common linter in it (like, gofmt, revive, etc.) are not enabled or skipped for packages. I'd recommend to have a `.golangci.yml` configured to make most out of it.

---

[1]: https://github.com/gojp/goreportcard
[2]: https://github.com/golang/lint
[3]: https://github.com/gojp/goreportcard/commit/5aeaec8b44c4446e92e66e23f0e04b24e95da4a5
[4]: https://github.com/alecthomas/gometalinter
[5]: https://github.com/golangci/golangci-lint

### DIFF
--- a/build/golint.sh
+++ b/build/golint.sh
@@ -24,6 +24,13 @@ echo "Running golangci-lint..."
 
 golangci-lint run --timeout ${TIMEOUT} --skip-dirs ${SKIP_DIR_REGEX} -E maligned,whitespace,gocognit,unparam -e '`ctx` is unused'
 
+# gofmt should run everywhere, including
+#   1. Skipped directories in previous step
+#   2. Build exempted files using build tags
+#      Note: Future build tags should be included.
+echo "Running gofmt..."
+golangci-lint run --timeout ${TIMEOUT} --disable-all --enable gofmt --build-tags integration
+
 echo "PASS"
 echo
 
--- a/pkg/app/bp_test.go
+++ b/pkg/app/bp_test.go
@@ -36,15 +36,15 @@ var _ = Suite(&BlueprintSuite{})
 func (bs *BlueprintSuite) TestUpdateImageTags(c *C) {
 	for _, bp := range []*crv1alpha1.Blueprint{
 		// BP with no phase with image arg
-		&crv1alpha1.Blueprint{
+		{
 			ObjectMeta: metav1.ObjectMeta{
 				GenerateName: "test-blueprint-",
 			},
 			Actions: map[string]*crv1alpha1.BlueprintAction{
-				"test": &crv1alpha1.BlueprintAction{
+				"test": {
 					Kind: "Deployment",
 					Phases: []crv1alpha1.BlueprintPhase{
-						crv1alpha1.BlueprintPhase{
+						{
 							Func: function.KubeExecFuncName,
 							Name: "test-kube-exec",
 							Args: map[string]interface{}{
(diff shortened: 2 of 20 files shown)


build/golint.sh
```diff
@@ -24,6 +24,13 @@ echo "Running golangci-lint..."
 
 golangci-lint run --timeout ${TIMEOUT} --skip-dirs ${SKIP_DIR_REGEX} -E maligned,whitespace,gocognit,unparam -e '`ctx` is unused'
 
+# gofmt should run everywhere, including
+#   1. Skipped directories in previous step
+#   2. Build exempted files using build tags
+#      Note: Future build tags should be included.
+echo "Running gofmt..."
+golangci-lint run --timeout ${TIMEOUT} --disable-all --enable gofmt --build-tags integration
```
PrasadG193, line 32: @Bhargav-InfraCloud If we add this, we don't need the previous call (Line no 25), right?

Bhargav-InfraCloud: It would **require both**, actually. The first one (line 25) runs 7 default linters (`errcheck`, `gosimple`, `govet`, `ineffassign`, `staticcheck`, `typecheck` & `unused`) with 4 enabled (`maligned`, `whitespace`, `gocognit` & `unparam`).

The second (line 32) only runs `gofmt` which is the core of this change.

Ideally, `gofmt` could've been added to the first. But in doing so, the extra conditions (skipping the directory and excluding ctx error) will get applied to this which is not wise. And also, `gofmt` without `--build-tags integration` would ignore 2 files dropping the % in report. Enable it in first directly and it would throw a lot of unexpected linter errors (not required as of now).

As I suggested in PR, using `.golangci-yaml` can solve this issue, where we can specify all the configs and run it only once. But that would be for future PR if agreed upon.

pavannd1 (commented):
LGTM. Will let @PrasadG193 +2 this

Bhargav-InfraCloud:
@pavannd1 @PrasadG193 The [pipeline](https://github.com/kanisterio/kanister/actions/runs/3520182351/jobs/5920413781) probably was failing because of commit [e956100](https://github.com/kanisterio/kanister/commit/e956100e20d3271778402d7d6449ada9c8c52418) or [b48647f](https://github.com/kanisterio/kanister/commit/b48647fdd57d2d91e111947f2b7c279202dfc32c) that I merged from master into this branch without checking the github action's status. I've now rebased with latest successful commit. Can you please trigger the pipeline again?

build/golint.sh
PrasadG193: Thanks @Bhargav-InfraCloud could you please file an issue to switch to `.golanci-yaml`

PrasadG193 (approved):

Bhargav-InfraCloud:
Thanks again @PrasadG193 @pavannd1 :blush:

build/golint.sh
Bhargav-InfraCloud: Opened #1761

PR #1779, opened 2022-12-08
Fixed steps for postgresql example

shahpratikr:
## Change Overview

- fixed kanctl installation link
- updated responses for actionset

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/docker/postgres-tools-9.6/Dockerfile
+++ b/docker/postgres-tools-9.6/Dockerfile
@@ -1,6 +1,4 @@
-FROM postgres:14-bullseye
-
-LABEL maintainer="vkamra@kasten.io"
+FROM postgres:15-bullseye
 
 ENV DEBIAN_FRONTEND noninteractive
 
--- a/examples/postgresql/README.md
+++ b/examples/postgresql/README.md
@@ -10,10 +10,10 @@ Bitnami charts can be used with [Kubeapps](https://kubeapps.com/) for deployment
 
 ## Prerequisites
 
-- Kubernetes 1.10+
+- Kubernetes 1.20+
 - PV provisioner support in the underlying infrastructure
 - Kanister controller version 0.85.0 installed in your cluster
-- Kanctl CLI installed (https://docs.kanister.io/tooling.html#kanctl)
+- Kanctl CLI installed (https://docs.kanister.io/tooling.html#install-the-tools)
 
 ## Installing the Chart
 To install the chart with the release name `my-release`:
@@ -141,8 +141,8 @@ $ kanctl create actionset --action backup --namespace kanister --blueprint postg
 actionset backup-llfb8 created
 
 $ kubectl --namespace kanister get actionsets.cr.kanister.io
-NAME           AGE
-backup-glptq   38s
+NAME           PROGRESS   LAST TRANSITION TIME   STATE
+backup-glptq   100.00     2022-12-08T18:14:09Z   complete
 
 # View the status of the actionset
 $ kubectl --namespace kanister describe actionset backup-glptq
@@ -197,12 +197,22 @@ $ kanctl --namespace kanister create actionset --action restore --from backup-gl
 actionset restore-backup-glptq-6jzt4 created
 
 ## Check status
-$ kubectl --namespace kanister describe actionset restore-backup-glptq-6jzt4
+$ kubectl --namespace kanister get actionset restore-backup-glptq-6jzt4
+NAME                         PROGRESS   LAST TRANSITION TIME   STATE
+restore-backup-glptq-6jzt4   100.00     2022-12-08T18:16:50Z   complete
 ```
 
 Once the ActionSet status is set to "complete", you can see that the data has been successfully restored to PostgreSQL
 
 ```bash
+## Log in into postgresql container and get shell access
+$ kubectl exec -ti my-release-postgresql-0 -n postgres-test -- bash
+
+## use psql cli to add entries in postgresql database
+$ PGPASSWORD=${POSTGRES_PASSWORD} psql -U postgres
+psql (11.5)
+Type "help" for help.
+
 postgres=# \l
                                   List of databases
    Name    |  Owner   | Encoding |   Collate   |    Ctype    |   Access privileges
@@ -231,11 +241,13 @@ test=# select * from company;
 The artifacts created by the backup action can be cleaned up using the following command:
 
 ```bash
-$ kanctl --namespace kanister create actionset --action delete --from backup-glptq
+$ kanctl --namespace kanister create actionset --action delete --from backup-glptq --namespacetargets kanister
 actionset delete-backup-glptq-cq6bw created
 
 # View the status of the ActionSet
-$ kubectl --namespace kanister describe actionset delete-backup-glptq-cq6bw
+$ kubectl --namespace kanister get actionset delete-backup-glptq-cq6bw
+NAME                         PROGRESS   LAST TRANSITION TIME   STATE
+delete-backup-glptq-cq6bw    100.00     2022-12-08T18:16:50Z   complete
 ```
 
 ## Troubleshooting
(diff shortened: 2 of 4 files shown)


examples/postgresql/README.md
```diff
@@ -23,7 +23,7 @@ $ helm repo add bitnami https://charts.bitnami.com/bitnami
 $ helm repo update
 
 $ kubectl create ns postgres-test
-$ helm install my-release --namespace postgres-test bitnami/postgresql
+$ helm install my-release --namespace postgres-test bitnami/postgresql --version 11.9.13
```
pavannd1, line 26: Shouldn't we support the default version helm picks?

examples/postgresql/README.md
```diff
@@ -141,8 +141,8 @@ $ kanctl create actionset --action backup --namespace kanister --blueprint postg
 actionset backup-llfb8 created
 
 $ kubectl --namespace kanister get actionsets.cr.kanister.io
-NAME           AGE
-backup-glptq   38s
+NAME           PROGRESS   LAST TRANSITION TIME   STATE
+backup-glptq   10.00      2022-12-08T18:14:09Z   complete
```
pavannd1, line 145: The `PROGRESS` here should be marked as 100.00 instead of 10.00. I think @akankshakumari393 reported this issue as well. Could you please change this to `100.00`?

examples/postgresql/README.md
```diff
@@ -197,12 +197,22 @@ $ kanctl --namespace kanister create actionset --action restore --from backup-gl
 actionset restore-backup-glptq-6jzt4 created
 
 ## Check status
-$ kubectl --namespace kanister describe actionset restore-backup-glptq-6jzt4
+$ kubectl --namespace kanister get actionset restore-backup-glptq-6jzt4
+NAME                         PROGRESS   LAST TRANSITION TIME   STATE
+restore-backup-glptq-6jzt4   10.00      2022-12-08T18:16:50Z   complete
```
pavannd1, line 202: Same comment for the progress

examples/postgresql/README.md
```diff
@@ -235,7 +245,9 @@ $ kanctl --namespace kanister create actionset --action delete --from backup-glp
 actionset delete-backup-glptq-cq6bw created
 
 # View the status of the ActionSet
-$ kubectl --namespace kanister describe actionset delete-backup-glptq-cq6bw
+$ kubectl --namespace kanister get actionset delete-backup-glptq-cq6bw
+NAME                         PROGRESS   LAST TRANSITION TIME   STATE
+delete-backup-glptq-cq6bw    10.00      2022-12-08T18:16:50Z   complete
```
pavannd1, line 250: Same comment for the progress

examples/postgresql/README.md
```diff
@@ -264,13 +276,16 @@ $ helm delete my-release -n postgres-test
 ```
 
 ### Delete CRs
-Remove Blueprint and Profile CR
+Remove Blueprint, Profile CR and Actionset
```
pavannd1, line 279:
```suggestion
Remove Blueprint, Profile CR and ActionSet
```

examples/postgresql/postgres-blueprint.yaml
```diff
@@ -70,7 +70,7 @@ actions:
       name: deleteDump
       args:
         image: ghcr.io/kanisterio/postgres-kanister-tools:0.84.0
-        namespace: "{{ .Namespace.Name }}"
+        namespace: "{{ .StatefulSet.Namespace }}"
```
pavannd1, line 73: Why is this required? Note that downstream [adopters](https://github.com/kanisterio/kanister/blob/master/ADOPTERS.md) of Kanister use Namespace as the subject when executing `delete` action and this change will break them. I would recommend reverting this

examples/postgresql/v10.16.2/postgres-blueprint.yaml
```diff
@@ -70,7 +70,7 @@ actions:
       name: deleteDump
       args:
         image: ghcr.io/kanisterio/postgres-kanister-tools:0.84.0
-        namespace: "{{ .Namespace.Name }}"
+        namespace: "{{ .StatefulSet.Namespace }}"
```
pavannd1, line 73: Same as above

Also, I believe this blueprint is no longer relevant. We can remove the v10.16.2 and v8.6.4 folders completely. @viveksinghggits @PrasadG193 WDYT?

examples/postgresql/README.md
```diff
@@ -13,7 +13,7 @@ Bitnami charts can be used with [Kubeapps](https://kubeapps.com/) for deployment
 - Kubernetes 1.10+
```
pavannd1, line 13:
```suggestion
- Kubernetes 1.20+
```
Let's bump this up since we don't really support older versions

examples/postgresql/postgres-blueprint.yaml
shahpratikr: I was getting nil pointer reference for `Namespace.Name`. if downstream projects are using `Namespace.Name`, it should be consistent across phases right? we only have `Namespace.Name` for delete phase here but other 2 phases have `StatefulSet.Namespace`

examples/postgresql/README.md
shahpratikr: fixed. updated dockerfile

examples/postgresql/README.md
shahpratikr: changed

examples/postgresql/README.md
shahpratikr: changed

examples/postgresql/README.md
shahpratikr: changed

examples/postgresql/README.md
shahpratikr: fixed

examples/postgresql/postgres-blueprint.yaml
shahpratikr: reverted the change

examples/postgresql/v10.16.2/postgres-blueprint.yaml
shahpratikr: reverted the change

examples/postgresql/README.md
shahpratikr: fixed

examples/postgresql/v10.16.2/postgres-blueprint.yaml
viveksinghggits: sorry, maybe I am missing some context. Why this blueprint not relevant?

pavannd1: @viveksinghggits are there any specific uses for this blueprint? Do we need to support this particular version of postgres?

docker/postgres-tools-9.6/Dockerfile
```diff
@@ -1,6 +1,4 @@
-FROM postgres:14-bullseye
-
-LABEL maintainer="vkamra@kasten.io"
+FROM postgres:15-bullseye
```
pavannd1, line 1: As a follow-up, we should remove this Dockerfile and stop building this image through goreleaser. I don't think we use this anymore

pavannd1 (approved):

examples/postgresql/v10.16.2/postgres-blueprint.yaml
viveksinghggits: I don't think we have to support that specific version. But I think it's ok to keep the stale versions of blueprint just for future reference. if someone wants to refer to specific postgres' versions blueprint.
I think this is the reason we decided to have these dirs with older blueprints right?
I am also ok if decide to remove them.

PR #1798, opened 2022-12-16
Add a field in actionset status to figure out currently running phase

viveksinghggits:
## Change Overview

This PR adds another field in the actionset status that can be used to figure out the currently running phase of the actionset set.
And also adds that field in the `additionalPrinterColumns` so that we can get the phase that is currently running in the kubectl output itself.

```
Every 2.0s: kubectl get actionsets.cr.kanister.io -n kanister                         vivek: Mon Dec 26 19:48:30 2022

NAME           PROGRESS   RUNNINGPHASE   LAST TRANSITION TIME   STATE
backup-mqw49   66.67                     2022-12-26T18:47:32Z   complete
backup-tcqb7   66.67                     2022-12-26T18:47:06Z   complete
backup-vqzq8   33.33                     2022-12-26T18:48:16Z   failed
backup-zxxwd   33.33                     2022-12-26T18:48:00Z   failed
```

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/controller/controller_test.go
```diff
@@ -664,13 +665,32 @@ func (s *ControllerSuite) TestDeferPhase(c *C) {
 	err = s.waitOnActionSetState(c, as, crv1alpha1.StateRunning)
 	c.Assert(err, IsNil)
 
+	onPhases := sets.NewString()
```
viveksinghggits, line 668: I don't think we need set, here. I will update this to just have slice of strings.

pkg/apis/cr/v1alpha1/types.go
```diff
@@ -134,6 +134,8 @@ type ActionStatus struct {
 
 // ActionProgress provides information on the progress of an action.
 type ActionProgress struct {
+	// OnPhae represents which phase of the action is being run
```
ankitjain235, line 137:
```suggestion
	// OnPhase represents which phase of the action is being run
```

pkg/apis/cr/v1alpha1/types.go
```diff
@@ -134,6 +134,8 @@ type ActionStatus struct {
...
+	OnPhase string `json:"onPhase,omitempty"`
```
ankitjain235, line 138: Should we name it CurrentPhase instead? Just my personal opinion. OnPhase could be confused with some non-generic field.

pkg/phase.go
```diff
@@ -47,7 +49,17 @@ func (p *Phase) Objects() map[string]crv1alpha1.ObjectReference {
 
 // Exec renders the argument templates in this Phase's Func and executes with
 // those arguments.
-func (p *Phase) Exec(ctx context.Context, bp crv1alpha1.Blueprint, action string, tp param.TemplateParams) (map[string]interface{}, error) {
+func (p *Phase) Exec(ctx context.Context, cli crclientv1alpha1.CrV1alpha1Interface, as *crv1alpha1.ActionSet, bp crv1alpha1.Blueprint, action string, tp param.TemplateParams) (map[string]interface{}, error) {
```
ankitjain235, line 52: Should these be part of Phase struct instead? Could be reusable later if needed for other functions as well.

pkg/customresource/actionset.yaml
```diff
@@ -266,6 +268,10 @@ spec:
       type: string
       description: Progress of completion in percentage
       jsonPath: .status.progress.percentCompleted
+    - name: OnPhase
+      type: string
+      description: The phase that is being run, if the actionset is completed this would be last phase of action
```
ankitjain235, line 273: Wouldn't it be confusing if onPhase is still on last phase after completion? Should it give a final value instead?

pkg/apis/cr/v1alpha1/types.go
viveksinghggits: yeah, I spent sometime trying to come up with a better name but was not able to. I didn't chose `CurrentPhase` because current is in the name and the value is not actually the current phase, value is the phase that was run recently or phase that was run last. I am open to change this btw.
I also considered `runningphase` but that is also a bit confusing.

pkg/phase.go
viveksinghggits: mmm, I don't think we have a lot of methods of phase that is going to need the actionset. Also, actionset is not a property of phase, that is the reason I decided not to include that in the Phase struct.
So, if we add this as field of Phase, everywhere we create phase, we will have to either somehow try to init actionset or init it as `nil`. And we wont have actionset everywhere we init phase.

pkg/customresource/actionset.yaml
viveksinghggits: We can consider changing this to, let's say, completed once the actionset is completed. But I don't think that is going to be useful because we already specify at other places that the actionset is completed.
If we document it even once, anywhere, I think we would be good. But if we think having completed would be better, I can do that as well.

pkg/apis/cr/v1alpha1/types.go
pavannd1: Not a fan of `OnPhase` either. What if we name it `LastCompletedPhase` if we are displaying the recently completed one?

pkg/phase.go
ankitjain235: So it can be nil by default wherever we won't have actionset. We right now anyway have to pass `nil` while making `phase.Exec` call.
But also makes sense since ActionSet is not a property of field.
We should still move cli to phase struct right?

pkg/customresource/actionset.yaml
ankitjain235: If we decide to keep onPhase as the field, then we should either not set anything to the field once it's completed, or set `Completed`. Otherwise at one point we would say `onPhase` `xxx` but actionset status would be completed.

pkg/phase.go
viveksinghggits: Hi @ankitjain235,
Like we discussed offline I think it would be a good idea if we update the actionset right before calling `Exec` for phase.

pkg/apis/cr/v1alpha1/types.go
viveksinghggits: We are not actually displaying the recently completed one, we are displaying the currently running phase but we didn't name is `runningPhase` because when actionset completed it pointed to last run phase in the actionset.
I think the best option is to name it `runningPhase` and then change this to `completed` once actionset is completed.

pavannd1: I don't think we should change the field to `failed` or `completed`. The `STATE` field already provides that information. How about setting it to `""` after completion or failure?

viveksinghggits:
```
Every 2.0s: kubectl get actionsets.cr.kanister.io -n kanister                         vivek: Mon Dec 26 19:48:30 2022

NAME           PROGRESS   RUNNINGPHASE   LAST TRANSITION TIME   STATE
backup-mqw49   66.67                     2022-12-26T18:47:32Z   complete
backup-tcqb7   66.67                     2022-12-26T18:47:06Z   complete
backup-vqzq8   33.33                     2022-12-26T18:48:16Z   failed
backup-zxxwd   33.33                     2022-12-26T18:48:00Z   failed

```

pkg/controller/controller.go
```diff
@@ -516,6 +522,19 @@ func (c *Controller) runAction(ctx context.Context, t *tomb.Tomb, as *crv1alpha1
...
+// updateActionSetRunningPhase updates the actionset's `status.Progress.RunningPhase` with the phase name
+// that is being run currently. It doesn't fail if there was a problem updating the actionset. It just logs
+// the failure.
+func (c *Controller) updateActionSetRunningPhase(ctx context.Context, as *crv1alpha1.ActionSet, phase string) {
+	err := reconcile.ActionSet(ctx, c.crClient.CrV1alpha1(), as.Namespace, as.Name, func(as *crv1alpha1.ActionSet) error {
+		as.Status.Progress.RunningPhase = phase
+		return nil
+	})
+	if err != nil {
+		log.Error().WithError(fmt.Errorf("updating actionset with running phase: %s", err.Error()))
```
PrasadG193, line 534:
```suggestion
		log.Error().WithError(err).Print("failed to update actionset with running phase")
```

pkg/controller/controller_test.go
```diff
@@ -664,13 +665,36 @@ func (s *ControllerSuite) TestDeferPhase(c *C) {
...
+
 	// make sure deferPhase is also run successfully
 	err = s.waitOnDeferPhaseState(c, as, crv1alpha1.StateComplete)
 	c.Assert(err, IsNil)
 
 	err = s.waitOnActionSetState(c, as, crv1alpha1.StateComplete)
 	c.Assert(err, IsNil)
 
+	// once actionset is completed make sure all the phases are in onPhases
+	// that would confirm that the actionset's status.progress.onPhase was set
```
PrasadG193, line 690: typo?

```suggestion
	// that would confirm that the actionset's status.progress.runningPhase was set
```

pkg/customresource/actionset.yaml
```diff
@@ -266,6 +268,10 @@ spec:
       type: string
       description: Progress of completion in percentage
       jsonPath: .status.progress.percentCompleted
+    - name: RunningPhase
+      type: string
+      description: The phase that is being run, if the actionset is completed it would be set to ""
```
PrasadG193, line 273:
```suggestion
      description: The phase that is being run, if the actionset is failed or completed it would be set to ""
```

PrasadG193 (commented):
Have some minor suggestions. Otherwise looks good

pkg/customresource/actionset.yaml
viveksinghggits: I knowingly put completed there, because it would cover failed or success both. Other wide we can write 
```
actionset is failed or succeeded it would be set to ""
```

PrasadG193: makes sense

PrasadG193 (commented):
LGTM

PrasadG193 (approved):

docs/tutorial.rst
```diff
@@ -168,6 +168,13 @@ look at the updated status of the ActionSet and tail the controller logs.
   # check the controller log
   $ kubectl --namespace kanister get pod -l app=kanister-operator
 
+ActionSet's status can also be used to figure out the currently running phase
+of that ActionSet. In the cases the action that an ActionSet is
+performing has several phases, the ActionSet's ``Status.Progress.RunningPhase``
+can be used to figure out the phase that is being run currently. Once the
+ActionSet has completed the value of ``Status.Progress.RunningPhase`` would be
+set to ``""``.
+
```
akankshakumari393, line 177:
```suggestion
ActionSet's status can also be used to figure out the currently running phase
of that ActionSet. In cases where the action that an ActionSet performs has several
phases, the ActionSet's Status.Progress.RunningPhase can be used to determine
the current phase being run. Once the ActionSet has completed the value
of ``Status.Progress.RunningPhase`` would be set to ``""``.

```

pkg/controller/controller_test.go
```diff
@@ -664,13 +665,36 @@ func (s *ControllerSuite) TestDeferPhase(c *C) {
...
 	err = s.waitOnActionSetState(c, as, crv1alpha1.StateComplete)
 	c.Assert(err, IsNil)
 
+	// once actionset is completed make sure all the phases are in runningPhases
+	// that would confirm that the actionset's status.progress.runningPhase was set
+	// to those phases
+	op := sets.NewString()
+	// these phases are from blueprint that we create above
+	// and after actionset completion the runningPhase becomes ""
+	op.Insert("backupPhaseOne").Insert("backupPhaseTwo").Insert("deferPhase")
```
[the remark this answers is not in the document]
viveksinghggits: I think didn't understand it well. Even if we capture the running phase in `waitOnActionSetState` we will do the same right? We would get an actionset and see the running phase. If the phases changes from a to b and then from b to c again within one `get` call, we would not get `b` in our test but this would be the, I think, even if we put this in `waitOnActionSetState`.
What do you think?

viveksinghggits:
CI failed once because of 
```
FAIL: <autogenerated>:1: PostgreSQL.TestRun

integration_test.go:220:
    c.Assert(err, IsNil)
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc000f722e8), Stderr:[]uint8(nil)} ("signal: killed")
```

It seems to be a resource issue. I have re-triggered the CI, if we see this again, we can talk to Github Actions team to know more about the machines where these tests run and what we can do about those.

pkg/controller/controller_test.go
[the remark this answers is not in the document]
pavannd1: That's correct. Capturing it in `waitOnActionSetState` doesn't guarantee a successful run. We have faced a race in that function too whenever the execution was fast. It would timeout waiting for the `running` state while the action set had already reached the `completed` state.
But since we are already getting the action set to check status, the failure chances would be the same as the state failure.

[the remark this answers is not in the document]
viveksinghggits: mmm, yeah. I think we can somehow achieve that. 
But I think it's not that straight forward. For example if we try store the running phases in this line
```
err = s.waitOnActionSetState(c, as, crv1alpha1.StateRunning)
```
we just wait for the action to be running. And after that we return. So we will not have all the phases stored in the `runningPhase` var.

If we are talking about storing the phases in this line 
```
err = s.waitOnActionSetState(c, as, crv1alpha1.StateComplete)
```
In that case there are chances that some of the phases are already run so we will not be able to capture those phases.

Should I try this?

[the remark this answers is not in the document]
pavannd1: Yeah, that seems problematic too. Let's merge and see if it flakes 🚀

pavannd1 (approved):

pkg/controller/controller_test.go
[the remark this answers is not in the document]
viveksinghggits: thank you.

PR #1818, opened 2022-12-26
Add `--egress-selector-mode` flag while starting k3d

viveksinghggits:
## Change Overview

Kanister CI tests were failing very frequently with below error while execing a command into a pod

```
error dialing backend: EOF
```

We are not able to reproduce this, and did look like a networking issue to me, because this was failing even before initiating the pod deletion. Which means, it's not like the pod is not in the correct state and that is the reason we see this failure.

By looking further I got to this issue https://github.com/k3s-io/k3s/issues/5633 which seems to be very closely related to our failure. This commit incorporates the suggested change in our code.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -43,7 +43,7 @@ jobs:
         args: >-
           --agents 3
           --no-lb
-          --k3s-arg "--no-deploy=traefik,--egress-selector-mode=agent,servicelb@server:*"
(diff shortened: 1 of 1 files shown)


pavannd1:
Moved to kind

PR #1821, opened 2022-12-27
Use kind to create cluster in CI, instead of k3d

viveksinghggits:
## Change Overview

This PR removes `k3d-action` from Kanister CI and  and uses `helm/kind-action@v1.4.0` action instead to create a Kubernetes cluster.
Using kind in CI for Kanister should be preferred because 
- It is easy to decide which Kubernetes version we want to run our tests against
- In case of k3d-action, since it depends on k3d and then k3d actually depends on k3s, it's a bit tedious to figure the dependency out. 

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -43,6 +43,12 @@ jobs:
         make install-csi-hostpath-driver
         make install-minio
       if: matrix.testSuite == 'integration-test' || matrix.testSuite == 'helm-test'
+    # A test (CRDSuite) that runs as part of `make test` requies atleast one CRD to
+    # be present on the cluster. That's why we are only installing csi-hostpath-driver
+    # before running `make test` to make sure some CRDs are installed in the cluster.
+    - run: |
+        make install-csi-hostpath-driver
+      if: matrix.testSuite == 'test'
     - run: make ${{ matrix.testSuite }}
   build:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -37,18 +37,12 @@ jobs:
...
-          --agents 3
-          --no-lb
-          --k3s-arg "--no-deploy=traefik,servicelb@server:*"
+    - uses: helm/kind-action@v1.4.0
     - run: tar -xvf ./src.tar.gz
     - run: |
         make install-csi-hostpath-driver
         make install-minio
-      if: matrix.testSuite == 'integration-test' || matrix.testSuite == 'helm-test'
+      if: matrix.testSuite == 'integration-test' || matrix.testSuite == 'helm-test' || matrix.testSuite == 'test'
```
pavannd1, line 45: I believe this change is not required - we don't need minio and hostpath driver for unit tests.
```suggestion
      if: matrix.testSuite == 'integration-test' || matrix.testSuite == 'helm-test'
```

pavannd1 (approved):

viveksinghggits:
Hi @pavannd1 ,
Kind cluster by default doesnt have any CRD but one of our tests needs atleast one crd that is the reason i was installing csi hostpath driver in all the tests. Test, helm test and integration test.
We can improve this a bit to jist install csi in case of kube test and not minio.

pavannd1 (approved):

PR #1835, opened 2023-01-03
Do not ignore permission errors in kopia restore command

TimonOmsk:
## Change Overview

By default kopia ignores all the errors about permissions (see https://github.com/kopia/kopia/blob/master/cli/command_restore.go#L144) 
It leads to hiding such errors and incorrectly restored apps without any notifications.

## Pull request type

- [x] :sunflower: Feature

## Test Plan
- [x] :zap: Unit test

### DIFF
--- a/pkg/kopia/command/restore.go
+++ b/pkg/kopia/command/restore.go
@@ -14,17 +14,20 @@
 
 package command
 
+import "strconv"
+
 type RestoreCommandArgs struct {
 	*CommandArgs
-	RootID     string
-	TargetPath string
+	RootID                 string
+	TargetPath             string
+	IgnorePermissionErrors bool
 }
 
 // Restore returns the kopia command for restoring root of a snapshot with given root ID
 func Restore(cmdArgs RestoreCommandArgs) []string {
 	args := commonArgs(cmdArgs.CommandArgs, false)
 	args = args.AppendLoggable(restoreSubCommand, cmdArgs.RootID, cmdArgs.TargetPath)
-	args = args.AppendLoggableKV(ignorePermissionsError, "false")
+	args = args.AppendLoggableKV(ignorePermissionsError, strconv.FormatBool(cmdArgs.IgnorePermissionErrors))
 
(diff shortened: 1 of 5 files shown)


pchistikov (commented):
I'd like @pavannd1 or someone else from his team to approve to make sure that they are aware of it and potential implications it can bring. 
Probably we should not hardcode it in kanister but propagate `--ignore-permission-errors` flag value from k10 to avoid any issues for external kanister users.

pkg/kopia/command/restore.go
```diff
@@ -24,6 +24,7 @@ type RestoreCommandArgs struct {
 func Restore(cmdArgs RestoreCommandArgs) []string {
 	args := commonArgs(cmdArgs.CommandArgs, false)
 	args = args.AppendLoggable(restoreSubCommand, cmdArgs.RootID, cmdArgs.TargetPath)
+	args = args.AppendLoggableKV(ignorePermissionsError, "false")
```
pavannd1, line 27: The default can be `false` but we should make this a part of `RestoreCommandArgs`. That way, as @pchistikov said, downstream users could configure it to a different value if required

pkg/kopia/command/snapshot.go
```diff
@@ -66,6 +66,7 @@ func SnapshotRestore(cmdArgs SnapshotRestoreCommandArgs) []string {
 	if cmdArgs.SparseRestore {
 		args = args.AppendLoggable(sparseFlag)
 	}
+	args = args.AppendLoggableKV(ignorePermissionsError, "false")
```
pavannd1, line 69: Same suggestion as above

TimonOmsk:
The only doubts I have is about using strconv in commands. Maybe it would be better to create AppendLoggableKVBool function and convert from bool to string in that function and not in command? @pavannd1 @pchistikov

pchistikov (approved):
Looks good to me

pavannd1 (approved):
LGTM. We can follow up with the improvement you suggested above 👍🏼

PR #1849, opened 2023-01-09
Format bool arguments in a way it expected by kopia's cmd lib.

TimonOmsk:
## Change Overview

Kopia uses Kingpin(https://github.com/alecthomas/kingpin) lib for handling cmd args. 
This lib does not allow to set bool var in k=v way. Instead it uses k\no-k format. 
It means that instead of --ingore-permission-errors=false we should use --no-ignore-permission-errors argument. 

From https://github.com/alecthomas/kingpin/blob/master/README.md#boolean-values 
> Boolean values are uniquely managed by Kingpin. Each boolean flag will have a negative complement: --\<name\> and --no-\<name\>.

## Pull request type

- [x] :bug: Bugfix

## Test Plan

- [x] :zap: Unit test

### DIFF
--- a/pkg/kopia/command/snapshot.go
+++ b/pkg/kopia/command/snapshot.go
@@ -64,7 +64,7 @@ type SnapshotRestoreCommandArgs struct {
 func SnapshotRestore(cmdArgs SnapshotRestoreCommandArgs) []string {
 	args := commonArgs(cmdArgs.CommandArgs, false)
 	args = args.AppendLoggable(snapshotSubCommand, restoreSubCommand, cmdArgs.SnapID, cmdArgs.TargetPath)
-	args = args.AppendLoggableKV(ignorePermissionsError, strconv.FormatBool(cmdArgs.IgnorePermissionErrors))
+	args = args.AppendLoggableBool(ignorePermissionsError, cmdArgs.IgnorePermissionErrors)
 	if cmdArgs.SparseRestore {
 		args = args.AppendLoggable(sparseFlag)
 	}
--- a/pkg/kopia/command/snapshot.go
+++ b/pkg/kopia/command/snapshot.go
@@ -64,7 +64,11 @@ type SnapshotRestoreCommandArgs struct {
 func SnapshotRestore(cmdArgs SnapshotRestoreCommandArgs) []string {
 	args := commonArgs(cmdArgs.CommandArgs, false)
 	args = args.AppendLoggable(snapshotSubCommand, restoreSubCommand, cmdArgs.SnapID, cmdArgs.TargetPath)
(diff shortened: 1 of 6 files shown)


pkg/logsafe/logsafe.go
```diff
@@ -69,6 +73,18 @@ func (c Cmd) AppendLoggable(vl ...string) Cmd {
...
 }
 
+// AppendLoggableBool appends a boolean value in a "value\no-value" format.
+// In case of "true" value it will append "key".
+// In case of "false" value it will append "no-key"
+func (c Cmd) AppendLoggableBool(k string, v bool) Cmd {
+	if v {
+		c = append(c, arg{key: k})
+	} else {
+		c = append(c, arg{key: fmt.Sprintf("%s%s", noboolpref, k[2:])})
```
pchistikov, line 83: I suggest changing the comment to the function to explicitly note that "--key" is changing to "--no-key". Or even better to add explicit check to the function that we have "--" in the beginning of the key.
Without a context, k[2:] looks confusing here.

pchistikov (approved):
LGTM

adowair (approved):

pkg/kopia/command/snapshot.go
```diff
@@ -64,7 +64,7 @@ type SnapshotRestoreCommandArgs struct {
 func SnapshotRestore(cmdArgs SnapshotRestoreCommandArgs) []string {
 	args := commonArgs(cmdArgs.CommandArgs, false)
 	args = args.AppendLoggable(snapshotSubCommand, restoreSubCommand, cmdArgs.SnapID, cmdArgs.TargetPath)
-	args = args.AppendLoggableKV(ignorePermissionsError, strconv.FormatBool(cmdArgs.IgnorePermissionErrors))
+	args = args.AppendLoggableBool(ignorePermissionsError, cmdArgs.IgnorePermissionErrors)
```
pavannd1, line 67: @TimonOmsk Wouldn't it be simpler if we do
```
if cmdArgs.IgnorePermissionErrors {
    args = args.AppendLoggable(ignorePermissionsError)
} else {
    args = args.AppendLoggable(noIgnorePermissionsError)
}
```
`noIgnorePermissionsError` is a constant with `--no-ignore-permission-errors` value

TimonOmsk: It's simpler, but we'd have to introduce 2 arguments for every single argument. I think it's up to you here. If you think it's easier I would do it :)

pchistikov: I like @pavannd1's proposal. It's simple and straightforward.

pchistikov (approved):

pavannd1 (approved):

PR #1883, opened 2023-02-02
Add conversion from s3Compliant type to s3 type

ankitjain235:
## Change Overview

This PR adds conversion from s3Compliant type to s3. For cases with kanister profiles, the profile type is still "s3Compliant". Since for kopia, the s3Compliant type is same as s3, this PR adds the necessary conversion in the utility.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kopia/command/storage/secret_utils.go
```diff
@@ -183,6 +184,9 @@ func GetMapForLocationValues(
 	}
 	if locType != "" {
 		m[typeKey] = []byte(locType)
+		if locType == LocType(v1alpha1.LocationTypeS3Compliant) {
```
PrasadG193, line 187:
```suggestion
		if loc, ok == LocType(v1alpha1.LocationTypeS3Compliant); ok && loc == locType {
```
Check if type assertion is successful to be on safer side. Otherwise it may panic at runtime of becomes incompatible in future

pkg/kopia/command/storage/secret_utils.go
```diff
@@ -183,6 +184,9 @@ func GetMapForLocationValues(
...
+			m[typeKey] = []byte(LocTypeS3)
+		}
```
viveksinghggits, line 189: what is the value of `locType` in case of the if statement? is it not LocTypeS3 already?

LGTM

pkg/kopia/command/storage/secret_utils.go
ankitjain235: Makes sense

pkg/kopia/command/storage/secret_utils.go
ankitjain235: For Kanister profiles, it's `s3Compliant` while we have `s3` as value of LocTypeS3.

pkg/kopia/command/storage/secret_utils.go
ankitjain235: I checked but type conversion here does not return two values here.
Type assertion works only on interface I think. Also compilation itself would fail if type of either(LocType or LocationTypeS3Compliant) changes so we shouldn't have an issue.

PrasadG193 (approved):

PR #1919, opened 2023-02-21
Add RepositoryServer ObjectReference in Actionset

r4rajat:
## Change Overview

Add RepositoryServer ObjectReference in ```kanctl``` command line to create various actions.
This PR adds support for passing `--repository-server` flag in `kanctl` command line

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

#### Build ```kanctl``` command line tool

```
go build -o kanctl cmd/kanctl/main.go
```

#### Run Actionset

```
./kanctl create actionset --action backup --namespace kanister --blueprint backupdate-bp --deployment time-logger/time-logger --repository-server=kopia-repo-server-1

actionset backup-czvrz created
```

Also there's a check which ensures that Repository Server CR is in ```ServerReady``` state before creating actionset. 

```
./kanctl create actionset --action backup --namespace kanister --blueprint backupdate-bp --deployment time-logger/time-logger --repository-server=kopia-repo-server-1

Please make sure that Repository Server CR 'kopia-repo-server-1' is in Ready State: Repository Server Not Ready
Error: resource verification failed
```


pkg/customresource/actionset.yaml
```diff
@@ -11,49 +11,78 @@ spec:
...
-      openAPIV3Schema:
-        properties:
-          apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
+    - name: v1alpha1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          properties:
```
viveksinghggits, line 19: Do we need changes in this file @r4rajat ?

r4rajat: Yes @viveksinghggits, as we're adding RepositoryServer field in actionset, so we've to modify it's customresource definition

pkg/customresource/actionset.yaml
```diff
@@ -11,49 +11,78 @@ spec:
...
-    served: true
-    storage: true
-    schema:
-      openAPIV3Schema:
-        properties:
-          apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
+    - name: v1alpha1
+      served: true
+      storage: true
```
viveksinghggits, line 16: We generate this yaml right?
Can we just update the new entry manually here?
We can generate the yaml but just copy the snipped for `RepositoryServer` and copy that here. We wont see all these unnecessary changes here.
Or if we are sure that this wouldn't result into any problem, which it wont I think, we can go ahead with this as well.

pkg/kanctl/actionset.go
```diff
@@ -92,6 +94,7 @@ func newActionSetCmd() *cobra.Command {
 	cmd.Flags().StringSliceP(deploymentFlagName, "d", []string{}, "deployment for the action set, comma separated namespace/name pairs (eg: --deployment namespace1/name1,namespace2/name2)")
 	cmd.Flags().StringSliceP(optionsFlagName, "o", []string{}, "specify options for the action set, comma separated key=value pairs (eg: --options key1=value1,key2=value2)")
 	cmd.Flags().StringP(profileFlagName, "p", "", "profile for the action set")
+	cmd.Flags().StringP(repositoryServerFlagName, "r", "", "kopia repository server custom resource reference")
```
viveksinghggits, line 97: Let's also specify the format here.
For, like its specified for other flags.
```
eg: --deployment namespace1/name1,namespace2/name2
```

pkg/kanctl/actionset.go
```diff
@@ -342,6 +355,25 @@ func parseProfile(cmd *cobra.Command, ns string) (*crv1alpha1.ObjectReference, e
...
 
+func parseRepositoryServer(cmd *cobra.Command, ns string) (*crv1alpha1.ObjectReference, error) {
+	repositoryServerName, _ := cmd.Flags().GetString(repositoryServerFlagName)
+	if repositoryServerName == "" {
+		return nil, nil
+	}
+	if strings.Contains(repositoryServerName, "/") {
+		temp := strings.Split(repositoryServerName, "/")
+		if len(temp) != 2 {
+			return nil, errors.Errorf("Invalid repository server name %s, should be of the form ( --repository-server namespace/name OR --repository-server name)", repositoryServerName)
```
viveksinghggits, line 366:
```suggestion
			return nil, errors.Errorf("Invalid repository server name %s, it should be of the form ( --repository-server namespace/name or --repository-server name)", repositoryServerName)
```

pkg/kanctl/actionset.go
```diff
@@ -342,6 +355,25 @@ func parseProfile(cmd *cobra.Command, ns string) (*crv1alpha1.ObjectReference, e
 	}, nil
 }
 
+func parseRepositoryServer(cmd *cobra.Command, ns string) (*crv1alpha1.ObjectReference, error) {
```
viveksinghggits, line 358: We don't  have to do anything here, but I would rather prefer `objectRefFromRepoServer` instead of `parseRepositoryServer`.
But I understand is ok to keep it as it is for consistency. Maybe just add a comment that it returns object reference for passed repo server.

pkg/kanctl/actionset.go
```diff
@@ -631,6 +664,21 @@ func verifyParams(ctx context.Context, p *PerformParams, cli kubernetes.Interfac
 		}
 	}()
 
+	// RepositoryServer
+	go func() {
+		defer wg.Done()
+		if p.RepositoryServer != nil {
+			rs, err := crCli.CrV1alpha1().RepositoryServers(p.RepositoryServer.Namespace).Get(ctx, p.RepositoryServer.Name, metav1.GetOptions{})
+			if err != nil {
```
viveksinghggits, line 672: Should we check if the error actually says that the resource is not found. So we would have three categories
- Resource is not found
- There was an error getting the resource
- Resource is found and not ready

You don't have to do it, let's  wait for others opinions.

pkg/param/param.go
```diff
@@ -299,6 +350,14 @@ func fetchSecretCredential(ctx context.Context, cli kubernetes.Interface, sr *cr
 	}, nil
 }
 
+func fetchSecretFromSecretRef(ctx context.Context, cli kubernetes.Interface, ref v1.SecretReference) (*v1.Secret, error) {
```
viveksinghggits, line 353:
```suggestion
func secretFromSecretRef(ctx context.Context, cli kubernetes.Interface, ref v1.SecretReference) (*v1.Secret, error) {
```

pkg/kanctl/actionset.go
r4rajat: Sure wil add that

pkg/kanctl/actionset.go
r4rajat: - So the checks for first scenerio, ```Resource is not found``` and the third scenerio ```Resource is found and not ready``` has been handled while creating actionset, we make sure Resources are Present and Resources are in ready state
- And the second scenerio, ```There was an error getting the resource``` is handled in the code above

pkg/customresource/actionset.yaml
r4rajat: Sure will try that !

pkg/kanctl/actionset.go
r4rajat: Added the comment !

pkg/kanctl/actionset.go
r4rajat: Will make changes

pkg/kanctl/actionset.go
```diff
@@ -342,6 +355,27 @@ func parseProfile(cmd *cobra.Command, ns string) (*crv1alpha1.ObjectReference, e
...
 
+// parseRepositoryServer returns the object reference
+// for the passed Repository Server
+func parseRepositoryServer(cmd *cobra.Command, ns string) (*crv1alpha1.ObjectReference, error) {
+	repositoryServerName, _ := cmd.Flags().GetString(repositoryServerFlagName)
+	if repositoryServerName == "" {
+		return nil, nil
+	}
+	if strings.Contains(repositoryServerName, "/") {
+		temp := strings.Split(repositoryServerName, "/")
```
viveksinghggits, line 366:
```suggestion
		nsName := strings.Split(repositoryServerName, "/")
```

pkg/kanctl/actionset.go
```diff
@@ -92,6 +94,7 @@ func newActionSetCmd() *cobra.Command {
 	cmd.Flags().StringSliceP(deploymentFlagName, "d", []string{}, "deployment for the action set, comma separated namespace/name pairs (eg: --deployment namespace1/name1,namespace2/name2)")
 	cmd.Flags().StringSliceP(optionsFlagName, "o", []string{}, "specify options for the action set, comma separated key=value pairs (eg: --options key1=value1,key2=value2)")
 	cmd.Flags().StringP(profileFlagName, "p", "", "profile for the action set")
+	cmd.Flags().StringP(repositoryServerFlagName, "r", "", "kopia repository server custom resource reference (eg: --repository-server namespace/name)")
```
viveksinghggits, line 97: After looking at the rest of the code it looks like `--repository-server name` is also allowed. In that case do we want  expect the repo server to be available in the controller namespace or in the app namespace. 
Also, instead of supporting both the formats we should just support `namespace/name` format, just to avoid confusion.

pkg/kanctl/actionset.go
```diff
@@ -727,3 +748,47 @@ func generateActionSetName(p *PerformParams) (string, error) {
 
 	return "", errMissingFieldActionName
 }
+
+func verifyRepositoryServerParams(repoServer *crv1alpha1.ObjectReference, crCli versioned.Interface, ctx context.Context) error {
+	if repoServer != nil {
+		rs, err := crCli.CrV1alpha1().RepositoryServers(repoServer.Namespace).Get(ctx, repoServer.Name, metav1.GetOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "Please make sure '%s' with name '%s' exists in namespace '%s'", "repository-server", repoServer.Name, repoServer.Namespace)
+		}
```
viveksinghggits, line 757:
```suggestion
		if err != nil {
		 if apierrors.IsNotFound(err){
			return errors.Wrapf(err, "Please make sure '%s' with name '%s' exists in namespace '%s'", "repository-server", repoServer.Name, repoServer.Namespace)
			}
			return "there was an error getting repo server"
		}
```

viveksinghggits (commented):
Almost, there.

pkg/kanctl/actionset.go
r4rajat: Yes @viveksinghggits, it is allowed but we want the repository server CR in controller namespace only

pkg/kanctl/actionset.go
r4rajat: Implemented !

pkg/kanctl/actionset.go
```diff
@@ -342,6 +356,28 @@ func parseProfile(cmd *cobra.Command, ns string) (*crv1alpha1.ObjectReference, e
 	}, nil
 }
 
+// parseRepositoryServer returns the object reference
+// for the passed Repository Server
+func parseRepositoryServer(cmd *cobra.Command) (*crv1alpha1.ObjectReference, error) {
+	repositoryServerName, _ := cmd.Flags().GetString(repositoryServerFlagName)
+	if repositoryServerName == "" {
+		return nil, nil
```
akankshakumari393, line 364: should we return an error here or an error message describing the issue instead of returning nil?

pkg/kanctl/actionset.go
```diff
@@ -727,3 +750,50 @@ func generateActionSetName(p *PerformParams) (string, error) {
...
 }
+
+func verifyRepositoryServerParams(repoServer *crv1alpha1.ObjectReference, crCli versioned.Interface, ctx context.Context) error {
+	if repoServer != nil {
+		rs, err := crCli.CrV1alpha1().RepositoryServers(repoServer.Namespace).Get(ctx, repoServer.Name, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return errors.Wrapf(err, "Please make sure '%s' with name '%s' exists in namespace '%s'", "repository-server", repoServer.Name, repoServer.Namespace)
+			}
+			return errors.New("there was an error getting repo server")
```
akankshakumari393, line 761:
```suggestion
			return errors.New("error while fetching repo server")
```

pkg/kanctl/actionset.go
r4rajat: We're handling errors where we're calling this function. I think we're doing similar while fetching other CRs. Let's see what others say.

pkg/param/param.go
```diff
@@ -135,6 +137,21 @@ type KopiaServerCreds struct {
 	ConnectOptions map[string]int
 }
 
+// RepositoryServer contains where to store Repository Server CR Artifacts
```
viveksinghggits, line 140: This isn't necessarily true right?
We are going to store the repo server CRs in the etcd itself right?

pkg/kanctl/actionset.go
viveksinghggits: It's not about handling errors, even if we are handling errors where we are calling this function from; that would be of no use because we are not returning error.
Again, handling errors comes into picture when we return error.
One thing that I can think of is, we don't necessarily want to fail the `kanctl` command when repo server is not specified because there are chances that profile is specified. 
Can you please point me to the code where this check is performed to make sure either profile or repo server is  specified.

viveksinghggits (changes requested):

pkg/param/param.go
r4rajat: Yeah, so this RepositoryServer struct basically contains the selective artifacts from the Repo Server CR. It is not the Respo Server CR. It's only templating parameters.

pkg/kanctl/actionset.go
r4rajat: @viveksinghggits, While creating actionsets, we don't check/verify that either `--profile` or `--repository-server` flag is passed. in case either is passed, only then, it would verify the resources.

r4rajat: Added check, if `--profile` or `--repository-server` flag is not provided, `kanctl` will warn user but will create actionset

pkg/param/param.go
viveksinghggits: what are selective artifacts?

viveksinghggits, line 140:
```suggestion
// RepositoryServer contains fields from Repository server CR that will be used to resolve go templates for repository server in blueprint
```

viveksinghggits (approved):

PR #1922, opened 2023-02-21
Handle errors when waiting for pod status change fails

e-sumin:
## Change Overview

When executing `kubeTask` it may happen that pod won't get into desired state (ready or succeeded). In such case we might lost information about failure reason. This PR introduces simple pod log tail capturing mechanism.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kube/log_tail.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"bufio"
+	"bytes"
+	"strings"
+)
+
+// LogTail interface allows to store last N lines of log written to it
+type LogTail interface {
+	Write(p []byte) (int, error) // Write log line(s) to circular buffer
+	ToString() string            // Join stored log lines to one string separated by newline
```
denisvmedia, line 26: Can we have the comments above the functions so that they build up a proper godoc format.

denisvmedia (approved):

pkg/kube/pod.go
```diff
@@ -235,37 +236,68 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
...
+// getErrorWithLogs fetches logs from pod and constructs error containing last ten lines of log and specified error message
+func getErrorWithLogs(ctx context.Context, cli kubernetes.Interface, namespace, podName string, err error, errorMessage string) error {
+	r, logErr := StreamPodLogs(ctx, cli, namespace, podName)
+	if logErr != nil {
+		return errors.Wrapf(logErr, "Failed to fetch logs from the pod")
+	}
+	defer r.Close()
+
+	// Grab last log lines and put them to an error
+	lt := NewLogTail(10)
```
viveksinghggits, line 248: There are chances that the actual error message is not included in the last 10 lines right? Or we are assuming that it will always be?

e-sumin: I suppose that it should be there, but of course there is a chance that after logging an error, some other information will be logged, but frankly, I've never seen such scenario.

pkg/kube/log_tail.go
```diff
@@ -0,0 +1,77 @@
...
+	ToString() string
+}
+
+type logTail struct {
+	lines []string
+	idx   int
+	len   int
+}
+
+// NewLogTail creates logTail struct containing circular buffer for storing `len` last lines of written through Write method
```
pavannd1, line 35:
```suggestion
// NewLogTail creates logTail struct containing circular buffer for storing `len` last lines of log written through Write method
```

pkg/kube/pod.go
```diff
@@ -235,37 +236,68 @@ func GetPodLogs(ctx context.Context, cli kubernetes.Interface, namespace, name s
 	return string(bytes), nil
 }
 
+// getErrorWithLogs fetches logs from pod and constructs error containing last ten lines of log and specified error message
+func getErrorWithLogs(ctx context.Context, cli kubernetes.Interface, namespace, podName string, err error, errorMessage string) error {
```
pavannd1, line 240: Should this be reversed? `getLogsWithError`?

e-sumin: This function is called when we definitely know that something wrong happened, and we would like to embed log to error message, here is how I constructed the function name - "get error containing logs" => `getErrorWithLogs`. Is it wrong ?

pavannd1: Not a blocker and can be done in a follow up PR but how about `getErrorFromLogs`?

e-sumin: agree, thank you for suggestion !

pkg/kube/pod.go
e-sumin: Discussed offline. Keeping this solution since we don't have better options here.

e-sumin:
Since there are no more objections, merging.

PR #1957, opened 2023-03-14
Add flag to enable/disable public access for RDS instances created during Test

akankshakumari393:
## Change Overview

This PR adds flag to enable/disable public access for RDS instances created during Test. It is for both RDS postgres and aurora instances. By default it would be set to false.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/aws/rds/rds.go
```diff
@@ -73,15 +74,25 @@ func (r RDS) CreateDBCluster(ctx context.Context, storage int64, instanceClass,
...
+
 func (r RDS) CreateDBInstanceInCluster(ctx context.Context, restoredClusterID, instanceID, instanceClass, dbEngine, dbSubnetGroup string) (*rds.CreateDBInstanceOutput, error) {
-	pa := true
 	dbi := &rds.CreateDBInstanceInput{
 		DBClusterIdentifier:  &restoredClusterID,
 		DBInstanceClass:      &instanceClass,
 		DBInstanceIdentifier: &instanceID,
 		DBSubnetGroupName:    &dbSubnetGroup,
 		Engine:               &dbEngine,
-		PubliclyAccessible:   &pa,
```
PrasadG193, line 84: What will be the default value if not specified?

pkg/app/rds_aurora_mysql.go
```diff
@@ -205,7 +206,7 @@ func (a *RDSAuroraMySQLDB) Install(ctx context.Context, namespace string) error
 	}
 
 	// create db instance in the cluster
-	_, err = rdsCli.CreateDBInstanceInCluster(ctx, a.id, fmt.Sprintf("%s-instance-1", a.id), AuroraDBInstanceClass, string(function.DBEngineAuroraMySQL), a.dbSubnetGroup)
+	_, err = rdsCli.CreateDBInstanceInClusterForTest(ctx, a.id, fmt.Sprintf("%s-instance-1", a.id), AuroraDBInstanceClass, string(function.DBEngineAuroraMySQL), a.dbSubnetGroup, a.publicAccess)
```
viveksinghggits, line 209: I think we should explicitly set `a.publicAccess` to `false` to make it clear that this value that we are passing here is false.

pkg/aws/rds/rds.go
```diff
@@ -73,15 +74,25 @@ func (r RDS) CreateDBCluster(ctx context.Context, storage int64, instanceClass,
 	return r.CreateDBClusterWithContext(ctx, dbi)
 }
 
+func (r RDS) CreateDBInstanceInClusterForTest(ctx context.Context, restoredClusterID, instanceID, instanceClass, dbEngine, dbSubnetGroup string, publicAccess bool) (*rds.CreateDBInstanceOutput, error) {
```
viveksinghggits, line 77: Can you explain the difference between these three methods? I see CreateDBInstance, InCluster and InClusterForTest.

akankshakumari393: CreateDBInstance -> creates a DB instance, this is used for RDS postgres apps test
InCluster -> creates a DB instance using cluster ID, this is required when we spin up instance during restore, so in argument it also accepts a clusterID

InClusterForTest -> same as above(2) but this has `publiclyAccessible` field set to false.

pkg/aws/rds/rds.go
akankshakumari393:
```
	// If DBSubnetGroupName isn't specified, and PubliclyAccessible isn't specified,
	// the following applies:
	//
	//    * If the default VPC in the target Region doesn’t have an internet gateway
	//    attached to it, the DB cluster is private.
	//
	//    * If the default VPC in the target Region has an internet gateway attached
	//    to it, the DB cluster is public.
	//
	// If DBSubnetGroupName is specified, and PubliclyAccessible isn't specified,
	// the following applies:
	//
	//    * If the subnets are part of a VPC that doesn’t have an internet gateway
	//    attached to it, the DB cluster is private.
	//
	//    * If the subnets are part of a VPC that has an internet gateway attached
	//    to it, the DB cluster is public.
	//
	// Valid for: Multi-AZ DB clusters only
```
this is what the docs says, now we would be falling in the second case i.e we are providing DBsubnetGroupName, so it depends on the subnet group

pkg/app/rds_aurora_mysql.go
viveksinghggits: Sorry, I didn't understand clearly.
So now if we explicitly set the value to be `false`, are you saying we will not be able to change that later? Because that is already happening according to me, the value is already set to false, I suggested that we should make that explicit so that someone who reads the code would understand that by default its `false`.

viveksinghggits (changes requested):

pkg/aws/rds/rds.go
```diff
@@ -48,15 +48,40 @@ func NewClient(ctx context.Context, awsConfig *aws.Config, region string) (*RDS,
...
+		dbi = &rds.CreateDBInstanceInput{
+			AllocatedStorage:     storage,
+			DBInstanceIdentifier: &instanceID,
+			VpcSecurityGroupIds:  convertSGIDs(sgIDs),
+			DBInstanceClass:      &instanceClass,
+			Engine:               &engine,
+			MasterUsername:       &username,
+			MasterUserPassword:   &password,
+			PubliclyAccessible:   publicAccess,
+		}
```
viveksinghggits, line 84: can we initialise the struct with common fields, outside of the switch and then in case just set the fields that are specific to that case?

viveksinghggits: Yes, init the struct with *common* fields before switch and then in case statements just add fields respective to that case.

akankshakumari393: Yeah did that. PTAL.

PR #1979, opened 2023-03-30
Rearrange order of snapshot APIs

shahpratikr:
## Change Overview

This PR rearranges the order of Snapshot APIs in v1 (stable), v1beta1 and v1alpha1.
This fixes taking snapshot when RKE2 snapshot-validation-hook is enabled.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E

- Manually applied this fix in a cluster and was able to take a snapshot of a namespace when RKE2 webhook was enabled.
- Triggered `staging-gke` pipeline with the changes. All E2E tests passed.


pavannd1 (commented):
Since v1 is the default version in the newer clusters, this change makes sense 👍🏼

PrasadG193:
Need to perform upgrade testing. E,g If Snapshot taken from older workflow can be restored by newer workflow. Please add detailed test plans in the PR discription. Thanks

viveksinghggits:
> 

@PrasadG193 we should also make sure that the snapshot related api version is different between these two tests right?

PrasadG193:
Yes, we need to test for backward compatibility. If snapshots taken with beta apis can be restored with alpha.

shahpratikr:
I observed that snapshots are taken with v1 APIs in minimum supported k8s version.

PrasadG193 (approved):

PR #1986, opened 2023-04-04
Refactor PodRunner and KubeTask to use PodController under the hood.

e-sumin:
## Change Overview

Here is a PodRunner refactoring using previously added PodController and PodCommandExecutor.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod_runner.go
+++ b/pkg/kube/pod_runner.go
@@ -32,7 +32,7 @@ type PodRunnerFunc func(context.Context, PodController) (PodOutputMap, error)
 
 // PodRunner allows us to start / stop pod, write file to pod and execute command within it
 type PodRunner interface {
-	Run(ctx context.Context, fn func(context.Context, *v1.Pod) (map[string]interface{}, error)) (map[string]interface{}, error)
+	Run(ctx context.Context, fn func(context.Context, *v1.Pod) (map[string]interface{}, error)) (PodOutputMap, error)
 	// RunEx utilizes the PodController interface and forwards it to the functor, simplifying the manipulation with
 	// particular pod from the functor.
 	// TODO: Since significant number of functions are currently using PodRunner, we'll keep Run for now.
@@ -54,7 +54,7 @@ func NewPodRunner(cli kubernetes.Interface, options *PodOptions) PodRunner {
 }
 
 // Run will create a new Pod based on PodRunner contents and execute the given function
-func (p *podRunner) Run(ctx context.Context, fn func(context.Context, *v1.Pod) (map[string]interface{}, error)) (map[string]interface{}, error) {
+func (p *podRunner) Run(ctx context.Context, fn func(context.Context, *v1.Pod) (map[string]interface{}, error)) (PodOutputMap, error) {
 	return p.RunEx(ctx, func(innerCtx context.Context, pc PodController) (PodOutputMap, error) {
 		return fn(innerCtx, pc.Pod())
 	})
--- a/pkg/kube/pod_runner.go
+++ b/pkg/kube/pod_runner.go
@@ -33,6 +33,11 @@ type PodRunnerFunc func(context.Context, PodController) (PodOutputMap, error)
 // PodRunner allows us to start / stop pod, write file to pod and execute command within it
 type PodRunner interface {
 	Run(ctx context.Context, fn func(context.Context, *v1.Pod) (map[string]interface{}, error)) (map[string]interface{}, error)
+	// RunEx utilizes the PodController interface and forwards it to the functor, simplifying the manipulation with
+	// particular pod from the functor.
+	// TODO: Since significant number of functions are currently using PodRunner, we'll keep Run for now.
+	// However, once all these functions have been refactored to use PodController,
+	// Run should be removed and RunEx has to be renamed to Run.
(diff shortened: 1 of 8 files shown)


denisvmedia:
@e-sumin could you please fix the linter issues?

pkg/kube/pod_runner.go
```diff
@@ -66,3 +71,26 @@ func (p *podRunner) Run(ctx context.Context, fn func(context.Context, *v1.Pod) (
 	}()
 	return fn(ctx, pod)
 }
+
+// RunEx will create a new Pod based on PodRunner contents and execute the given function
+func (p *podRunner) RunEx(ctx context.Context, fn PodRunnerFunc) (PodOutputMap, error) {
```
denisvmedia, line 76: It looks like this function is mostly copy-paste for Run. Can we reduce code duplication?

e-sumin: the problem with the Run is that even though it already uses a PodController, it still calls a functor with the pod's name, and the functor works with the pod directly, not through the pod controller.
the RunEx was an attempt to move all users of the function step by step to use under the controller.

denisvmedia: But I still see that they share 99% of each other. Can we reduce code duplication?

e-sumin: Done.

pkg/kube/pod_runner.go
```diff
@@ -45,11 +50,17 @@ func NewPodRunner(cli kubernetes.Interface, options *PodOptions) PodRunner {
 
 // Run will create a new Pod based on PodRunner contents and execute the given function
 func (p *podRunner) Run(ctx context.Context, fn func(context.Context, *v1.Pod) (map[string]interface{}, error)) (map[string]interface{}, error) {
+	return p.RunEx(ctx, func(ctx2 context.Context, pc PodController) (PodOutputMap, error) {
+		return fn(ctx2, pc.Pod())
```
denisvmedia, line 54: Nit. `ctx2` looks weird to me. I'd rename it to something more meaningful, e.g.:

```suggestion
	return p.RunEx(ctx, func(innerCtx context.Context, pc PodController) (PodOutputMap, error) {
		return fn(innerCtx, pc.Pod())
```

pkg/kube/pod_runner.go
```diff
@@ -26,9 +26,14 @@ import (
...
 )
 
+type PodOutputMap map[string]interface{}
+
+type PodRunnerFunc func(context.Context, PodController) (PodOutputMap, error)
+
 // PodRunner allows us to start / stop pod, write file to pod and execute command within it
 type PodRunner interface {
 	Run(ctx context.Context, fn func(context.Context, *v1.Pod) (map[string]interface{}, error)) (map[string]interface{}, error)
+	RunEx(ctx context.Context, fn PodRunnerFunc) (PodOutputMap, error)
```
denisvmedia, line 36: Nit. I think we should either make named function types for both - Run and RunEx, or for none. The same is for the returned map type.

viveksinghggits:
@e-sumin,
There are some other kanister functions (for example `prepare_data.go`) that use podRunner to create and execute pods in them. Are we planning to handle those in other/separate PRs?

pkg/kube/pod_runner.go
viveksinghggits: I think we should add a comment where on why we have to methods that are doing to the same things. Basically what you have explained in the comment below.
Also, I think `RunEx` is not very intuitive, do we have other alternate names :) ?

viveksinghggits (commented):
LGTM overall.

pkg/kube/pod_runner.go
e-sumin: I've added the comment, but don't have better name. If you have something in mind, please suggest.

e-sumin: @denisvmedia I made returned type similar for both functions, but was unable to invent name for functor type, so I left it as is.

e-sumin:
@viveksinghggits yep, there are several places where pod runner is used. I think it's better to refactor them in separate PRs, so that the reviewing process will be faster.

pkg/kube/pod_runner.go
denisvmedia: If no one has better names I suggest these:

- `PodAwareRunFunc`
- `PodControllerAwareRunFunc`

denisvmedia:
@e-sumin please fix the signatures in the interface before merging. I put `hold off merging` to prevent the merge. You can remove it as soon as you sort out the issue.

pkg/kube/pod_runner.go
viveksinghggits: I am assuming that we are going to get rid of `Run` method eventually. And then we can rename `RunEx` to `Run`.
I am ok with `RunEx` as well.

viveksinghggits (approved):
Approving, we can merge after addressing Denis' interface signature comment.

PR #1988, opened 2023-04-05
Upgrade openshift/api to release-4.15 branch

julio-lopez:
https://github.com/openshift/api/tree/release-4.15
https://github.com/openshift/client-go/tree/release-4.15

## Pull request type

- [x] Dependency upgrade

## Test Plan

- [x] :zap: CI


julio-lopez:
#1987

julio-lopez:
@viveksinghggits @pavannd1 Please DO NOT merge yet, it is not complete. I'd like to include an upgrade for Openshift's client-go module as well in the same PR.
Nevertheless, PTAL and LMK if there are any concerns.

Also, what is the plan for upgrading Openshift going forward if automatic upgrades don't work reliably?

julio-lopez:
Also, this includes a couple of minor upgrades to k8s dependencies. Those can be pushed separately in case we need to revert this change.

go.mod
```diff
@@ -40,8 +40,8 @@ require (
 	github.com/onsi/gomega v1.24.1
 
 	//pinned openshift to release-4.5 branch
```
viveksinghggits, line 42:
```suggestion
	//pinned openshift to release-4.15 branch
```

go.mod
```diff
@@ -54,13 +54,13 @@ require (
 	gopkg.in/tomb.v2 v2.0.0-20161208151619-d5d1b5820637
 
 	//pinned k8s.io to v0.26.1 tag
-	k8s.io/api v0.26.1
+	k8s.io/api v0.26.3
 	k8s.io/apiextensions-apiserver v0.26.1
-	k8s.io/apimachinery v0.26.1
+	k8s.io/apimachinery v0.26.3
```
viveksinghggits, line 59: If we are changing one k8s deps I think we should change others as well. For example `apiextensions-apiserver`, `client-go`, cli-runtime` etc. 
What do you think?

julio-lopez: These were automatically pulled when the openshift/client-go package was updated.
It actually makes sense to put all of the k8s deps in a separate PR altogether. Let me create a separate PR for the k8s deps.

go.mod
```diff
@@ -39,9 +39,9 @@ require (
 	github.com/onsi/ginkgo/v2 v2.6.0
 	github.com/onsi/gomega v1.24.1
 
-	//pinned openshift to release-4.5 branch
-	github.com/openshift/api v0.0.0-20200526144822-34f54f12813a
-	github.com/openshift/client-go v0.0.0-20200521150516-05eb9880269c
+	//pinned openshift to release-4.15 branch
```
julio-lopez, line 42: This is technically not 'pinned'
```suggestion
	// use openshift from 'release-4.15' branch
```

pavannd1 (approved):

PR #2042, opened 2023-05-02
Fix error in creating aws copy snaphot request

leuyentran:
## Change Overview

Upgrading `aws-sdk-go` from `v1.44.244` to `v1.44.2501 exposed failures in `awsebs.SanpshotCopy()`. The upgrade was subsequently [reverted](https://github.com/kanisterio/kanister/commit/08e01640ffd651e78569df415e546667cd26dfbc) in order to mitigate errors while waiting for root cause analysis. 

This PR has the following changes:

- And root cause has been found: A recent [update](https://github.com/aws/aws-sdk-go/commit/42260bb91c37d4e9f31587e92c434d462077d266) in `aws-sdk-go` `v1.44.247`  has made it explicit that client's config region is used as `DestinationRegion` for AWS Copy Snapshot operation. This means that when `CopySnapshotRequest()` is made to create a `presignedURL`, the call needs to be made with the destination's client. This PR fix that for `awsebs` client . 

- As a result, `aws-sdk-go` can now be upgraded to `v1.44.250`

- Lastly, the error mentioned above was not caught in unit testing because `blockstorage_test.BlockStorageProviderSuite.TestSnapshotCopy()` was skipped due to the following reason. This PR commented out the skip statement. The test took between 10-15s across 3 runs. 
```go
c.Skip("Sometimes, snapcopy takes over 10 minutes. go test declares failure if tests are that slow.")
```

Note that the entire `BlockStorageProviderSuite` test are currently skipped by `make test` due to lack of env config. Due to the test's nature of calling external provider's API,  they should be placed in integration tests instead of unit tests. Issue is created to track: https://github.com/kanisterio/kanister/issues/2043. 
As a result, `blockstorage_test.BlockStorageProviderSuite.TestSnapshotCopy()` is not run at the moment even with the change of this PR

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/blockstorage/awsebs/awsebs.go
+++ b/pkg/blockstorage/awsebs/awsebs.go
@@ -282,16 +282,12 @@ func (s *EbsStorage) SnapshotCopy(ctx context.Context, from, to blockstorage.Sna
 	// independent of whether or not the snapshot is encrypted.
 	var presignedURL *string
 	if to.Region != from.Region {
-		fromCli, err2 := newEC2Client(from.Region, s.Ec2Cli.Config.Copy())
-		if err2 != nil {
-			return nil, errors.Wrap(err2, "Could not create client to presign URL for snapshot copy request")
-		}
 		si := ec2.CopySnapshotInput{
 			SourceSnapshotId:  aws.String(from.ID),
 			SourceRegion:      aws.String(from.Region),
 			DestinationRegion: ec2Cli.Config.Region,
 		}
-		rq, _ := fromCli.CopySnapshotRequest(&si)
+		rq, _ := ec2Cli.CopySnapshotRequest(&si)
 		su, err2 := rq.Presign(120 * time.Minute)
 		if err2 != nil {
 			return nil, errors.Wrap(err2, "Could not presign URL for snapshot copy request")
--- a/pkg/blockstorage/blockstorage_test.go
+++ b/pkg/blockstorage/blockstorage_test.go
@@ -144,8 +144,6 @@ func (s *BlockStorageProviderSuite) TestCreateSnapshot(c *C) {
 }
 
 func (s *BlockStorageProviderSuite) TestSnapshotCopy(c *C) {
-	// c.Skip("Sometimes, snapcopy takes over 10 minutes. go test declares failure if tests are that slow.")
-
 	var snap *blockstorage.Snapshot
 	var err error
 	srcSnapshot := s.createSnapshot(c)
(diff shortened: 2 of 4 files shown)


pkg/blockstorage/awsebs/awsebs.go
```diff
@@ -282,16 +282,12 @@ func (s *EbsStorage) SnapshotCopy(ctx context.Context, from, to blockstorage.Sna
...
-		if err2 != nil {
-			return nil, errors.Wrap(err2, "Could not create client to presign URL for snapshot copy request")
-		}
 		si := ec2.CopySnapshotInput{
 			SourceSnapshotId:  aws.String(from.ID),
 			SourceRegion:      aws.String(from.Region),
 			DestinationRegion: ec2Cli.Config.Region,
 		}
-		rq, _ := fromCli.CopySnapshotRequest(&si)
+		rq, _ := ec2Cli.CopySnapshotRequest(&si)
```
julio-lopez, line 290: IIUC, the region used to create the client (in the constructor above) needs to match the region in the request as well, is that correct? (that's what I gather from the comment above (_// Copy operation must be initiated from the destination region._), so double checking.

pkg/blockstorage/blockstorage_test.go
```diff
@@ -144,7 +144,7 @@ func (s *BlockStorageProviderSuite) TestCreateSnapshot(c *C) {
 }
 
 func (s *BlockStorageProviderSuite) TestSnapshotCopy(c *C) {
-	c.Skip("Sometimes, snapcopy takes over 10 minutes. go test declares failure if tests are that slow.")
+	// c.Skip("Sometimes, snapcopy takes over 10 minutes. go test declares failure if tests are that slow.")
```
julio-lopez, line 147: Is the intent to remove the `//` before merging to prevent checking in commented code?

leuyentran: My intention of leaving this as a comment is to leave some easy breadcrumbs behind on which test took a long time in the past, in case the problem re-surface.  Alternatively, a log of test runs should also quickly provide the same information. I'll remove the `Skip` statement entirely

leuyentran: Done

leuyentran:
Follow up issue on testing: `blockstorage_test.BlockStorageProviderSuite.TestSnapshotCopy()`

kidgilson (approved):
LGTM

PR #2056, opened 2023-05-10
Pod controller change stop signature

e-sumin:
## Change Overview

Since PodController is a general purpose interface, it makes no sense to pass timeout to StartPod method and keep it internally. It's much better to pass it to StopPod.
For the same reason it makes no sense to StopPod when WaitForPodReady failed. Interface user might want to do something before stopping pod.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod_controller.go
+++ b/pkg/kube/pod_controller.go
@@ -132,7 +132,6 @@ func (p *podController) WaitForPodReady(ctx context.Context) error {
 
 	if err := p.pcp.waitForPodReady(ctx, p.podName); err != nil {
 		log.WithError(err).Print("Pod failed to become ready in time", field.M{"PodName": p.podName, "Namespace": p.podOptions.Namespace})
-		_ = p.StopPod(ctx) // best-effort
 		return errors.Wrap(err, "Pod failed to become ready in time")
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/kube/pod_controller.go
```diff
@@ -40,9 +40,9 @@ var (
 type PodController interface {
 	PodName() string
 	Pod() *corev1.Pod
-	StartPod(ctx context.Context, stopTimeout time.Duration) error
+	StartPod(ctx context.Context) error
```
viveksinghggits, line 43: Removing `stopTimeout` is good idea. I was thinking would it be a good idea if `StartPod` accepts another parameter that decides if `StartPod` wold also wait for the pod to be Running.
But I see we have another method that does that.
I am not sure if removing `WaitForPodReady` method and having that handled in `StartPod` would be better idea but it's worth thinking about.

e-sumin: Originally it was written in a way you are describing, but I think that generic interface should provide granular methods, and consumers are able to combine them as they would like. Also that makes tests easier.

pavannd1: Agree with @e-sumin too 👍🏼

pavannd1 (approved):

PR #2072, opened 2023-05-24
Fix crash in copyVolumeDataPodFunc when location type is not GCS

e-sumin:
## Change Overview

copyVolumeDataPodFunc is crashing when location type is not GCS

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/function/copy_volume_data.go
+++ b/pkg/function/copy_volume_data.go
@@ -99,7 +99,9 @@ func copyVolumeDataPodFunc(cli kubernetes.Interface, tp param.TemplateParams, na
 			return nil, errors.Wrapf(err, "Failed to write credentials to Pod %s", pc.PodName())
 		}
 
-		defer remover.Remove(context.Background()) //nolint:errcheck
+		if remover != nil { // WriteCredsToPod might return nil when nothing was written
+			defer remover.Remove(context.Background()) //nolint:errcheck
+		}
 
 		pod := pc.Pod()
 		// Get restic repository
--- a/pkg/function/copy_volume_data.go
+++ b/pkg/function/copy_volume_data.go
@@ -99,9 +99,8 @@ func copyVolumeDataPodFunc(cli kubernetes.Interface, tp param.TemplateParams, na
 			return nil, errors.Wrapf(err, "Failed to write credentials to Pod %s", pc.PodName())
 		}
 
-		if remover != nil { // WriteCredsToPod might return nil when nothing was written
-			defer remover.Remove(context.Background()) //nolint:errcheck
-		}
+		// Parent context could already be dead, so removing file within new context
+		defer remover.Remove(context.Background()) //nolint:errcheck
 
 		pod := pc.Pod()
 		// Get restic repository
--- a/pkg/function/utils.go
+++ b/pkg/function/utils.go
@@ -99,6 +99,19 @@ func ValidateProfile(profile *param.Profile) error {
 	return nil
 }
 
+type nopRemover struct {
+}
+
+var _ kube.PodFileRemover = (*nopRemover)(nil)
+
+func (nr *nopRemover) Remove(ctx context.Context) error {
+	return nil
+}
+
+func (nr *nopRemover) Path() string {
+	return ""
+}
+
 // WriteCredsToPod creates a file with Google credentials if the given profile points to a GCS location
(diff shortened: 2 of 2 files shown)


viveksinghggits (approved):

pkg/function/copy_volume_data.go
```diff
@@ -99,7 +99,9 @@ func copyVolumeDataPodFunc(cli kubernetes.Interface, tp param.TemplateParams, na
 			return nil, errors.Wrapf(err, "Failed to write credentials to Pod %s", pc.PodName())
 		}
 
-		defer remover.Remove(context.Background()) //nolint:errcheck
+		if remover != nil { // WriteCredsToPod might return nil when nothing was written
+			defer remover.Remove(context.Background()) //nolint:errcheck
+		}
```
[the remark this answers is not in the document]
denisvmedia: If the parent context dies, the clean up call will be aborted as well. Thus, I think there's no issue here. But maybe it makes sense to put a note.

denisvmedia:
I think the real cause of the failure is that the remover is _unexpectedly_ nil. I strongly believe (however, my beliefs might be wrong) that returning nil for the object when there is no error (thus `return nil, nil`) is wrong in most of the cases (and here as well).

If we check

```go
// WriteCredsToPod creates a file with Google credentials if the given profile points to a GCS location
func WriteCredsToPod(ctx context.Context, writer kube.PodFileWriter, profile *param.Profile) (kube.PodFileRemover, error) {
	if profile.Location.Type == crv1alpha1.LocationTypeGCS {
		remover, err := writer.Write(ctx, consts.GoogleCloudCredsFilePath, bytes.NewBufferString(profile.Credential.KeyPair.Secret))
		if err != nil {
			return nil, errors.Wrapf(err, "Unable to write Google credentials to the pod.")
		}

		return remover, nil
	}
	return nil, nil
}
```

I'd instead implement a `NopRemover` that would do nothing, and return it instead of `nil`. This will prevent further mistakes like the one solved in this PR.

pkg/function/copy_volume_data.go
```diff
@@ -99,7 +99,9 @@ func copyVolumeDataPodFunc(cli kubernetes.Interface, tp param.TemplateParams, na
 			return nil, errors.Wrapf(err, "Failed to write credentials to Pod %s", pc.PodName())
 		}
 
-		defer remover.Remove(context.Background()) //nolint:errcheck
+		if remover != nil { // WriteCredsToPod might return nil when nothing was written
```
denisvmedia, line 102: Check my PR comment: https://github.com/kanisterio/kanister/pull/2072#issuecomment-1560867072

pkg/function/copy_volume_data.go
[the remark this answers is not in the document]
e-sumin: Exactly, I'll add a comment

pkg/function/copy_volume_data.go
e-sumin: Ok, I agree with NopRemover. I thought about this but decided that probably it's an overkill. But since this came not only to my mind, it probably makes sense.

pkg/function/utils.go
```diff
@@ -99,6 +99,19 @@ func ValidateProfile(profile *param.Profile) error {
...
+
+var _ kube.PodFileRemover = (*nopRemover)(nil)
+
+func (nr *nopRemover) Remove(ctx context.Context) error {
+	return nil
+}
+
+func (nr *nopRemover) Path() string {
+	return ""
+}
```
denisvmedia, line 113:
```suggestion
var _ kube.PodFileRemover = (nopRemover)(nil)

func (nr nopRemover) Remove(ctx context.Context) error {
	return nil
}

func (nr nopRemover) Path() string {
	return ""
}
```

pkg/function/utils.go
```diff
@@ -109,7 +122,7 @@ func WriteCredsToPod(ctx context.Context, writer kube.PodFileWriter, profile *pa
 
 		return remover, nil
 	}
-	return nil, nil
+	return &nopRemover{}, nil
```
denisvmedia, line 125:
```suggestion
	return nopRemover{}, nil
```

e-sumin: done

pavannd1 (approved):

PR #2105, opened 2023-06-13
Actions cache gomod

viveksinghggits:
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


.github/workflows/main.yaml
```diff
@@ -9,8 +9,22 @@ jobs:
   gomod:
     runs-on: ubuntu-20.04
     steps:
+    - id: go-cache-paths
```
julio-lopez, line 12: It looks like this workflow (and maybe others as well) is not using the [setup-go](https://github.com/actions/setup-go) action. That action utilizes caching by default and does "the right thing" in terms of assigning a key to the cache, etc. It probably makes sense to update the workflows to use the `setup-go` action.

Also, with caching enabled, the need for uploading some of the artifacts below goes away.

.github/workflows/main.yaml
```diff
@@ -9,8 +9,22 @@ jobs:
...
+      uses: actions/cache@v2
+      with:
+        path: /go/pkg
+        key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
+    - name: Go Mod Cache
+      uses: actions/cache@v2
+      with:
+        path: /go/pkg
+        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+    - run: make go-mod-tidy
```
julio-lopez, line 27: Somewhat related: it seems that a few of these jobs should be consolidated into a single job with many steps. That could amortize a significant portion of the setup cost.

julio-lopez (changes requested):

PR #2106, opened 2023-06-13
fix deletion of the pod created by BP without NS

AKhoria:
## Change Overview

This PR fixes pod which aren't deleted if create by blueprints without namespace

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

 run blueprint without namespace:

`actions:
  post-export:
    name: ""
    kind: Namespace
    phases:
      - func: KubeTask
        name: hookPhase
        args:
          command:
            - /bin/sh
            - -c
            - |
              sleep 50
          image: bitnami/kubectl
          podOverride:`

ensure that kanister-job was deleted

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod_controller.go
+++ b/pkg/kube/pod_controller.go
@@ -177,7 +177,7 @@ func (p *podController) StopPod(ctx context.Context, stopTimeout time.Duration, 
 		defer cancel()
 	}
 
-	if err := p.pcp.deletePod(ctx, p.podOptions.Namespace, p.podName, metav1.DeleteOptions{GracePeriodSeconds: &gracePeriodSeconds}); err != nil {
+	if err := p.pcp.deletePod(ctx, p.pod.Namespace, p.podName, metav1.DeleteOptions{GracePeriodSeconds: &gracePeriodSeconds}); err != nil {
 		log.WithError(err).Print("Failed to delete pod", field.M{"PodName": p.podName, "Namespace": p.podOptions.Namespace})
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/kube/pod_controller.go
```diff
@@ -177,7 +177,7 @@ func (p *podController) StopPod(ctx context.Context, stopTimeout time.Duration,
 		defer cancel()
 	}
 
-	if err := p.pcp.deletePod(ctx, p.podOptions.Namespace, p.podName, metav1.DeleteOptions{GracePeriodSeconds: &gracePeriodSeconds}); err != nil {
+	if err := p.pcp.deletePod(ctx, p.pod.Namespace, p.podName, metav1.DeleteOptions{GracePeriodSeconds: &gracePeriodSeconds}); err != nil {
 		log.WithError(err).Print("Failed to delete pod", field.M{"PodName": p.podName, "Namespace": p.podOptions.Namespace})
```
pavannd1, line 181:
```suggestion
		log.WithError(err).Print("Failed to delete pod", field.M{"PodName": p.podName, "Namespace": p.pod.Namespace})
```

e-sumin (approved):
LGTM but I think it would be good to have unit test coverage for both cases (with and without namespace).

AKhoria:
> LGTM but I think it would be good to have unit test coverage for both cases (with and without namespace).
Do you mean StartPod function?
But this logic (choosing namespace) is performed not in pod_controller but inside pod.go which is mocked.

PR #2132, opened 2023-06-22
Fix pkg/kube to use standard kubeconfig machinery

miquella:
## Change Overview

The standard kubeconfig loading machinery provided by `k8s.io/client-go/tools/clientcmd` handles all of the same variants that `./pkg/kube.LoadConfig` was trying to handle. It also better handles certain cases, such as the `KUBECONFIG` environment variable referencing multiple kubeconfig files.

Since the `newClientConfig` function was already using the standard machinery (`clientcmd.NewNonInteractiveDeferredLoadingClientConfig` and `clientcmd.NewDefaultClientConfigLoadingRules`), simply removing the other special cases and letting it fall through was all that was necessary.

## Pull request type

- [X] :bug: Bugfix

## Issues

- N/A

## Test Plan

- [ ] :muscle: Manual
- [X] :zap: Unit test
- [X] :green_heart: E2E


PrasadG193 (approved):
LGTM

PR #2134, opened 2023-06-22
Import the general client/auth package

miquella:
## Change Overview

Rather than manually importing each auth sub-package, the `…/client/auth` package is imported instead. This package imports the default set of built-in client auth methods (which is a superset of the ones previously imported).

## Pull request type

- [X] :hamster: Trivial/Minor

## Issues 

- N/A

## Test Plan

- [ ] :muscle: Manual
- [X] :zap: Unit test
- [X] :green_heart: E2E


PrasadG193 (approved):
LGTM

PR #2144, opened 2023-06-28
Add fields for kopia cache, log and config directory in repository server CR

kale-amruta:
## Change Overview

This PR adds optional fields for kopia cache,log and config directory in the repository server CR. If these fields are not provided, default values for these fields would be used which connecting to kopia repository

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/apis/cr/v1alpha1/repositoryserver_types.go
```diff
@@ -81,13 +81,18 @@ type Repository struct {
...
 	CacheSizeSettings CacheSizeSettings      `json:"cacheSizeSettings,omitempty"`
+	// LogDirectory is an optional field to specify kopia log directory
+	LogDirectory string `json:"logDirectory,omitempty"`
+	// ConfigDirectory is an optional field to specify kopia config directory
+	ConfigFilePath string `json:"configFilePath,omitempty"`
 }
 
 // CacheSizeSettings are the metadata/content cache size details
 // that can be used while establishing connection to the kopia repository
 type CacheSizeSettings struct {
```
ankitjain235, line 92:
```suggestion
type CacheSettings struct {
```
Should we rename it now since it's not just `CacheSizeSettings` anymore?
Although that could break the CRD so maybe we can add another object, say `Configuration`, under which we can keep adding new fields. WDYT?

pkg/controllers/repositoryserver/repository.go
```diff
@@ -52,8 +53,21 @@ func (h *RepoServerHandler) connectToKopiaRepository() error {
...
+	if h.RepositoryServer.Spec.Repository.ConfigFilePath != "" {
+		configFilePath = h.RepositoryServer.Spec.Repository.ConfigFilePath
+	}
+	if h.RepositoryServer.Spec.Repository.LogDirectory != "" {
+		logDirectory = h.RepositoryServer.Spec.Repository.LogDirectory
+	}
+	return
+}
+
+func (h *RepoServerHandler) getRepositoryCacheSettings() (contentCacheMB, metadataCacheMB int, cacheDirectory string, err error) {
```
ankitjain235, line 68:
```suggestion
func (h *RepoServerHandler) getRepositoryCacheSettings() (cacheSizeSettings CacheSizeSettings, err error) {
```
We should return the struct if we can to avoid returning multiple fields.

pkg/controllers/repositoryserver/repository.go
```diff
@@ -52,8 +53,21 @@ func (h *RepoServerHandler) connectToKopiaRepository() error {
...
+func (h *RepoServerHandler) getRepositoryCommonArguments() (configFilePath, logDirectory string) {
+	configFilePath = command.DefaultConfigFilePath
+	logDirectory = command.DefaultLogDirectory
+	if h.RepositoryServer.Spec.Repository.ConfigFilePath != "" {
+		configFilePath = h.RepositoryServer.Spec.Repository.ConfigFilePath
+	}
+	if h.RepositoryServer.Spec.Repository.LogDirectory != "" {
+		logDirectory = h.RepositoryServer.Spec.Repository.LogDirectory
+	}
+	return
```
ankitjain235, line 65:
```suggestion
	return configFilePath, logDirectory
```
We should avoid naked returns since it forces the one changing code later to read the code very carefully and increases chances of error.

pkg/controllers/repositoryserver/repository.go
```diff
@@ -66,5 +80,8 @@ func (h *RepoServerHandler) getRepositoryCacheSettings() (contentCacheMB, metada
 			return
 		}
 	}
+	if h.RepositoryServer.Spec.Repository.CacheSizeSettings.CacheDirectory != "" {
+		cacheDirectory = h.RepositoryServer.Spec.Repository.CacheSizeSettings.CacheDirectory
+	}
 	return
```
ankitjain235, line 86: Although this was merged, let's update to return values here as well.

pkg/customresource/repositoryserver.yaml
```diff
@@ -42,6 +42,8 @@ spec:
                       size details that can be used while establishing connection
                       to the kopia repository
                     properties:
+                      cacheDirectory:
+                        type: string
```
ankitjain235, line 46: Same thing here as well, we should move this to some place else since it does not make sense under `cacheSizeSettings`.

ankitjain235 (commented):
Should we consider adding a common object in CRD under which all these configurable values, and new ones that could come up be added?

pkg/controllers/repositoryserver/repository.go
kale-amruta: @ankitjain235 I will address this as a part of [this](https://github.com/kanisterio/kanister/issues/2231) issue where I will be making cacheSizeSettings fields as `*int` , In current scenarios the fields in the struct are strings. Hence lot of conversion - deconversion  is happening, which I am not liking. 
For now I have removed naked returns but still returning multiple fields . WDYT ?

pkg/apis/cr/v1alpha1/repositoryserver_types.go
```diff
@@ -81,6 +81,18 @@ type Repository struct {
...
+}
+
+// Configuration can be used to specify the optional fields used
+// for repository operations
+type Configuration struct {
+	// CacheDirectory is an optional field to specify kopia cache directory
+	CacheDirectory string `json:"cacheDirectory,omitempty"`
+	// LogDirectory is an optional field to specify kopia log directory
+	LogDirectory string `json:"logDirectory,omitempty"`
+	// configFilePath is an optional field to specify kopia config file path
```
ankitjain235, line 94:
```suggestion
	// ConfigFilePath is an optional field to specify kopia config file path
```

pkg/controllers/repositoryserver/utils.go
```diff
@@ -248,3 +249,12 @@ func getVolumes(ctx context.Context, cli kubernetes.Interface, secret *corev1.Se
 	}
 	return vols, nil
 }
+
+func GetIntOrDefault(value string, defaultValue int) (int, error) {
```
ankitjain235, line 253: There are similar utilities in pkg/utils/utils.go, should we move this utility as well there?

pkg/customresource/repositoryserver.yaml
```diff
@@ -51,6 +51,23 @@ spec:
...
+                  configuration:
+                    description: Configuration can be used to specify the optional
+                      fields used for repository operations
+                    properties:
+                      cacheDirectory:
+                        description: CacheDirectory is an optional field to specify
+                          kopia cache directory
+                        type: string
+                      configFilePath:
+                        description: configFilePath is an optional field to specify
```
ankitjain235, line 63:
```suggestion
                        description: ConfigFilePath is an optional field to specify
```

ankitjain235 (commented):
Added minor comments, looks good otherwise.

ankitjain235 (approved):
LGTM

viveksinghggits (approved):

PR #2146, opened 2023-06-28
Use claimName field in the filestore location secret for repository server controller 

### LINKED ISSUE #2140: [BUG] Missing mount when using kopia filestore repository
**Describe the bug**

I've been trying to use Kanister with the new repository implementation with Kopia, using a filestore backend. However, it seems the the `claimName` parameter in the secret is not used, and as such the Kopia repository is never mounted in neither the kanister operator or the repo server pod. https://github.com/kanisterio/kanister/blob/b4552a63635b5e7df3609f28948430a7a7b89a0c/design/kanister-kopia-integration.md?plain=1#L295

**To Reproduce**

1. Follow the example in https://github.com/kanisterio/kanister/blob/b4552a63635b5e7df3609f28948430a7a7b89a0c/design/kanister-kopia-integration.md
2. Set `type` to `filestore` in the `location` secret
3. Create a PVC and reference it in the `location` secret

**Expected behavior**

I would expect the claim to be mounted in the Kopia repository server so the data can be backed up to the linked PVC

**Screenshots**

N/A

**Environment**
Kubernetes Version/Provider: v1.27.2+rke2r1
Storage Provider: NFS
Cluster Size (#nodes): 3
Data Size: N/A

**Additional context**

Kanister deployed using the latest Helm Chart version.

kale-amruta:
## Change Overview
If filestore location storage is used as repository in repository server controller , claimName field in the location secret is not honoured. This field should be used to mount the pvc volume on repository server pod as a `kopia repository`. This PR fixes the issue by adding pvc as volumes on the repository server pod

## Pull request type

- [] :construction: Work in Progress
- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2140 

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/controllers/repositoryserver/utils.go
```diff
@@ -225,3 +229,22 @@ func WaitTillCommandSucceed(ctx context.Context, cli kubernetes.Interface, cmd [
...
+	if len(secret.Data) == 0 {
+		return nil, errors.Errorf(secerrors.EmptySecretErrorMessage, secret.Namespace, secret.Name)
+	}
+	if locationType, ok := (secret.Data[reposerver.TypeKey]); ok && reposerver.LocType(string(locationType)) == reposerver.LocTypeFilestore {
+		if claimName, ok = secret.Data[reposerver.ClaimNameKey]; !ok {
+			return nil, errors.New("Claim name not set for file store location secret,failed to retrieve PVC")
+		}
+		claimNameString := string(claimName)
+		if _, err := cli.CoreV1().PersistentVolumeClaims(namespace).Get(ctx, string(claimNameString), metav1.GetOptions{}); err != nil {
+			return nil, errors.Wrapf(err, "Failed to retrieve PVC,%s:%s", namespace, claimName)
```
ankitjain235, line 245:
```suggestion
			return nil, errors.Wrapf(err, "Failed to validate is PVC exists,%s:%s", namespace, claimName)
```
Should we update the error message here if we are just validating if PVC exists?

pkg/testutil/testutil.go
```diff
@@ -494,8 +501,8 @@ func CreateTestKopiaRepository(cli kubernetes.Interface, rs *v1alpha1.Repository
 	)
 }
 
-func GetTestKopiaRepositoryServerCR(namespace string) *crv1alpha1.RepositoryServer {
-	repositoryServer := &crv1alpha1.RepositoryServer{
+func GetTestKopiaRepositoryServerCR(namespace string) crv1alpha1.RepositoryServer {
```
ankitjain235, line 504: Why do we need this change? Can we separate it out in a different PR if it's not related to this PR?

ankitjain235 (commented):
Can we separate out unrelated changes here? Looks good to me otherwise.

pkg/testutil/testutil.go
kale-amruta: since we are using pointer for default repository server spec, if we make changes to its spec in 1 test, other tests are also impacted. The test for this PR was also failing because of that. Hence I had added this change. I will add it to another PR which will be used as a base in this PR/

kale-amruta: Added it in separate PR - https://github.com/kanisterio/kanister/pull/2170

ankitjain235 (approved):
LGTM

pkg/controllers/repositoryserver/utils.go
```diff
@@ -225,3 +229,22 @@ func WaitTillCommandSucceed(ctx context.Context, cli kubernetes.Interface, cmd [
...
+	if len(secret.Data) == 0 {
+		return nil, errors.Errorf(secerrors.EmptySecretErrorMessage, secret.Namespace, secret.Name)
+	}
+	if locationType, ok := (secret.Data[reposerver.TypeKey]); ok && reposerver.LocType(string(locationType)) == reposerver.LocTypeFilestore {
+		if claimName, ok = secret.Data[reposerver.ClaimNameKey]; !ok {
+			return nil, errors.New("Claim name not set for file store location secret,failed to retrieve PVC")
+		}
+		claimNameString := string(claimName)
+		if _, err := cli.CoreV1().PersistentVolumeClaims(namespace).Get(ctx, string(claimNameString), metav1.GetOptions{}); err != nil {
+			return nil, errors.Wrapf(err, "Failed to validate if PVC exists,%s:%s", namespace, claimName)
```
pavannd1, line 245:
```suggestion
			return nil, errors.Wrapf(err, "Failed to validate if PVC %s:%s exists", namespace, claimName)
```
[nit]

pkg/controllers/repositoryserver/utils.go
```diff
@@ -225,3 +229,22 @@ func WaitTillCommandSucceed(ctx context.Context, cli kubernetes.Interface, cmd [
...
+
+func getVolumes(ctx context.Context, cli kubernetes.Interface, secret *corev1.Secret, namespace string) (map[string]string, error) {
+	vols := make(map[string]string, 0)
+	var claimName []byte
+	if len(secret.Data) == 0 {
+		return nil, errors.Errorf(secerrors.EmptySecretErrorMessage, secret.Namespace, secret.Name)
+	}
+	if locationType, ok := (secret.Data[reposerver.TypeKey]); ok && reposerver.LocType(string(locationType)) == reposerver.LocTypeFilestore {
+		if claimName, ok = secret.Data[reposerver.ClaimNameKey]; !ok {
+			return nil, errors.New("Claim name not set for file store location secret,failed to retrieve PVC")
```
pavannd1, line 241:
```suggestion
			return nil, errors.New("Claim name not set for file store location secret, failed to retrieve PVC")
```
`,` followed by a space improves readability

pkg/controllers/repositoryserver/repositoryserver_controller_test.go
```diff
@@ -368,6 +369,56 @@ func (s *RepoServerControllerSuite) TestInvalidRepositoryPassword(c *C) {
...
+
+	c.Assert(len(pod.Spec.Volumes), Equals, 3)
+
+	var volumeattached bool
+	for _, vol := range pod.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil && vol.PersistentVolumeClaim.ClaimName == pvc.Name {
+			volumeattached = true
+		}
+	}
+	c.Assert(volumeattached, Equals, true)
```
PrasadG193, line 419:
```suggestion
	for _, vol := range pod.Spec.Volumes {
		if vol.PersistentVolumeClaim == nil || vol.PersistentVolumeClaim.ClaimName != pvc.Name {
			c.Errorf("Could not found expected PVC Claim in Pod spec. Expected: %s", pvc.Name)
		}
	}
```
[nit]

pkg/controllers/repositoryserver/repositoryserver_controller_test.go
```diff
@@ -368,6 +369,56 @@ func (s *RepoServerControllerSuite) TestInvalidRepositoryPassword(c *C) {
...
+
+	repoServerCRCreated, err := s.crCli.RepositoryServers(s.repoServerControllerNamespace).Create(ctx, &repoServerCR, metav1.CreateOptions{})
+	c.Assert(err, IsNil)
+
+	err = s.waitForRepoServerInfoUpdateInCR(repoServerCRCreated.Name)
+	c.Assert(err, IsNil)
+
+	//Get repository server CR with the updated server information
+	repoServerCRCreated, err = s.crCli.RepositoryServers(s.repoServerControllerNamespace).Get(ctx, repoServerCRCreated.Name, metav1.GetOptions{})
+	c.Assert(err, IsNil)
```
PrasadG193, line 406: [nit] Please explore if these steps about setting kopia server are common across most of the tests. If that is the case we can think about moving these into a separate helper function

pkg/controllers/repositoryserver/utils.go
```diff
@@ -225,3 +229,22 @@ func WaitTillCommandSucceed(ctx context.Context, cli kubernetes.Interface, cmd [
...
+	var claimName []byte
+	if len(secret.Data) == 0 {
+		return nil, errors.Errorf(secerrors.EmptySecretErrorMessage, secret.Namespace, secret.Name)
+	}
+	if locationType, ok := (secret.Data[reposerver.TypeKey]); ok && reposerver.LocType(string(locationType)) == reposerver.LocTypeFilestore {
+		if claimName, ok = secret.Data[reposerver.ClaimNameKey]; !ok {
+			return nil, errors.New("Claim name not set for file store location secret, failed to retrieve PVC")
+		}
+		claimNameString := string(claimName)
+		if _, err := cli.CoreV1().PersistentVolumeClaims(namespace).Get(ctx, string(claimNameString), metav1.GetOptions{}); err != nil {
```
PrasadG193, line 244:
```suggestion
		if _, err := cli.CoreV1().PersistentVolumeClaims(namespace).Get(ctx, claimNameString, metav1.GetOptions{}); err != nil {
```
Its already in the string type, converted on the live above

PrasadG193 (approved):
Looks good, left some minor suggestions. Can be merged once addressed.

pkg/controllers/repositoryserver/repositoryserver_controller_test.go
kale-amruta: As discussed, I will explore this as a part of separate ticket

pkg/controllers/repositoryserver/repositoryserver_controller_test.go
kale-amruta: @PrasadG193 
- if `vol.PersistentVolumeClaim == nil` how can we check if `vol.PersistentVolumeClaim.ClaimName != pvc.Name`
- Also if we check `vol.PersistentVolumeClaim != nil`, and then `vol.PersistentVolumeClaim.ClaimName != pvc.Name` , then it will check each volume in the pod should have claim name as the one that we are creating, which is not the case. The Pod can have other volumes have other PVCs

I am resolving the conversion since the logic seems to be wrong here.

kale-amruta: @PrasadG193 
- There are volumes in the repo server pod which has PVC as `Nil`, hence the test will always with this logic

I am resolving the conversion since the logic seems to be wrong here.

PR #2166, opened 2023-07-11
Enable Return snapInfo in kopiaLocationPush

r4rajat:
## Change Overview

Enable returning snapshot information in function kopiaLocationPush in order to add tests for Kopia RepositoryServer based Profiles and CR

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/datamover/repository_server.go
+++ b/pkg/datamover/repository_server.go
@@ -52,7 +52,8 @@ func (rs *repositoryServer) Push(ctx context.Context, sourcePath, destinationPat
 	if err != nil {
 		return err
 	}
-	return kopiaLocationPush(ctx, destinationPath, rs.outputName, sourcePath, password)
+	_, err = kopiaLocationPush(ctx, destinationPath, rs.outputName, sourcePath, password)
+	return err
 }
 
 func (rs *repositoryServer) Delete(ctx context.Context, destinationPath string) error {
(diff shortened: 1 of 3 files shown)


pkg/datamover/repository_server.go
```diff
@@ -52,7 +52,8 @@ func (rs *repositoryServer) Push(ctx context.Context, sourcePath, destinationPat
 	if err != nil {
 		return err
 	}
-	return kopiaLocationPush(ctx, destinationPath, rs.outputName, sourcePath, password)
+	_, err = kopiaLocationPush(ctx, destinationPath, rs.outputName, sourcePath, password)
```
viveksinghggits, line 55: It looks like we are not consuming the value that is being returned from the function? How is that being used?

viveksinghggits (approved):

pkg/datamover/repository_server.go
r4rajat: It would be used while Testing Kopia Repository Server based changes

PR #2169, opened 2023-07-12
Set RepositoryServerCR Conditions

akankshakumari393:
## Change Overview

This PR sets Repository Server CR Conditions based on different errors.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

[https://gist.github.com/akankshakumari393/bcb59453b69e4501b5d4cb6a8184493c](https://gist.github.com/akankshakumari393/bcb59453b69e4501b5d4cb6a8184493c)


pkg/controllers/repositoryserver/handler.go
```diff
@@ -331,26 +342,38 @@ func (h *RepoServerHandler) updateRepoServerProgress(ctx context.Context, progre
 func (h *RepoServerHandler) setupKopiaRepositoryServer(ctx context.Context, logger logr.Logger) (ctrl.Result, error) {
 	logger.Info("Start Kopia Repository Server")
 	if err := h.startRepoProxyServer(ctx); err != nil {
-		if uerr := h.updateRepoServerProgress(ctx, crkanisteriov1alpha1.Failed); uerr != nil {
+		if uerr := h.handleEvent(ctx, corev1.EventTypeWarning, serverInitializedErrReason, err.Error(), err.Error(), crkanisteriov1alpha1.ServerInitialized, crkanisteriov1alpha1.Failed, metav1.ConditionFalse); uerr != nil {
 			return ctrl.Result{}, uerr
 		}
-		return ctrl.Result{}, err
+		return ctrl.Result{Requeue: false}, err
```
kale-amruta, line 348: Why are we setting `Requeue: false` explicitly here? do we not want to requeue when we get an error while starting the server?

pkg/controllers/repositoryserver/server.go
```diff
@@ -48,35 +48,43 @@ func (h *RepoServerHandler) startRepoProxyServer(ctx context.Context) (err error
 		return nil
 	}
 
-	cmd := command.ServerStart(
-		command.ServerStartCommandArgs{
-			CommandArgs: &command.CommandArgs{
-				RepoPassword:   "",
-				ConfigFilePath: command.DefaultConfigFilePath,
-				LogDirectory:   command.DefaultLogDirectory,
+	maxRetries := 5
```
kale-amruta, line 51: Can we move this to a const ?

kale-amruta:
@akankshakumari393 can you check if we can add unit tests for these changes under `repositoryserver_controller_test.go` ?

pkg/controllers/repositoryserver/server.go
```diff
@@ -48,35 +48,43 @@ func (h *RepoServerHandler) startRepoProxyServer(ctx context.Context) (err error
...
+				return nil
+			}
+		}
 
-	err = h.waitForServerReady(ctx, repoServerAddress, serverAdminUserName, serverAdminPassword)
-	if err != nil {
-		return errors.Wrap(err, "Failed to check Kopia API server status")
+		// Retry after a delay
+		if retries < maxRetries {
+			delay := time.Second * 5
```
kale-amruta, line 82: Lets move this to a const?

pkg/controllers/repositoryserver/server.go
```diff
@@ -48,35 +48,43 @@ func (h *RepoServerHandler) startRepoProxyServer(ctx context.Context) (err error
...
+					LogDirectory:   command.DefaultLogDirectory,
+				},
+				ServerAddress:    repoServerAddress,
+				TLSCertFile:      tlsCertPath,
+				TLSKeyFile:       tlsKeyPath,
+				ServerUsername:   serverAdminUserName,
+				ServerPassword:   serverAdminPassword,
+				AutoGenerateCert: false,
+				Background:       true,
 			},
```
kale-amruta, line 67: Can we move command creation out of the loop, since it wont change on each iteration?

akankshakumari393:
> @akankshakumari393 can you check if we can add unit tests for these changes under repositoryserver_controller_test.go?
will create a follow up PR on this.

pkg/controllers/repositoryserver/utils.go
```diff
@@ -59,6 +59,21 @@ const (
...
+	serverSetupSuccessReason string = "KopiaRepositoryServerSetupSucceeded"
+
+	repositoryConnectedErrReason     string = "KopiaRepositoryConnectionFailed"
+	repositoryConnectedSuccessReason string = "KopiaRepositoryConnectionSucceeded"
+
+	serverInitializedErrReason     string = "KopiaRepositoryServerInitializationFailed"
+	serverInitializedSuccessReason string = "KopiaRepositoryServerInitializationSucceeded"
+
+	clientsInitializedErrReason     string = "ClientInitializationFailed"
+	clientsInitializedSuccessReason string = "ClientInitializationSuccessful"
```
kale-amruta, line 73: [nit]

```suggestion
	clientsInitializedSuccessReason string = "ClientInitializationSucceded"
```

All other reason fields have succeded appended to it , hence lets be consistent with it

kale-amruta (approved):
lgtm with minor change

pkg/controllers/repositoryserver/handler.go
```diff
@@ -171,20 +168,26 @@ func (h *RepoServerHandler) reconcilePod(ctx context.Context, svc *corev1.Servic
 		return nil, nil, err
 	}
 	h.Logger.Info("Pod resource not found. Creating new pod")
+	return h.createPodUpdateStatus(ctx, repoServerNamespace, svc)
```
viveksinghggits, line 171:
```suggestion
	return h.createPodAndUpdateStatus(ctx, repoServerNamespace, svc)
```

pkg/controllers/repositoryserver/handler.go
```diff
@@ -171,20 +168,26 @@ func (h *RepoServerHandler) reconcilePod(ctx context.Context, svc *corev1.Servic
...
+}
+
+func (h *RepoServerHandler) setCondition(ctx context.Context, reason, conditionMsg, conditionType string, progress crv1alpha1.RepositoryServerProgress, status metav1.ConditionStatus) error {
```
viveksinghggits, line 174: I think we should accept these parameters as a struct type, maybe with name `Condition`.

pkg/controllers/repositoryserver/handler.go
```diff
@@ -284,7 +287,7 @@ func (h *RepoServerHandler) preparePodOverride(ctx context.Context) (map[string]
 	return podOverride, nil
 }
 
-func (h *RepoServerHandler) updateServerInfoInCRStatus(ctx context.Context, info crv1alpha1.ServerInfo) error {
+func (h *RepoServerHandler) updateServerInfoInCRStatus(ctx context.Context, podName string, serviceName string) error {
 	h.Logger.Info("Fetch latest version of RepositoryServer to update the ServerInfo in its status")
```
viveksinghggits, line 291: This log is not part of this PR and we don't have to change it now. But it looks like a comment more than a log. And there are other instances of similar logs I see in these files.

pkg/controllers/repositoryserver/utils.go
```diff
@@ -59,6 +59,21 @@ const (
 	tlsCertDefaultMountPath = "/mnt/secrets/tlscert"
 	tlsKeyPath              = "/mnt/secrets/tlscert/tls.key"
 	tlsCertPath             = "/mnt/secrets/tlscert/tls.crt"
+
+	serverSetupErrReason     string = "KopiaRepositoryServerSetupFailed"
```
viveksinghggits, line 63:
```suggestion
	conditionReasonServerSetupErr     string = "KopiaRepositoryServerSetupFailed"
```

viveksinghggits: and to other consts here.

pkg/controllers/repositoryserver/handler.go
viveksinghggits: I think we should create a task just to correct the logs.

pkg/controllers/repositoryserver/handler.go
```diff
@@ -284,7 +287,7 @@ func (h *RepoServerHandler) preparePodOverride(ctx context.Context) (map[string]
 	return podOverride, nil
 }
 
-func (h *RepoServerHandler) updateServerInfoInCRStatus(ctx context.Context, info crv1alpha1.ServerInfo) error {
+func (h *RepoServerHandler) updateServerInfoInCRStatus(ctx context.Context, podName string, serviceName string) error {
```
viveksinghggits, line 290: I see that this method is called when we create Pod as well as when we create service. When we call it after creating pod there wouldnt be a service for it and even after that we are always setting service name to passed service name?

pkg/controllers/repositoryserver/handler.go
akankshakumari393: the function set the condition and also updates the status, do you mean creating condition struct including `progress` as well?

viveksinghggits: The function has 6 parameters now. My recommendation was to create a struct with these parameters so that the function would accept that type as parameter

```
type Condition struct{
reason, 
conditionMsg,
 conditionType string, 
progress crv1alpha1.RepositoryServerProgress, 
status metav1.ConditionStatus
}

func (h *RepoServerHandler) setCondition(ctx context.Context, c Condition){
}
```

pkg/controllers/repositoryserver/handler.go
viveksinghggits: > Could you please point me to where are we directly referencing service?
I am not sure what do you mean by this. I didn't say that we are directly referencing service.

> I think first we are creating the service and then creating the pod.
Can you check if that's what is happening.

viveksinghggits: And if we are creating the service first, what would be the value of pod when we call `updateServerInfoInCRStatus(ctx context.Context, podName string, serviceName string)`?

pkg/controllers/repositoryserver/handler.go
akankshakumari393: updated

pkg/controllers/repositoryserver/handler.go
akankshakumari393: so in the first run, the pod value would be blank, and lets say the further steps failed, then on further cycle, the pod value would be set as in the previous one.

viveksinghggits: my concern is if we know that podname is empty why do we still pass it in the function?

kale-amruta: @viveksinghggits The values that are passed to this function when
1. After `create service` is successful,
  ```
   h.updateServerInfoInCRStatus(ctx, h.RepositoryServer.Status.ServerInfo.PodName, svc.Name)
  ```
This means we are setting the the pod name to whatever the field value we have in repository server CR currently. So we are rewriting the pod name with its exisitng value from the repository server CR's status. Hence while creating service only service name is updated with new value

2. After `create Pod` is successful

``` h.updateServerInfoInCRStatus(ctx, pod.Name, h.RepositoryServer.Status.ServerInfo.ServiceName)``` . Here we are rewriting the service name to its existing value from repository server CR's status. While creating pod, only pod name is updated with new value

Does this address your concern?

Comment (viveksinghggits) [reply pkg/controllers/repositoryserver/handler.go:298]: mm, I think so.
Maybe this question should have been part of other PR because the change doesn't seem to be related to this PR directly. I will not block the PR anymore and approve.

File: pkg/controllers/repositoryserver/handler.go [hunk 11]
```diff
@@ -171,20 +176,26 @@ func (h *RepoServerHandler) reconcilePod(ctx context.Context, svc *corev1.Servic
 		return nil, nil, err
 	}
 	h.Logger.Info("Pod resource not found. Creating new pod")
+	return h.createPodAndUpdateStatus(ctx, repoServerNamespace, svc)
+}
+
+func (h *RepoServerHandler) setCondition(ctx context.Context, conditionStatus ConditionStatus) error {
```
viveksinghggits, line 182:
```suggestion
func (h *RepoServerHandler) setCondition(ctx context.Context, condition metav1.Condition, progress WhatEverTypeThisIs) error {
```
This is going to simplify a lot of things. We wont have to form the arguments first and then figure out metav1.Condition from those arguments. We can just get the metav1.Condition and pass that around.

viveksinghggits (changes requested):

pkg/controllers/repositoryserver/handler.go
akankshakumari393: Thanks @kale-amruta 👍🏻

pkg/controllers/repositoryserver/handler.go
akankshakumari393: we are forming metav1.Condition from the getCondition() functions inside, If we want to pass metav1.condition, we would not maybe need the function also.

akankshakumari393: this function was added to replace duplicate calls for getCondition() and updatestatus() in the Reconciliation.

viveksinghggits: Right, so we should just remove that function? And that's the reason I suggested this change, to make things simpler and don't have extra type.

akankshakumari393: Ok I can do this but then we would have duplicate calls in Reconciliation and the function would be long, I had added this to address other comments that was to move duplicate things to a function. Would that be ok?

akankshakumari393: just want to make sure we are on the same page.

viveksinghggits: duplicated calls to which function?

akankshakumari393: I have addressed those comments please check.

viveksinghggits (approved):

PR #2235, opened 2023-08-01
Feature flagging for exporting of Kanister prometheus metrics

mellon-collie:
## Change Overview

Adding a new helm flag and environment variable for feature toggling of the new Kanister prometheus metrics integration. 
1. Added new boolean helm flag called `controller.metrics.enabled` in the Values.yaml
2. Added new environment variable called `KANISTER_METRICS_ENABLED` in deployment.yaml to capture the metrics.enabled flag in our codebase. 

End goal: In order to switch on prometheus metrics, the end user can install the local helm chart with the following command: 
`helm install kanister ./helm/kanister-operator \
  --create-namespace \
  --namespace kanister \
  --set image.repository=<your_registry>/<your_controller_image> \
  --set image.tag=<your_image_tag>
  --controller.metrics.enabled=true`

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes N/A

## Test Plan

After the above changes were made, the following tests were performed: 
1.  `helm uninstall kanister -n kanister` to uninstall existing kanister deployment
2. `helm install kanister ./helm/kanister-operator --create-namespace --namespace kanister` to run the local helm chart for kanister
4. `k get pods -n kanister` to identify the new kanister operator pod that's running
5. `kubectl exec -n kanister -it <pod-name> -- /bin/bash` to exec into the pod. 
6. `echo $KANISTER_METRICS_ENABLED` - the output should be a false.

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/helm/kanister-operator/values.yaml
+++ b/helm/kanister-operator/values.yaml
@@ -53,3 +53,5 @@ resources:
 # requests:
 #  cpu: 100m
 #  memory: 128Mi
+metrics:
+  enabled: false
--- a/helm/kanister-operator/values.yaml
+++ b/helm/kanister-operator/values.yaml
@@ -28,6 +28,11 @@ controller:
   # true : CRDs would be created by kanister controller
   updateCRDs: true
   parallelism: 8
+  metrics:
+    # metrics.enabled specified if the kanister-prometheus framework has been enabled
+    # false : kanister-prometheus framework has been disabled
+    # true: kanister-prometheus framework has been enabled
+    enabled: false
 bpValidatingWebhook:
   enabled: true
   tls:
@@ -53,5 +58,3 @@ resources:
 # requests:
 #  cpu: 100m
 #  memory: 128Mi
-metrics:
-  enabled: false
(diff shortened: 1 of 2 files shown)


helm/kanister-operator/values.yaml
```diff
@@ -53,3 +53,5 @@ resources:
 # requests:
 #  cpu: 100m
 #  memory: 128Mi
+metrics:
```
pavannd1, line 56: Should we add this as a sub-value under `controller` above?
Additionally, add a comment

mellon-collie: Sure. Makes sense. I've made it as a sub-value under `controller`. I've added a comment.
One more point - I'm maintaining two levels of hierarchy for the helm flags - like `metrics.enabled` instead of `metrics_enabled` because we may want to add more flags under `metrics` hierarchy in the future if we consider adding labels for a new metric partially.

pavannd1: Yes, that's the right way to do it 👍🏼

viveksinghggits: do we just want to set the env var in the pod as part of this PR? Do we not want to use it anywhere?

viveksinghggits (approved):

pavannd1 (approved):

PR #2238, opened 2023-08-03
Enable lll, nakedret, asciicheck, depguard, dogsled, dupl, exportloopref golang linters

ankitjain235:
## Change Overview

This PR enables following Golang linters:
- lll
- nakedret
- asciicheck
- depguard
- dogsled
- dupl
- exportloopref

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

- `depguard` was enabled in this PR but had to be disabled temporarily until https://github.com/golangci/golangci-lint/issues/3906 is fixed and a new release is available with the fix.


.golangci.yml
```diff
@@ -35,3 +44,7 @@ issues:
 linters-settings:
   gocyclo:
     min-complexity: 20
+  lll:
+    line-length: 240
```
PrasadG193, line 48: Trying to understand, what is the ideal value? How did we come up with 240?

pkg/app/cockroachdb.go
```diff
@@ -57,7 +57,7 @@ func (c *CockroachDB) Init(context.Context) error {
 	return err
 }
 
-func (c *CockroachDB) Install(ctx context.Context, namespace string) error {
+func (c *CockroachDB) Install(ctx context.Context, namespace string) error { //nolint:dupl
```
PrasadG193, line 60: Ignore mark should have explanation comment

.golangci.yml
ankitjain235: I took reference from some other project. I think we can reduce the value. Default is 120. Does it make sense to make this change in a follow up PR?

r4rajat (commented):
LGTM

pkg/app/cockroachdb.go
```diff
@@ -57,7 +57,7 @@ func (c *CockroachDB) Init(context.Context) error {
 	return err
 }
 
-func (c *CockroachDB) Install(ctx context.Context, namespace string) error {
+func (c *CockroachDB) Install(ctx context.Context, namespace string) error { //nolint:dupl // Not a duplicate, common code already extracted
```
PrasadG193, line 60: This is strange. Why the check is failing for `Install()` but not other methods like `IsReady()`, `Uninstall()`

pkg/function/scale_workload.go
```diff
@@ -115,14 +113,27 @@ func (s *scaleWorkloadFunc) ExecutionProgress() (crv1alpha1.PhaseProgress, error
 	}, nil
 }
 
-func getArgs(tp param.TemplateParams, args map[string]interface{}) (namespace, kind, name string, replicas int32, waitForReady bool, err error) {
+type scaleArgs struct {
+	namespace    string
+	kind         string
+	name         string
+	replicas     int32
+	waitForReady bool
```
PrasadG193, line 121: Should we make these fields part of `scaleWorkloadFunc`? Like we have `progressPercent`?

pkg/function/scale_workload.go
```diff
@@ -115,14 +113,27 @@ func (s *scaleWorkloadFunc) ExecutionProgress() (crv1alpha1.PhaseProgress, error
...
+	err := Arg(args, ScaleWorkloadReplicas, &rep)
 	if err != nil {
-		return
+		return nil, err
 	}
 
+	var (
+		namespace, kind, name string
+		replicas              int32
+	)
```
PrasadG193, line 135:
```suggestion
		var namespace, kind, name string
		var replicas int32
```
The space between  replicas and int32 looks weird

PrasadG193 (commented):
Left minor suggestions and a question. Otherwise looks good

pkg/function/scale_workload.go
ankitjain235: It's auto-formatting, based on `namespace, kind, name string`.

pkg/app/cockroachdb.go
ankitjain235: Not so sure, we still do have common code, but logs are different mostly and it didn't make sense to move them to a separate utility.

pkg/function/scale_workload.go
ankitjain235: The rule here is: We should not return these many variables. Since scaleWorkloadFunc.Exec is calling `getArgs()`, and `getArgs()` is called in this function only, do you think it would make sense to move it under `scaleWorkloadFunc.getArgs()`? These variables are usable in `Exec` function only, and not others.

pkg/app/cockroachdb.go
ankitjain235: I suppose it's not taking comments into consideration.

pkg/function/scale_workload.go
PrasadG193: My suggestion was to declare `var` on two lines instead of using `var()` pattern

pkg/function/scale_workload.go
PrasadG193: The arguments are properties of the function only, right? We should not make decisions based on whether they are used in one function or multiple. It also makes sense from an extensibility point of view, they can be referred to in other functions if required.

pkg/function/scale_workload.go
ankitjain235: Ohh, I misunderstood it somehow. Address the review comment now.

pkg/function/scale_workload.go
ankitjain235: I have moved them to `scaleWorkloadFunc.setArgs()`.

PrasadG193 (approved):
LGTM

PR #2240, opened 2023-08-03
Enable misspell, nestif golang linter

ankitjain235:
## Change Overview

This PR enables misspell and nestif golang linters.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


mellon-collie (changes requested):
Please rebase the commits to show logical progression, i.e., one commit could be to just enable go listing for `misspell` and the other commit could be for actually making the listing changes.

mellon-collie:
Should we also raise separate PRs for enabling each of these linters and fixing them? For instance, one PR could be just for enabling `misspell` and fixing all the `misspell` lint issues.

ankitjain235:
@mellon-collie I kept them this way so that we can track which changes were made for which enabled linter respectfully. It would be easy to track things for someone who is making changes in the same code path later.
When the PR is merged, then also the git blame would show this as part of linter enable PRs only and won't make users worry that this commit could be causing change in code behaviour. WDYT?

mellon-collie:
> @mellon-collie I kept them this way so that we can track which changes were made for which enabled linter respectfully. It would be easy to track things for someone who is making changes in the same code path later.
> When the PR is merged, then also the git blame would show this as part of linter enable PRs only and won't make users worry that this commit could be causing change in code behaviour. WDYT?

Sure, that makes sense. Thanks for clarifying!

.golangci.yml
```diff
@@ -22,6 +22,9 @@ linters:
     - dupl
     - exportloopref
     - importas
+    # - stylecheck It needs a lot of changes
```
PrasadG193, line 25: Should we remove this?

.golangci.yml
```diff
@@ -37,6 +40,8 @@ issues:
         - unparam                       # Tests might have unused function parameters.
         - lll
         - dupl
+        - misspell
```
PrasadG193, line 43: I think it is fine to enable `misspell` in test file if doesn't require much changes

.golangci.yml
```diff
@@ -60,3 +65,5 @@ linters-settings:
         alias: metav1
       - pkg: github.com/kanisterio/kanister/pkg/apis/cr/v1alpha1
         alias: crv1alpha1
+  nestif:
+    min-complexity: 6
```
PrasadG193, line 69: Just curious, what is the ideal value?

pkg/function/restore_rds_snapshot.go
```diff
@@ -199,11 +199,7 @@ func restoreRDSSnapshot(
 		// If securityGroupID arg is nil, we will try to find the sgIDs by describing the existing instance
 		// Find security group ids
 		if sgIDs == nil {
-			if !isAuroraCluster(string(dbEngine)) {
-				sgIDs, err = findSecurityGroups(ctx, rdsCli, instanceID)
-			} else {
-				sgIDs, err = findAuroraSecurityGroups(ctx, rdsCli, instanceID)
-			}
+			sgIDs, err = findSecurityGroupIDs(ctx, rdsCli, instanceID, string(dbEngine))
```
PrasadG193, line 202: same here

pkg/function/create_rds_snapshot.go
```diff
@@ -160,6 +139,36 @@ func createRDSSnapshot(ctx context.Context, instanceID string, dbEngine RDSDBEng
 	return output, nil
 }
 
+func createSnapshot(ctx context.Context, rdsCli *rds.RDS, snapshotID, dbEngine, instanceID string) (int64, error) {
```
PrasadG193, line 142:
```suggestion
func createRDSSnapshot(ctx context.Context, rdsCli *rds.RDS, snapshotID, dbEngine, instanceID string) (int64, error) {
```

.golangci.yml
ankitjain235: It does require changes. Enabling these linters for tests would need a lot of extra work. I'll create a separate issue for the same?

PrasadG193 (approved):

PR #2260, opened 2023-08-11
Test only latest version of CSI snapshotter present

ewhamilton:
## Change Overview

Change test to select latest version of CSI snapshotter present on system and only
test that version.

Avoids issues on GKE 1.27 where v1 and v1beta1 are both present for reading but
only v1 can be used.

## Pull request type

- [x] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [x] :robot: Test


ewhamilton:

pavannd1 (approved):

PR #2269, opened 2023-08-15
Fix issue with missing container name and namespace

chaitalisg:
## Change Overview

When creating pods with sidecars, we have an issue with writing files or streaming logs from these pods, because such operations requires container name.
This PR fixes this issue.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kube/pod_controller.go
```diff
@@ -204,12 +211,24 @@ func (p *podController) StopPod(ctx context.Context, stopTimeout time.Duration,
 	return nil
 }
 
+// getContainerName returns container name, which should be passed to operations which requires it
+// if container name was specified in podOptions, it will be used.
+// Otherwise, the first container name from specs will be taken as best effort (when pods are created with sidecars
+// sidecar containers are placed after main container).
```
pavannd1, line 217:
```suggestion
// getContainerName returns container name, which should be passed to
// operations that require it.
// If the container name was specified in podOptions, it will be used.
// Otherwise, the first container name from specs will be taken as best effort
// (when pods are created with sidecars, sidecar containers are placed after
// main container).
```
80 char wrap

pavannd1 (approved):
LGTM

chaitalisg:
> Hi @chaitalisg,
> Like mentioned in the PR description in case of multiple containers in a pod when we try to read logs there was a problem. But how do we know that it's the first container that we want to read the logs from?
> cc @pavannd1

Good question. It is a calculated risk. When pods are created with sidecars, sidecar containers are placed after main container. So, I am using the first container. Let me know if you know a way to make this better!

viveksinghggits:
> When pods are created with sidecars, sidecar containers are placed after main container.

I don't think this is the case always.
I think we should pass the container name explicitly instead. For example if we have a utility to exec into a pod it should expect a container name, similarly in this case since we are trying to get logs of a container we should explicitly expect the container name.
This can default to the container name that we set in podOptions while creating the pod because the is the that we would be interested in.

viveksinghggits:
I have raised this PR that fixes above problem 
https://github.com/kanisterio/kanister/pull/2310#event-10306105151

PR #2275, opened 2023-08-18
Update kopia commit in kanister-tools image to latest master

KastenMike:
Bumping since it was last done 6 months ago and kopia uses newer azure APIs since then


PrasadG193 (approved):

pavannd1 (changes requested):
Holding off until all the downstream tests are complete

ankitjain235 (changes requested):
We should also update Kopia dependency version in go.mod so that both CLI and SDK are in sync.

KastenMike:
> We should also update Kopia dependency version in go.mod so that both CLI and SDK are in sync.

What do you mean? This PR is to uptake the go.mod changes or I'm misunderstanding

KastenMike:
Closing and trying to create directly in kanister repo

PR #2283, opened 2023-08-22
update kopia to a more recent version

KastenMike:
Bumping since it was last done 6 months ago and kopia uses newer azure APIs since then


pavannd1 (commented):
LGTM. Merge after downstream tests.

julio-lopez (approved):

PR #2296, opened 2023-08-31
Add functionality to create Kanister job pods with custom labels and annotations

viveksinghggits:
## Change Overview

Up until now there wasn't a way to set labels and annotations to the pod that kanister action set creates for the functions that are configured in the Kanister blueprint.
This PR introduces new helm fields that can be used to set custom labels and annotations to the Kanister job pods that get crated by Kanister actionset.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

```
go test -check.f "UtilsTestSuite"
OK: 5 passed
PASS
ok  	github.com/kanisterio/kanister/pkg/function	0.036s
```

Install Kanister using below command

```
helm install kanister -n kanister helm/kanister-operator --create-namespace --set controller.kanisterPodCustomLabels="a=b\,c=d" --set controller.kanisterPodCustomAnnotations="e=f\,g=h"
```
and look at the label and annotations for the KubeTask pod

```
 k get pods -n mysql kanister-job-wh8rm -oyaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    e: f
    g: h
  creationTimestamp: "2023-08-31T14:14:43Z"
  generateName: kanister-job-
  labels:
    a: b
    c: d
    createdBy: kanister
  name: kanister-job-wh8rm
  namespace: mysql
  resourceVersion: "134753"
  uid: 42e997b9-0209-4856-8db9-fdb2a9880d6d
spec:
```

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/helm/kanister-operator/values.yaml
+++ b/helm/kanister-operator/values.yaml
@@ -37,9 +37,9 @@ controller:
     # false : kanister-prometheus framework has been disabled
     # true: kanister-prometheus framework has been enabled
     enabled: false
-  # labels to set on the pods created by kanister functions
+  # labels to set on the pods created by kanister functions, labels (key=value) are separated by ,
   kanisterPodCustomLabels : ""
-  # annotations to set on the pods created by kanister functions
+  # annotations to set on the pods created by kanister functions, annotations (key=value) are separated by ,
   kanisterPodCustomAnnotations : ""
 bpValidatingWebhook:
   enabled: true
--- a/helm/kanister-operator/values.yaml
+++ b/helm/kanister-operator/values.yaml
@@ -37,6 +37,10 @@ controller:
     # false : kanister-prometheus framework has been disabled
     # true: kanister-prometheus framework has been enabled
     enabled: false
+  # labels to set on the pods created by kanister functions
+  kanisterPodCustomLabels : ""
+  # annotations to set on the pods created by kanister functions
+  kanisterPodCustomAnnotations : ""
 bpValidatingWebhook:
   enabled: true
   tls:
--- a/pkg/function/utils.go
+++ b/pkg/function/utils.go
@@ -3,6 +3,7 @@ package function
 import (
 	"bytes"
 	"context"
+	"os"
 	"path"
 	"strings"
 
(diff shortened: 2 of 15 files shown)


helm/kanister-operator/values.yaml
```diff
@@ -37,6 +37,10 @@ controller:
     # false : kanister-prometheus framework has been disabled
     # true: kanister-prometheus framework has been enabled
     enabled: false
+  # labels to set on the pods created by kanister functions
+  kanisterPodCustomLabels : ""
```
PrasadG193, line 41: What is the value format? Is it a map of key-value or comma-separated pair?
We should add a comment/example about that.

pkg/function/utils.go
```diff
@@ -27,8 +28,57 @@ import (
...
+}
+
+func getKanisterPodAnnotations() (bool, *map[string]string) {
+	return parseToLabelSelector(KanisterPodCustomAnnotationsEnv)
+}
+
+// SetLabelsToPodOptionsIfRequired sets labels to PodOptions
+func SetLabelsToPodOptionsIfRequired(options *kube.PodOptions) {
+	updateNeeded, labels := getKanisterPodLabels()
+	if updateNeeded {
```
PrasadG193, line 59: What are the cases when `updateNeeded` is false and `labels` is non-nil?

pkg/function/utils.go
```diff
@@ -27,8 +28,57 @@ import (
...
+		for k, v := range *labels {
+			options.Labels[k] = v
+		}
+	}
+}
+
+// SetAnnotationsToPodOptionsIfRequired sets annotations to PodOptions
+func SetAnnotationsToPodOptionsIfRequired(options *kube.PodOptions) {
+	updateNeeded, annotations := getKanisterPodAnnotations()
+	if updateNeeded {
```
PrasadG193, line 72: Same here, what are the scenarios when `updateNeeded` is false when `annotations` are non-nil

helm/kanister-operator/values.yaml
viveksinghggits: It's documented separately (part of this PR itself) as part of the docs.

viveksinghggits: https://github.com/kanisterio/kanister/pull/2296/commits/b20b725953509afe4417986faa47b2f6d1744b8f

pkg/function/utils.go
viveksinghggits: `updatedNeedeed` would be false if user has not provided labels/annotations for kanister annotation while installing Kanister. And labels would be non-nil if the kube.options struct that we have initialised doesn't have labels/annotations already specified.

PrasadG193: Do we really need `updateNeeded`? Can we just not check if labels is nil?
```suggestion
	labels := getKanisterPodLabels()
	if labels != nil {
```

helm/kanister-operator/values.yaml
PrasadG193: I think we should also add a comment here as well.

PrasadG193 (approved):
LGTM

PR #2310, opened 2023-09-07
Change the way we figure out container to read the artifacts from

viveksinghggits:
## Change Overview

While running the kanister function, to read the artifact that the phase has produced we get the logs of the kanister function pod (let's say `KubeTask`). If there is just one container in the KubeTask pod in that case things would work but if there are multiple contianers in the pod (let's say service mesh injected one), in that case the logic that we have to figure out the container name was not efficient. If was relying on the first container of the pod.

This PR changes that logic to read the container name from `podOptions` or fallback to the default container name. This should work in all the cases, if container name is not specified while creating `podOptions` we will use default container name everywhere otherwise we will use the container name that is specified in the podOptions.

## Pull request type

- [x] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/pod_controller.go
+++ b/pkg/kube/pod_controller.go
@@ -211,18 +211,12 @@ func (p *podController) StopPod(ctx context.Context, stopTimeout time.Duration, 
 	return nil
 }
 
-// getContainerName returns container name, which should be passed to
-// operations that require it.
-// If the container name was specified in podOptions, it will be used.
-// Otherwise, the first container name from specs will be taken as best effort
-// (when pods are created with sidecars, sidecar containers are placed after
-// main container).
-func (p *podController) getContainerName() string {
-	if p.podOptions.ContainerName != "" {
-		return p.podOptions.ContainerName
+func containerNameFromPodOpts(po *PodOptions) string {
+	if po.ContainerName != "" {
+		return po.ContainerName
 	}
 
-	return p.pod.Spec.Containers[0].Name
+	return defaultContainerName
 }
 
 func (p *podController) StreamPodLogs(ctx context.Context) (io.ReadCloser, error) {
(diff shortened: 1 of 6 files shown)


viveksinghggits:
@kale-amruta FYI, I have added some code related to repo server as well.

viveksinghggits:
@pavannd1 the main change is in `pod_controller.go` that needs to be reviewed. Rest of the change is just because of that.

pkg/kube/pod_controller.go
```diff
@@ -211,26 +211,20 @@ func (p *podController) StopPod(ctx context.Context, stopTimeout time.Duration,
...
-// getContainerName returns container name, which should be passed to
-// operations that require it.
-// If the container name was specified in podOptions, it will be used.
-// Otherwise, the first container name from specs will be taken as best effort
-// (when pods are created with sidecars, sidecar containers are placed after
-// main container).
-func (p *podController) getContainerName() string {
-	if p.podOptions.ContainerName != "" {
-		return p.podOptions.ContainerName
+func ContainerNameFromPodOpts(po *PodOptions) string {
```
pavannd1, line 214: If this is not limited to `podController`, we should move it to `pod.go`.
Additionally, I preferred the previous name. It was generic. It's getting the name from `PodOptions` only when available. So, `GetContainerName` or `ContainerNameFromPodOptsOrDefault` would be more appropriate.
The comment section should be modified and added back as well.

pkg/kube/pod_controller_test.go
```diff
@@ -309,3 +309,27 @@ func (s *PodControllerTestSuite) TestPodControllerGetCommandExecutorAndFileWrite
 		tc(pcp, pc)
 	}
 }
+
+func (s *PodControllerTestSuite) TestContainerNameFromPodOpts(c *C) {
```
pavannd1, line 313: If we move the helper in `pod.go`, this should be moved as well

pavannd1 (approved):

PR #2316, opened 2023-09-08
Revert "Revert "Change the way we figure out container to read the artifacts from""

viveksinghggits:
Reverts kanisterio/kanister#2314


ankitjain235 (approved):
LGTM

PR #2352, opened 2023-09-22
[WIP] chore(ci): speedup CI

julio-lopez:
## Change Overview

Leverages the GH cache managed by the @actions/setup-go action to speed up CI. This avoids the step for creating, uploading and downloading the artifacts.

The GH cache can be shared across jobs, workflows and multiple executions of workflows, be it in the same PR, separate PRs or branch push events.

This should be faster since caches are more easily reused across jobs and worker instances.

This approach also requires less storage space, since the cache can be easily reused across multiple executions, so it does not require creating artifacts per workflow execution.

## Pull request type

- [x] :robot: Test speedup

## Test Plan

- [x] :green_heart: CI workflow execution in GH

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -49,7 +49,7 @@ jobs:
         skip-cache: true
   test:
     runs-on: ubuntu-20.04
-    needs: [gomod, lint]
+    needs: gomod
     strategy:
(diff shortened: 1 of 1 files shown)


pavannd1:
@julio-lopez should we close this?

PR #2379, opened 2023-10-05
Running `make test` locally using `minio` & `minikube` fixed

k0taperk0t:
## Change Overview

This PR fixes running `make test` locally using `minio` & `minikube`.

**Note: This is draft just because these changes have to be based on https://github.com/kanisterio/kanister/pull/2377, but it's not merged yet.**

PR series:
* https://github.com/kanisterio/kanister/pull/2376
* https://github.com/kanisterio/kanister/pull/2366
* https://github.com/kanisterio/kanister/pull/2378
* https://github.com/kanisterio/kanister/pull/2377
* => https://github.com/kanisterio/kanister/pull/2379
* https://github.com/kanisterio/kanister/pull/2383
* https://github.com/kanisterio/kanister/pull/2386
* https://github.com/kanisterio/kanister/pull/2365

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

### DIFF
--- a/build/local_kubernetes.sh
+++ b/build/local_kubernetes.sh
@@ -73,14 +73,43 @@ install_csi_hostpath_driver() {
     kubectl apply -fhttps://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/{rbac-snapshot-controller.yaml,setup-snapshot-controller.yaml}
 
     # Deploy the CSI Hostpath Driver
-    cd /tmp
-    git clone https://github.com/kubernetes-csi/csi-driver-host-path.git
-    cd csi-driver-host-path
-    sed -i 's/mountPropagation: Bidirectional/\#mountPropagation: Bidirectional/g' deploy/kubernetes-latest/hostpath/csi-hostpath-plugin.yaml
-    ./deploy/kubernetes-latest/deploy.sh
+    pushd /tmp
+      git clone https://github.com/kubernetes-csi/csi-driver-host-path.git
+      pushd csi-driver-host-path
+        sed -i 's/mountPropagation: Bidirectional/\#mountPropagation: Bidirectional/g' deploy/kubernetes-latest/hostpath/csi-hostpath-plugin.yaml
 
-    # Create StorageClass
-    kubectl apply -f ./examples/csi-storageclass.yaml
+        ./deploy/kubernetes-latest/deploy.sh
+
+        # Create StorageClass
+        kubectl apply -f ./examples/csi-storageclass.yaml
+      popd
+    popd
+}
+
+check_csi_hostpath_driver_installed() {
+    # Check VolumeSnapshot CRDs are installed
+    if ! kubectl diff -fhttps://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/client/config/crd/snapshot.storage.k8s.io_{volumesnapshots.yaml,volumesnapshotclasses.yaml,volumesnapshotcontents.yaml} 2>&1 > /dev/null ; then
+        echo "VolumeSnapshot CRDs are not installed."
+        exit 1
+    fi
+
+    # Check snapshot controller created
+    if ! kubectl diff -fhttps://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/{rbac-snapshot-controller.yaml,setup-snapshot-controller.yaml} 2>&1 > /dev/null ; then
+        echo "Snapshot controller is not created."
+        exit 1
+    fi
+
+    # Deploy the CSI Hostpath Driver
+    pushd /tmp
+      git clone https://github.com/kubernetes-csi/csi-driver-host-path.git
+      pushd csi-driver-host-path
+        # Check StorageClass created
+        if ! kubectl diff -f ./examples/csi-storageclass.yaml 2>&1 > /dev/null ; then
+            echo "StorageClass is not created."
+            exit 1
+        fi
+      popd
+    popd
 }
 
 stop_localkube() {
@@ -158,6 +187,9 @@ EOM
 check_or_get_dependencies
 case "${1}" in
         # Alphabetically sorted
+        check_csi_hostpath_driver_installed)
+            time -p check_csi_hostpath_driver_installed
+            ;;
         get_localkube)
             time -p get_localkube
             ;;
--- a/build/run_container.sh
+++ b/build/run_container.sh
@@ -71,7 +71,6 @@ run_build_container() {
       -v "${PWD}:/go/src/${PKG}"                                  \
       -v "${PWD}/bin/${ARCH}:/go/bin"                             \
       -v "${PWD}/.go/std/${ARCH}:/usr/local/go/pkg/linux_${ARCH}" \
-      -v "${HOME}/.docker:/root/.docker"                          \
       -v /var/run/docker.sock:/var/run/docker.sock                \
       -w /go/src/${PKG}                                           \
       ${BUILD_IMAGE}                                              \
--- a/build/run_container.sh
+++ b/build/run_container.sh
@@ -52,9 +52,9 @@ run_build_container() {
   # In case of `minikube`, kube config stores full path to a certificates,
   # thus the simples way to get working minikube in the build container is
   # to bind original path to minikube settings to the container.
-  local path_to_minikube_dir="${HOME}/.minikube"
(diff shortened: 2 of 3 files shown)


k0taperk0t:

viveksinghggits:
> This PR fixes running tests locally using minio & minikube.

Hi @k0taperk0t ,
Can you please explain this a bit, what exactly are we fixing here? What do we mean by fixing running tests?

k0taperk0t:
>Can you please explain this a bit, what exactly are we fixing here? What do we mean by fixing running tests?

Hi @viveksinghggits!
I needed to add https://github.com/kanisterio/kanister/pull/2365 functionality to Kanister. Since I've never meet with Kanister, I've tried to follow [official documentation](https://github.com/kanisterio/kanister/blob/master/BUILD.md) and run tests, but I was not able to do this. I met few problems, such as:
* Unability to run even `make golint`, because of I have M1 Mac: https://github.com/kanisterio/kanister/pull/2366
* Unability to run tests locally because of `make test` every time returned "Please install MinIO using 'make install-minio' and try again." error even when `minio` has been installed, without any details: https://github.com/kanisterio/kanister/pull/2378

Even after fixing previous problems I met with failing of tests, if I try to run it locally.

Then I realized that to fix this I need to make changes for running docker containers, but I think it's a bad idea to make the same changes in a few places, so I did refactoring of running containers: https://github.com/kanisterio/kanister/pull/2377.

Thus, changes in this PR fixes running tests and further commits in this PR will fix failing of these test in local environment. But since these changes base on https://github.com/kanisterio/kanister/pull/2377, at this time this PR contains commits of https://github.com/kanisterio/kanister/pull/2377 as well. I can't rebase them in PR because then this PR will become "local" for my fork.

The main goal is to make ability to run tests locally to fix https://github.com/kanisterio/kanister/pull/2365.

build/local_kubernetes.sh
```diff
@@ -73,14 +73,43 @@ install_csi_hostpath_driver() {
...
+
+check_csi_hostpath_driver_installed() {
+    # Check VolumeSnapshot CRDs are installed
+    if ! kubectl diff -fhttps://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/client/config/crd/snapshot.storage.k8s.io_{volumesnapshots.yaml,volumesnapshotclasses.yaml,volumesnapshotcontents.yaml} 2>&1 > /dev/null ; then
+        echo "VolumeSnapshot CRDs are not installed."
+        exit 1
+    fi
+
+    # Check snapshot controller created
+    if ! kubectl diff -fhttps://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${SNAPSHOTTER_VERSION}/deploy/kubernetes/snapshot-controller/{rbac-snapshot-controller.yaml,setup-snapshot-controller.yaml} 2>&1 > /dev/null ; then
```
viveksinghggits, line 97: how would `diff` work if the snapshot controller  is already installed using different `SNAPSHOT_VERSION` and we have updated `SNAPSHOT_VERSION` to newer versions?
have we considered that, would that work as expected?

build/run_container.sh
```diff
@@ -49,19 +49,28 @@ run_build_container() {
       cmd=(/bin/bash)
   fi
 
+  # In case of `minikube`, kube config stores full path to a certificates,
+  # thus the simples way to get working minikube in the build container is
+  # to bind original path to minikube settings to the container.
+  local path_to_minikube_dir="${HOME}/.minikube"
```
viveksinghggits, line 55:
```suggestion
  local minikube_dir_path="${HOME}/.minikube"
```

viveksinghggits: I am not able to understand the use of minikube while starting the build container. Can you please describe that?
`run_build_container` should just be able to run a command on a container right? Or does it also need to know how the kubernetes is setup.
If it needs to know how Kubernetes is setup, how does it work for kind?

build/run_container.sh
```diff
@@ -49,19 +49,28 @@ run_build_container() {
...
       --net host                                                  \
       -e GITHUB_TOKEN="${github_token}"                           \
+      ${minikube_dir_binding}                                     \
       -v "${HOME}/.kube:/root/.kube"                              \
       -v "${PWD}/.go/pkg:/go/pkg"                                 \
       -v "${PWD}/.go/cache:/go/.cache"                            \
       -v "${PWD}:/go/src/${PKG}"                                  \
       -v "${PWD}/bin/${ARCH}:/go/bin"                             \
       -v "${PWD}/.go/std/${ARCH}:/usr/local/go/pkg/linux_${ARCH}" \
-      -v "${HOME}/.docker:/root/.docker"                          \
```
viveksinghggits, line 64: Is removing the docker socket here intentional if yes, what would happen if the command that is being run in build container creates another container? Where would that container be created.

build/test.sh
```diff
@@ -63,9 +64,18 @@ check_dependencies() {
...
         exit 1
     fi
+
+    # A test (CRDSuite) that runs as part of `make test` requires at least one CRD to
+    # be present on the cluster. That's why we are checking that `csi-hostpath-driver`
+    # installed before running tests.
+    if ! ${SCRIPT_DIR}/local_kubernetes.sh check_csi_hostpath_driver_installed ; then
+        echo "Please install CSI hostpath driver using 'make install-csi-hostpath-driver' and try again."
+        exit 1
+    fi
```
viveksinghggits, line 74: Should is say that CSI hostpath driver is not installed or it should say CRDs are not installed in the cluster, but a test requires at least one CRD to be available.

viveksinghggits: something like

```
CRDs are not installed on the cluster but a test (CRDSuite) requires at least one CRD to be available on the cluster. One can be installed by running make install-csi-hostpath-driver command.
```

build/local_kubernetes.sh
k0taperk0t: Well, it's good question. But now we don't check existent of snapshot controller at all, thus people most likely don't even know that it have to be installed before running a tests. If version of snapshot controller is set somewhere in YAML and already installed version differs from the actual version, the `diff` will treat it as difference, so this check gonna be fail.

build/run_container.sh
k0taperk0t: In case of `minikube` in `kubeconfig` stored full paths to the `minio`'s certificates. It's mentioned in https://github.com/kanisterio/kanister/pull/2378:

> Error: Kubernetes cluster unreachable: invalid configuration: [unable to read client-cert /Users/sergey.aksenov/.minikube/profiles/minikube/client.crt for minikube due to open /Users/sergey.aksenov/.minikube/profiles/minikube/client.crt: no such file or directory, unable to read client-key /Users/sergey.aksenov/.minikube/profiles/minikube/client.key for minikube due to open /Users/sergey.aksenov/.minikube/profiles/minikube/client.key: no such file or directory, unable to read certificate-authority /Users/sergey.aksenov/.minikube/ca.crt for minikube due to open /Users/sergey.aksenov/.minikube/ca.crt: no such file or directory]

and in comment above this variable.

build/run_container.sh
k0taperk0t: With this change, if the command that is being run in build container creates another container, a run will successful in comparing with current behavior.

viveksinghggits: sorry I didn't  understand this `a run will successful in comparing with current behavior.`

build/run_container.sh
viveksinghggits: does this mean the tests were not running on minkube earlier they were just supported on kind clusters? And after this change they would also run on minikube?

k0taperk0t: Currently if you'll try to run some command within container, which will try to run another container, you'll get `ERROR: Cannot connect to the Docker daemon at unix:///path_to_socket. Is the docker daemon running?` error, it has been fixed in second commit of this PR. Exactly because of binding `~/.docker` directory.

build/run_container.sh
k0taperk0t: Currently if you'll try to run some command within container, which will try to run another container, you'll get `ERROR: Cannot connect to the Docker daemon at unix:///path_to_socket. Is the docker daemon running?` error, it has been fixed in second commit of this PR. Exactly because of binding `~/.docker` directory.

build/run_container.sh
k0taperk0t: Exactly :)

build/test.sh
k0taperk0t: fixed

build/run_container.sh
k0taperk0t: >  local minikube_dir_path="${HOME}/.minikube"

fixed

viveksinghggits (approved):

PR #2384, opened 2023-10-09
Update ubi-minimal base image to ubi-minimal:9.2-750.1696515534

infraq:
Update ubi-minimal base image to ubi-minimal:9.2-750.1696515534


pavannd1:
Already merged #2391

PR #2430, opened 2023-10-24
Fix the way action set name was being configured

### LINKED ISSUE #2285: Dynamic Naming of Actionsets using kanctl
The `kanctl` Tool should be able to allow dynamic naming. At the moment 'kanctl' always sets the 'generateName: ' field in the metadata of the Actionset to the value of 'backup-'. This results in an Actionset having a generated name with `backup-#####`. Ideally, there should be a way to change the value of the generateName field for example by using a flag like `--name {NAME}`.

**Describe alternatives you've considered**
- Dynamic Naming using the CLI
- Generated Name using other information like name of application, time/date of backup, ... e.g. backup-postgres-2023-01-01-23-59

viveksinghggits:
## Change Overview

Fixes: https://github.com/kanisterio/kanister/issues/2285

`kanctl` was able to configure the actionset name using the flag called `action-set`.
But it had two problems
1. The better flag to give actionset name would be `--name` instead of `--action-set`. Because there is another flag called `action`, so it gets confusing between `--action-set` and `--action`. This PR renames the flag to provide the actionset name to `--name`.
3. Even though user would expect that if they have provided `--name` the actionset would be named exactly that. But the problem was we were using `generateName` in code that resulted in a string being added in the name. This PR makes sure that if `--name` is provided, the actionset would have exactly this name.

## Pull request type

- [x] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2285 

## Test Plan

```
go test 
OK: 3 passed
PASS
ok  	github.com/kanisterio/kanister/pkg/kanctl	0.031s
```

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

```
kanctl create actionset --action backup --namespace kanister --blueprint mysql-blueprint --statefulset mysql/mysql-release --profile mysql/s3-profile-qgbmj --dry-run
kind: ActionSet
metadata:
  creationTimestamp: null
  name: backup-wvb4r
```

```
kanctl create actionset --action backup --namespace kanister --blueprint mysql-blueprint --statefulset mysql/mysql-release --profile mysql/s3-profile-qgbmj --name=newname --dry-run
kind: ActionSet
metadata:
  creationTimestamp: null
  name: newname
spec:
```

```
kanctl create actionset --action restore --from=backup-6cztn --dry-run --profile mysql/s3-profile-qgbmj -n kanister
kind: ActionSet
metadata:
  creationTimestamp: null
  name: restore-backup-6cztn-cmlrl
spec:
```


pkg/kanctl/actionset.go
```diff
@@ -190,7 +191,7 @@ func newActionSet(params *PerformParams) (*crv1alpha1.ActionSet, error) {
 
 	return &crv1alpha1.ActionSet{
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: name,
```
PrasadG193, line 193: Should we also have a way to generate name like we have right now? It this is getting used in automation where they use same name to create actionset, that would not break.

PrasadG193 (commented):
LGTM

pkg/kanctl/actionset.go
viveksinghggits: If actionset is being created using automation in that case users can just skip the `--name` flag, and we would generate the name for them.

pkg/kanctl/actionset_test.go
```diff
@@ -102,6 +102,12 @@ func (k *KanctlTestSuite) TestGenerateActionSetName(c *C) {
 
 		actual, err := generateActionSetName(params)
 		c.Assert(err, DeepEquals, tc.expectedErr)
-		c.Assert(actual, DeepEquals, tc.expected)
+		if tc.actionSetName != "" || tc.expected == "" {
+			// if --name is provided we just use that we dont derive name
+			c.Assert(actual, DeepEquals, tc.expected)
+		} else {
+			// randomg 5 chars are added at the end if name is derived by us
```
e-sumin, line 109:
```suggestion
			// random 5 chars are added at the end if name is derived by us
```

e-sumin (approved):

pkg/kanctl/actionset.go
viveksinghggits: @PrasadG193 I am kueueing this.

PrasadG193: Cool, thanks

PR #2447, opened 2023-10-29
Update kopia to the latest version

KastenMike:
This PR updates to the latest kopia + implements a new interface method (the same way as it is done in kopia)


pkg/virtualfs/directory.go
```diff
@@ -70,6 +70,11 @@ func (d *Directory) Child(ctx context.Context, name string) (fs.Entry, error) {
 	return fs.IterateEntriesAndFindChild(ctx, d, name)
 }
 
+// Iterate returns directory iterator.
+func (d *Directory) Iterate(ctx context.Context) (fs.DirectoryIterator, error) {
```
KastenMike, line 74: I think the previous `IterateEntries` can be removed. It is no longer part of the Kopia interface and there's no references to it in the repo. lmk if you agree @pavannd1 / @ankitjain235

ankitjain235: I don't think we were using the function `IterateEntries` directly. It was being used internally from kopia. If it's no longer there in upstream kopia, I think it should be okay to remove it.

KastenMike: removed. Kopia replaced it with `Iterate`

pavannd1 (approved):
LGTM.
This will also resolve the GHSA-m425-mq94-257g vulnerability.

leuyentran (approved):
LGTM

PR #2449, opened 2023-10-30
Set soft limit for content cache while connecting to server in kopia client options

kale-amruta:
## Change Overview

Currently while connecting to kopia server using client we are setting only hard limits in the connectOptions . But [this](https://github.com/kopia/kopia/blob/master/repo/api_server_repository.go#L353) kopia function which is called while connecting to server has the logic of setting caching as specified [here](https://github.com/kopia/kopia/blob/master/repo/caching.go#L42). It checks if the soft limit for contenct cache is set. If it is not set, caching options are not initalized and is returned as empty struct. Hence hard limits for cache can never be set while connecting to server. This makes it compulsory to set soft limit for the content cache while connecting to server

This PR sets soft limit of content cache(It uses same value as hard value)

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pavannd1:
@julio-lopez Could you please take a look at kopia upstream and confirm if it's a bug or intentionally done?

pavannd1 (approved):

pkg/kopia/repository/client.go
```diff
@@ -64,6 +64,7 @@ func ConnectToAPIServer(
 		CachingOptions: content.CachingOptions{
 			CacheDirectory:              kopia.DefaultClientCacheDirectory,
 			ContentCacheSizeLimitBytes:  int64(contentCacheMB << 20),
+			ContentCacheSizeBytes:       int64(contentCacheMB << 20),
```
pavannd1, line 67:
```suggestion
			ContentCacheSizeBytes:       int64(contentCacheMB << 20),
			MetadataCacheSizeBytes:      int64(metadataCacheMB << 20),
```
Let's set both for now. No harm

PR #2505, opened 2023-12-01
Enable stylecheck linter

ankitjain235:
## Change Overview

This PR enables stylecheck golang linter

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


.golangci.yml
```diff
@@ -75,4 +76,6 @@ linters-settings:
     sections:
       - standard
       - default
-      - prefix(github.com/kanisterio/kanister)
```
PrasadG193, line 78: Extra newlines?

pkg/controller/metrics.go
```diff
@@ -27,26 +27,26 @@ type metrics struct {
 }
 
 const (
-	ACTION_SET_COUNTER_VEC_LABEL_RES         = "resolution"
-	ACTION_SET_COUNTER_VEC_LABEL_RES_SUCCESS = "success"
-	ACTION_SET_COUNTER_VEC_LABEL_RES_FAILURE = "failure"
+	ACTION_SET_COUNTER_VEC_LABEL_RES         = "resolution" //nolint:stylecheck // Already defined in docs
+	ACTION_SET_COUNTER_VEC_LABEL_RES_SUCCESS = "success"    //nolint:stylecheck // Already defined in docs
+	ACTION_SET_COUNTER_VEC_LABEL_RES_FAILURE = "failure"    //nolint:stylecheck // Already defined in docs
```
PrasadG193, line 32: What is the error we are getting?

pkg/kopia/command/maintenance.go
```diff
@@ -16,7 +16,7 @@ package command
 
 type MaintenanceInfoCommandArgs struct {
 	*CommandArgs
-	GetJsonOutput bool
+	GetJsonOutput bool //nolint:stylecheck // Validate workflows before enabling stylecheck
```
PrasadG193, line 19: What is the error we are getting here? Why do we need to ignore this?

pkg/kopia/command/storage/s3_args.go
```diff
@@ -38,7 +38,7 @@ func s3Args(location map[string][]byte, repoPathPrefix string) logsafe.Cmd {
 		s3Endpoint := ResolveS3Endpoint(e)
 		args = args.AppendLoggableKV(s3EndpointFlag, s3Endpoint)
 
-		if HttpInsecureEndpoint(e) {
+		if HTTPInsecureEndpoint(e) {
```
PrasadG193, line 41: If its not used outside the pkg, can we make it internal/private?

pkg/kube/pod.go
```diff
@@ -442,9 +442,9 @@ func patchDefaultPodSpecs(defaultPodSpecs corev1.PodSpec, override crv1alpha1.JS
 }
 
 // CreateAndMergeJsonPatch uses Strategic Merge to merge two Pod spec configuration
-func CreateAndMergeJsonPatch(original, override crv1alpha1.JSONMap) (crv1alpha1.JSONMap, error) {
+func CreateAndMergeJsonPatch(original, override crv1alpha1.JSONMap) (crv1alpha1.JSONMap, error) { //nolint:stylecheck // Validate workflows before enabling stylecheck
```
PrasadG193, line 445: Why do we need to ignore the check?

pkg/version/version.go
```diff
@@ -19,9 +19,10 @@ package version
 import "fmt"
 
 // The below variables will be overrriden using ldflags set by goreleaser during the build process
+
 var VERSION = "DEV"
-var GIT_COMMIT = "NONE"
-var BUILD_DATE = "UNKNOWN"
+var GIT_COMMIT = "NONE"    //nolint:stylecheck // Validate goreleaser works find after the change
+var BUILD_DATE = "UNKNOWN" //nolint:stylecheck // Validate goreleaser works find after the change
```
PrasadG193, line 25: Why do we need to ignore the checks? We can add the reason in the comment?

PrasadG193: `// Validate goreleaser works find after the change` Is it a TODO comment? Something we are supposed to do as a part of this PR? I am assuming it will be removed once we validate that, right>

pkg/kopia/command/policy_show_global.go
```diff
@@ -16,7 +16,7 @@ package command
 
 type PolicyShowGlobalCommandArgs struct {
 	*CommandArgs
-	GetJsonOutput bool
+	GetJsonOutput bool //nolint:stylecheck // Validate workflows before enabling stylecheck
```
PrasadG193, line 19: @ankitjain235 Is it a todo comment? The explanation is not clear

pkg/controller/metrics.go
pavannd1: Could you double-check if the metrics setting is enabled by default? Also, the label values are going to be the same right? We would only be changing the constant used to refer to them

pavannd1:
Not stale

pkg/version/version.go
saima-s: `GIT_COMMIT` and `BUILD_DATE` is referred only in VersionString functions [here](https://github.com/kanisterio/kanister/pull/2505/files#diff-73b8ec8a05213857da6e9598e34c5d7a18a6e5349250908ae6710578236977f2L28) and its value being used. Do we really need to test if goreleaser works fine? Beacuse value as `NONE` and `UNKNOWN` will be used in reference to the var names.

ankitjain235: These fields do seen to be used from goreleaser https://github.com/kanisterio/kanister/blob/1ebea1abd67ab4c8070a4385d9609f125438c33e/.goreleaser.yml#L11-L13

saima-s: ACK.Checking the testing method.

pkg/kopia/command/policy_show_global.go
saima-s: refactored

pkg/controller/metrics.go
saima-s: checked. Only `ACTION_SET_COUNTER_VEC_LABEL_RES` variable is used in documentation, others are not referenced anywhere but code. Refactored those.

pkg/validate/validate.go
```diff
@@ -72,7 +72,7 @@ func actionSpec(s crv1alpha1.ActionSpec) error {
 		// Not a known type. ActionSet must specify API group and resource
 		// name in order to populate `Object` TemplateParam
 		if s.Object.APIVersion == "" || s.Object.Resource == "" {
-			return errorf(validateErr, "Not a known object Kind %s. Action %s must specify Resource name and API version", s.Object.Kind, s.Name)
+			return errorf(errValidate, "Not a known object Kind %s. Action %s must specify Resource name and API version", s.Object.Kind, s.Name)
```
viveksinghggits, line 75: does `validateErr` not pass but `errValidate` pass?

pkg/controller/metrics.go
```diff
@@ -27,46 +27,46 @@ type metrics struct {
 }
 
 const (
-	ACTION_SET_COUNTER_VEC_LABEL_RES         = "resolution"
-	ACTION_SET_COUNTER_VEC_LABEL_RES_SUCCESS = "success"
-	ACTION_SET_COUNTER_VEC_LABEL_RES_FAILURE = "failure"
+	ACTION_SET_COUNTER_VEC_LABEL_RES   = "resolution" //nolint:stylecheck // Already defined in docs
```
pavannd1, line 30: 1. Let's verify if the metrics are enabled by default controller install or through a separate helm flag.
2. If it's through a helm flag, then it would be okay to change the documentation.
3. If it's enabled by default, what is the impact of the change? We should have a plan for the transition - a release note about the breaking change and actual change in a subsequent release.
@viveksinghggits @hairyhum wdyt?

hairyhum: Metrics are enables via helm flag `metrics.enabled`, but we don't configure labels via helm, as I understand.
I don't see why we shouldn't change it. The variable name is not an input anywhere and I can only find it in `kanister_prometheus_integration.md` which describes how it works but does not suggest you use it.
Is there some other docs which mention this variable?

pkg/validate/validate.go
saima-s: yes, aacording to stylecheck lint(ST1012), it says `error var validateErr should have name of the form errFoo (stylecheck)`

pkg/controller/metrics.go
saima-s: In `kanister_prometheus_integration.md` there are other 2 variables `ACTION_SET_COUNTER_VEC_LABEL_OP_TYPE` and `ACTION_SET_COUNTER_VEC_LABEL_BLUEPRINT` , there is no occurrence of these 2 in codebase apart from this doc. I am not sure what is the use of those and whether to we should rename it the same way as we are doing it for `ACTION_SET_COUNTER_VEC_LABEL_RES` to `ActionSetCounterVecLabelRes`. It seems inconsistent in doc if we don't do.

hairyhum: It seems like `kanister-prometheus-integration.md` is outdated as we don't have `ACTION_SET_COUNTER_VEC_LABEL_OP_TYPE` in the code anymore. 
Maybe we should make it consistent with the code or don't have code in `md` files. We can do that in a separate PR of course.

pkg/app/couchbase.go
```diff
@@ -52,6 +52,8 @@ type CouchbaseDB struct {
 }
 
 // Last tested woking version "2.3.0"
+//
+//nolint:stylecheck
```
viveksinghggits, line 56: why did we have to add the `nolint` here?

viveksinghggits (changes requested):

pkg/consts/consts.go
```diff
@@ -47,6 +48,7 @@ const (
 )
 
 // These consts are used to query Repository server API objects
+
```
viveksinghggits, line 51: is this required? i would prefer is removed.

pkg/metrics/metrics.go
```diff
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
```
viveksinghggits, line 14: I see that in some source files we are adding a new line immediately before the `package` declaration and in some files not (example `aws_secrets.go`) why is that?

pkg/testing/integration_register.go
```diff
@@ -85,7 +85,7 @@ var _ = Suite(&CockroachDB{
 	},
 })
 
-// time-log app for csi volumesnapshot
+// TimeLogCSI registers time-log app for csi volumesnapshot.
```
viveksinghggits, line 88: I think these comments should be improved, these are just types they don't register the app right?

pkg/testing/integration_register.go
```diff
@@ -279,7 +279,7 @@ var _ = Suite(&MysqlDBDepConfig{
 	},
 })
 
-// MongoDB deployed on openshift cluster
+// MongoDBDepConfig deploys MongoDB on openshift cluster
```
viveksinghggits, line 282: It's just a type, it doesn't deploy the app.

viveksinghggits (changes requested):

pkg/app/couchbase.go
saima-s: This is because style check expects exported function should have description of what its doing.
```
ST1020: comment on exported function NewCouchbaseDB should be of the form "NewCouchbaseDB ...
```

pkg/metrics/metrics.go
saima-s: There is always a line after `// limitations under the License.`
incase of `aws_secrets.go`, there a package description ` Package repositoryserver validates the secret credentials for all location profiles.` written before `package` (for style check linter ST1000) which is a different issue to be solved in another PR(as I am not aware of all the package descriptions)

pkg/app/couchbase.go
viveksinghggits: I think instead of `nolint` we should just add `NewCouchbaseDB initialises an instance of Couchbase DB` or something in these line. We are anyway adding the comment of `last tested` and a new line after `last tested` doesn't make sense.

pkg/metrics/metrics.go
viveksinghggits: ok

pkg/testing/integration_register.go
saima-s: done

pkg/testing/integration_register.go
saima-s: done

pkg/consts/consts.go
saima-s: refactored.

pkg/app/couchbase.go
saima-s: done

pkg/testing/integration_register.go
```diff
@@ -115,7 +115,7 @@ var _ = Suite(&Maria{
 	},
 })
 
-// Elasticsearch app
+// Elasticsearch type is an app for elasticsearch for integration test..
```
viveksinghggits, line 118:
```suggestion
// Elasticsearch type is an app for elasticsearch for integration test.
```

viveksinghggits (approved):

design/kanister-prometheus-integration.md
```diff
@@ -215,7 +215,7 @@ controller/metrics.go
 
 ```golang
 const (
-	ACTION_SET_COUNTER_VEC_LABEL_RES     = "resolution"
```
PrasadG193, line 218: [nit] I think we should exclude `./design` directory from the link checks

pkg/secrets/repositoryserver/aws_secrets.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package repositoryserver validates the secret credentials for all location profiles.
```
PrasadG193, line 15: [question] Do we have a lint checker for comment on package name? I don't think that is necessary

pkg/version/version.go
```diff
@@ -19,12 +19,13 @@ package version
 import "fmt"
 
 // The below variables will be overrriden using ldflags set by goreleaser during the build process
-var VERSION = "DEV"
-var GIT_COMMIT = "NONE"
-var BUILD_DATE = "UNKNOWN"
+
+var Version = "DEV"
+var GitCommit = "NONE"
+var BuildDate = "UNKNOWN"
```
PrasadG193, line 25: This would break the release. We have to update ldflags in `.goreleaser` as per the changes. https://github.com/kanisterio/kanister/blob/master/.goreleaser.yml#L15

PrasadG193 (changes requested):

pkg/version/version.go
PrasadG193: Oh, right. My bad

PrasadG193 (approved):

pkg/secrets/repositoryserver/aws_secrets.go
saima-s: Yes, it is. ST1000(stylecheck linter). It says we should describe package about what its doing in any one file atleast. Although I have disabled this `ST1000` in this PR, but had already made this change, so decided to keep the description as it is and later on `ST1000` will be enabled as a part of another PR will add similar description for other packages as well.

viveksinghggits:
@saima-s let's merge 🚀 it.

saima-s:
> @saima-s let's merge 🚀 it.

Will merge with https://github.com/kastenhq/k10/pull/26456/files.

pkg/secrets/repositoryserver/aws_secrets.go
saima-s: @PrasadG193 Checked kubernates repository. Found few examples where this linter is handled:
1. https://github.com/kubernetes/kubernetes/blob/0fc167103189a4d462c3cc7a17b360c3d998f4bf/cmd/kube-apiserver/app/options/options.go#L17
2. https://github.com/kubernetes/kubernetes/blob/0fc167103189a4d462c3cc7a17b360c3d998f4bf/cmd/kube-apiserver/app/aggregator.go#L17
3. https://github.com/kubernetes/kubernetes/blob/0fc167103189a4d462c3cc7a17b360c3d998f4bf/cmd/kube-apiserver/apiserver.go#L17

design/kanister-prometheus-integration.md
saima-s: This is not checking linter on `./design` package. The change ActionSetCounterVecLabelRes is made because this is the variable referred in code, and there changes needs to be done.

PR #2506, opened 2023-12-01
ExecWithOption - get rid of stdout/stderr buffers

e-sumin:
## Change Overview

Invoker of `kube.ExecWithOptions` has the possibility to get the stdout/stderr streams by passing appropriate `io.Writer`s via `ExecOptions`. Since all its invokers do this, it's redundant to collect the same streams into internal buffers and return them as a string.

At the same time, we have `kube.Exec` which is simplified version of `kube.ExecWithOptions`. This `kube.Exec` is the only case when we need to collect stdout/stderr to a buffer and return them after execution is done. 

Therefore, let's move this buffering to `kube.Exec` and simplify `kube.ExecWithOptions`

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


hairyhum (commented):
The change makes sense. 
@pavannd1 @viveksinghggits but I wonder if we have any process (e.g. versioning) to notify the user of public API changes?

pkg/kube/exec.go
```diff
@@ -113,32 +120,20 @@ func ExecOutput(cli kubernetes.Interface, namespace, pod, container string, comm
...
+	return ExecWithOptions(cli, opts)
 }
 
-// ExecWithOptions executes a command in the specified container,
-// returning stdout, stderr and error. `options` allowed for
-// additional parameters to be passed.
-func ExecWithOptions(kubeCli kubernetes.Interface, options ExecOptions) (string, string, error) {
+// ExecWithOptions executes a command in the specified container, returning an error.
+// `options` allowed for additional parameters to be passed.
+func ExecWithOptions(kubeCli kubernetes.Interface, options ExecOptions) error {
```
viveksinghggits, line 128: After looking at the desc, I thought, are there use cases where some consumers would need `Exec` and others would need `ExecWithOptions`, can we not just expose one. So that it's easier to use these APIs?

viveksinghggits: LGTM otherwise.

PrasadG193 (approved):
LGTM

pkg/kube/exec.go
e-sumin: In the ideal case, of course, it would be better to use `ExecWithOptions` only. But, unfortunately, both are widely used.

Looking on the code, I figured out that `ExecOutput` is probably even more redundant.

I can do bigger refactoring and reduce number of execs, but it will be bigger change, so maybe we can do that by two or three steps ? This one, drops buffering in ExecWithOptions, Second one will remove `ExecOutput` and, if we will agree, that we'd like to do that, the third (and the biggest one) can get rid of pure `Exec` ?

viveksinghggits: Now, that I think about it again, I don't think we should block the PR in this discussion. I think names are sufficient enough to not cause any confusion.
If we want to exec with some options use `ExecWithOptions` otherwise use `Exec`.

PR #2508, opened 2023-12-01
Retry on transient error when waiting for snapshot to be ready

hairyhum:
## Change Overview

CSI snapshot controller might add errors to the snapshot status which it will recover from.

Make snapshotter.WaitOnReadyToUse retry (up to 100 times) on those errors. Backoff mechanism makes it so 100 retries is minutes, hopefuly should be enough for most cases.

Unfortunately CSI snapshotter uses strings to inform of error reason and does not provide error code or type when reporting, hence for now we use regexp to match on transient error. If CSI snapshotter uses better error format in the future, we can also change that.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

## Follow up

Need to run some integration tests and monitor future issues in case we need to adjust the backoff retries count.
Monitor https://github.com/kubernetes-csi/external-snapshotter/issues/748 as those improvements could reduce the chance of errors (probably won't eliminate them completely though)
Monitor https://github.com/kubernetes-csi/external-snapshotter/issues/970 if there are improvements to the format of error

### DIFF
--- a/pkg/kube/snapshot/snapshot.go
+++ b/pkg/kube/snapshot/snapshot.go
@@ -16,6 +16,7 @@ package snapshot
 
 import (
 	"context"
+	"regexp"
 
 	v1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	"github.com/pkg/errors"
@@ -152,4 +153,15 @@ func NewSnapshotter(kubeCli kubernetes.Interface, dynCli dynamic.Interface) (Sna
 		return NewSnapshotAlpha(kubeCli, dynCli), nil
 	}
 	return nil, errors.New("Snapshot resources not supported")
+}
+
+// We use regexp to match because errors written in vs.Status.Error.Message are strings
+// and we don't have any status code or other metadata in there.
+var transientErrorRegexp = regexp.MustCompile("the object has been modified; please apply your changes to the latest version and try again")
+
+// Use regexp to detect resource conflict error
+// If CSI snapshotter changes error reporting to use more structured errors,
+// we can improve this function to parse and recognise error codes or types.
+func isTransientError(err error) bool {
+	return transientErrorRegexp.MatchString(err.Error())
 }
--- a/pkg/kube/snapshot/snapshot_alpha.go
+++ b/pkg/kube/snapshot/snapshot_alpha.go
@@ -320,8 +320,10 @@ func (sna *SnapshotAlpha) CreateContentFromSource(ctx context.Context, source *S
 
 // WaitOnReadyToUse will block until the Volumesnapshot in namespace 'namespace' with name 'snapshotName'
 // has status 'ReadyToUse' or 'ctx.Done()' is signalled.
+// FIXME: merge this with beta wait
 func (sna *SnapshotAlpha) WaitOnReadyToUse(ctx context.Context, snapshotName, namespace string) error {
-	return poll.Wait(ctx, func(context.Context) (bool, error) {
+	retries := 100
+	return poll.WaitWithRetries(ctx, retries, isTransientError, func(context.Context) (bool, error) {
 		us, err := sna.dynCli.Resource(v1alpha1.VolSnapGVR).Namespace(namespace).Get(ctx, snapshotName, metav1.GetOptions{})
 		if err != nil {
 			return false, err
--- a/pkg/kube/snapshot/snapshot_beta.go
+++ b/pkg/kube/snapshot/snapshot_beta.go
@@ -305,7 +305,8 @@ func (sna *SnapshotBeta) WaitOnReadyToUse(ctx context.Context, snapshotName, nam
 }
 
 func waitOnReadyToUse(ctx context.Context, dynCli dynamic.Interface, snapGVR schema.GroupVersionResource, snapshotName, namespace string) error {
-	return poll.Wait(ctx, func(context.Context) (bool, error) {
+	retries := 100
+	return poll.WaitWithRetries(ctx, retries, isTransientError, func(context.Context) (bool, error) {
 		us, err := dynCli.Resource(snapGVR).Namespace(namespace).Get(ctx, snapshotName, metav1.GetOptions{})
 		if err != nil {
 			return false, err
--- a/pkg/kube/snapshot/snapshot_test.go
+++ b/pkg/kube/snapshot/snapshot_test.go
@@ -364,6 +364,157 @@ func (s *SnapshotTestSuite) TestVolumeSnapshotCloneFake(c *C) {
 	}
 }
 
+func (s *SnapshotTestSuite) TestWaitOnReadyToUse(c *C) {
+	snapshotName := "snap-1-fake"
+	volName := "pvc-1-fake"
+	scheme := runtime.NewScheme()
+	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "snapshot.storage.k8s.io", Version: "v1alpha1", Kind: "VolumeSnapshotClassList"}, &unstructured.UnstructuredList{})
+	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "snapshot.storage.k8s.io", Version: "v1beta1", Kind: "VolumeSnapshotClassList"}, &unstructured.UnstructuredList{})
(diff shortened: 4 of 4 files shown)


pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -318,24 +317,23 @@ func (sna *SnapshotAlpha) CreateContentFromSource(ctx context.Context, source *S
...
+	// Error can be set while waiting for creation
+	if vs.Status.Error != nil {
+		return false, errors.New(vs.Status.Error.Message)
+	}
+	return (vs.Status.ReadyToUse && vs.Status.CreationTime != nil), nil
+}
+
 // WaitOnReadyToUse will block until the Volumesnapshot in namespace 'namespace' with name 'snapshotName'
 // has status 'ReadyToUse' or 'ctx.Done()' is signalled.
+// FIXME: merge this with beta wait
```
pavannd1, line 334: I think we should eventually add a `stable` wait and merge both alpha and beta into it

pkg/kube/snapshot/snapshot_stable.go
```diff
@@ -176,7 +176,7 @@ func (sna *SnapshotStable) CreateContentFromSource(ctx context.Context, source *
 // WaitOnReadyToUse will block until the Volumesnapshot in 'namespace' with name 'snapshotName'
 // has status 'ReadyToUse' or 'ctx.Done()' is signalled.
 func (sna *SnapshotStable) WaitOnReadyToUse(ctx context.Context, snapshotName, namespace string) error {
-	return waitOnReadyToUse(ctx, sna.dynCli, VolSnapGVR, snapshotName, namespace)
+	return waitOnReadyToUse(ctx, sna.dynCli, VolSnapGVR, snapshotName, namespace, isReadyToUseBeta)
```
pavannd1, line 179: We should add a TODO to switch to the stable version of wait function

pkg/kube/snapshot/snapshot.go
```diff
@@ -153,3 +157,25 @@ func NewSnapshotter(kubeCli kubernetes.Interface, dynCli dynamic.Interface) (Sna
...
+var transientErrorRegexp = regexp.MustCompile("the object has been modified; please apply your changes to the latest version and try again")
+
+// Use regexp to detect resource conflict error
+// If CSI snapshotter changes error reporting to use more structured errors,
+// we can improve this function to parse and recognise error codes or types.
+func isTransientError(err error) bool {
+	return transientErrorRegexp.MatchString(err.Error())
+}
+
+func waitOnReadyToUse(ctx context.Context, dynCli dynamic.Interface, snapGVR schema.GroupVersionResource, snapshotName, namespace string, isReadyFunc func(*unstructured.Unstructured) (bool, error)) error {
```
pavannd1, line 172:
```suggestion
func waitOnReadyToUse(
	ctx context.Context,
	dynCli dynamic.Interface,
	snapGVR schema.GroupVersionResource,
	snapshotName,
	namespace string,
	isReadyFunc func(*unstructured.Unstructured) (bool, error),
) error {
```
Trying to slowly incorporate [this](https://thanos.io/tip/contributing/coding-style-guide.md/#only-two-ways-of-formatting-functionsmethods) coding style for readability improvements

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -318,24 +317,23 @@ func (sna *SnapshotAlpha) CreateContentFromSource(ctx context.Context, source *S
 	return nil
 }
 
+func isReadyToUseAlpha(us *unstructured.Unstructured) (bool, error) {
+	vs := v1alpha1.VolumeSnapshot{}
+	if err := TransformUnstructured(us, &vs); err != nil {
+		return false, err
+	}
+	// Error can be set while waiting for creation
+	if vs.Status.Error != nil {
```
viveksinghggits, line 326: Just wanted to confirm that the events are in `VolumeSnapshot` resource right, and not in the `VolumeSnapshotContent` resource?

hairyhum: Yes. I don't think we read `VolumeSnapshotContent` status anywhere in our code. But status errors are propagated by snapshot_controller to `VolumeSnapshot` if there is an error in `VolumeSnapshotContent` as I understand.

pkg/kube/snapshot/snapshot_alpha.go
hairyhum: I did actually merge the wait functions so this one is addressed.

pkg/kube/snapshot/snapshot.go
```diff
@@ -153,3 +157,32 @@ func NewSnapshotter(kubeCli kubernetes.Interface, dynCli dynamic.Interface) (Sna
...
+func waitOnReadyToUse(
+	ctx context.Context,
+	dynCli dynamic.Interface,
+	snapGVR schema.GroupVersionResource,
+	snapshotName,
+	namespace string,
+	isReadyFunc func(*unstructured.Unstructured) (bool, error),
+) error {
+	retries := 100
+	return poll.WaitWithRetries(ctx, retries, isTransientError, func(context.Context) (bool, error) {
```
ewhamilton, line 181: Question: Is this the right poll to use here? Why is 100 retries the correct value?

It appears to me that there is currently only 1 other file where Kanister code uses poll.WaitWithRetries.

It is more common to use `poll.Wait`, sometimes with the ctx set to have a timeout. Or to call `poll.WaitWithBackoff` and a maximum time set.

hairyhum: We could do that. The difference here is that we only retry on this specific type of error, while `poll.WaitWithBackoff` would not distinguish that.
The `100` value is pretty arbitrary just to avoid the infinite wait. We could use `poll.WaitWithBackoffWithRetries` to configure max time to wait as well, but I don't know what's the good value in this context.

Current setting for the context timeout is infinity because there's a long-running process. Inserting a new time limit here would mean potentially creating a new failure scenario. Setting retries would worst case result in similar behaviour to what we have right now.

ewhamilton: As @pavannd1 indicated we can check this in as is and iterate on the polling mechanism-- this will improve robustness as is.

We can retry only on a specific error when using poll.Wait or poll.WaitWithBackoff by making the check for that error within the function called by poll. See for example how an `apierrors.IsNotFound` error is handled in RepoServiceHandler.createService.

See https://github.com/kanisterio/kanister/blob/2756ffb60874c5194df3ad1bc0f959806b223d5c/pkg/controllers/repositoryserver/handler.go#L138

Yes, there are some subtle differences between terminating a poll with a timeout or by retry count-- in particular whether ctx expiration can affect a call made within the poll loop. But I don't think those differences matter when the call being made is just a Get.

Ultimately the retry count is going to be picked based on a some knowledge of how long it is reasonable to wait for the snapshot controller to resolve the transient. In my mind it is clearer to just express that as a time value instead of trying to calculate the retry count from desired time and the backoff parameters.

See the many instances of `poll.Wait(timeoutCtx,...` in pkg/app where an app-specific timeout is known and used. Or pkg/kube.WaitForPodReady.

ewhamilton (approved):
I'll approve as is-- it's better than what we have and correct if not ideal.

I do think switching to an explicit time limit on the poll will be a significant improvement in understandability and maintainability.

pkg/kube/snapshot/snapshot.go
ewhamilton: Upon further reflection, I've realized that the code as implemented has the following behavior:
* Will wait indefinitely for the controller to update the status to be complete or some form of error.
* Once it encounters a transient error it will poll a bounded number of times (with default backoff parameters)
* If a non-transient error is encountered it will return an error immediately.

Importantly, setting a timeout on the context at beginning of all polling will affect the first wait, the wait for controller to ready snapshot before setting any error, transient or not.

While I still think it would be better to have an explicit time limit on retrying of transients, and we probably could work out a way to build it into a retry function closure, at that point it is no longer simple and clear.

Best solution is probably to keep retry count on trasients and add a comment about expected time that count corresponds to using default backoff.

PR #2533, opened 2023-12-13
Set original replica count as op artifact for `ScaleWorkload` fun

### LINKED ISSUE #2443: `ScaleWorkload` should set the original replica count as output artifact
If `ScaleWorkload` function is being used in one of the blueprint's phase, the phase that scales up the workload needs to know the original replica count.
One way to do this is setting the replica count as output artifact manually from one of the previous phases. We are doing this in some of the blueprints. https://github.com/kanisterio/kanister/blob/master/examples/cassandra/cassandra-blueprint.yaml#L116

 

It would be a good idea if `ScaleWorkload` is designed in such a way that when we scale up the workload the replicas can be figured out automatically.

viveksinghggits:
## Change Overview

This PR updates the `ScaleWorkload` function in such a way that it now outputs the original replica count of the workload
as output artifact.
After this change, users don't need to figure out the replica count of the workload and set that as output artifact using kando. This will simplify how we use `ScaleWorkload` function in blueprints.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2443 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/docs/functions.rst
+++ b/docs/functions.rst
@@ -147,7 +147,9 @@ Example:
 ScaleWorkload
 -------------
 
-ScaleWorkload is used to scale up or scale down a Kubernetes workload.
+ScaleWorkload is used to scale up or scale down a Kubernetes workload. It
+also sets the original replica count of the workload as output artifact with
+the key ``originalReplicaCount``.
 The function only returns after the desired replica state is achieved:
 
 * When reducing the replica count, wait until all terminating pods
@@ -155,7 +157,8 @@ The function only returns after the desired replica state is achieved:
 
 * When increasing the replica count, wait until all pods are ready.
 
-Currently the function supports Deployments and StatefulSets.
+Currently the function supports Deployments, StatefulSets and
+DeploymentConfigs.
 
 It is similar to running
 
@@ -165,7 +168,8 @@ It is similar to running
 
 This can be useful if the workload needs to be shutdown before processing
 certain data operations. For example, it may be useful to use ``ScaleWorkload``
-to stop a database process before restoring files.
+to stop a database process before restoring files. See
+:ref:`scaleworkloadexample` for an example with new ``ScaleWorkload`` function.
 
 .. csv-table::
    :header: "Argument", "Required", "Type", "Description"
--- a/docs/tasks/scaleworkload.rst
+++ b/docs/tasks/scaleworkload.rst
@@ -1,0 +1,50 @@
+.. _scaleworkloadexample:
+
+Using ScaleWorkload function with output artifact
+-------------------------------------------------
+
+Traditionally, when ``ScaleWorkload`` functions were used to scale down the
+workload, it was required to get the original replica count of the workload and
+set it as output artifact. So that it can later be used to scale up the
+workload to same number of replicas.
+
+After the new changes, the ``ScaleWorkload`` function automatically sets the
+original replica count of the workload as output artifact, which makes using
+``ScaleWorkload`` function in blueprints a lot easier.
+
+Below is an example of how this function can be used
+
+
+.. code-block:: yaml
+
+  apiVersion: cr.kanister.io/v1alpha1
+  kind: Blueprint
+  metadata:
+    name: my-blueprint
+  actions:
+    backup:
+      outputArtifacts:
+        backupOutput:
+          keyValue:
+            origReplicas: "{{ .Phases.shutdownPod.Output.originalReplicaCount }}"
+      phases:
+      # before scaling replicas 0, ScaleWorkload will get the original replica count
+      # to set that as output artifact (originalReplicaCount)
+      - func: ScaleWorkload
+        name: shutdownPod
+        args:
+          namespace: "{{ .StatefulSet.Namespace }}"
+          name: "{{ .StatefulSet.Name }}"
+          kind: StatefulSet
+          replicas: 0 # this is the replica count, the STS will scaled to
+    restore:
+      inputArtifactNames:
+        - backupOutput
+      phases:
+      - func: ScaleWorkload
+        name: bringUpPod
+        args:
+          namespace: "{{ .StatefulSet.Namespace }}"
+          name: "{{ .StatefulSet.Name }}"
+          kind: StatefulSet
+          replicas: "{{ .ArtifactsIn.backupOutput.KeyValue.origReplicas }}"
--- a/docs/tasks/scaleworkload.rst
+++ b/docs/tasks/scaleworkload.rst
@@ -3,14 +3,10 @@
 Using ScaleWorkload function with output artifact
 -------------------------------------------------
 
-Traditionally, when ``ScaleWorkload`` functions were used to scale down the
-workload, it was required to get the original replica count of the workload and
-set it as output artifact. So that it can later be used to scale up the
-workload to same number of replicas.
-
-After the new changes, the ``ScaleWorkload`` function automatically sets the
-original replica count of the workload as output artifact, which makes using
-``ScaleWorkload`` function in blueprints a lot easier.
+``ScaleWorkload`` function can be used to scale a workload to specified
+replicas. It automatically sets the original replica count of the workload
+as output artifact, which makes using ``ScaleWorkload`` function in blueprints
+a lot easier.
 
 Below is an example of how this function can be used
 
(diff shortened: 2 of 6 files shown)


docs/functions.rst
```diff
@@ -147,15 +147,18 @@ Example:
 ScaleWorkload
 -------------
 
-ScaleWorkload is used to scale up or scale down a Kubernetes workload.
+ScaleWorkload is used to scale up or scale down a Kubernetes workload. It
+also sets the original replica count of the workload as output artifact with
+the key ``originalReplicaCount``.
```
PrasadG193, line 152: Just wanted to confirm - it won't break older BP, right? I don't see any breaking changes.

docs/tasks/scaleworkload.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Using ScaleWorkload function with output artifact
+-------------------------------------------------
+
+Traditionally, when ``ScaleWorkload`` functions were used to scale down the
+workload, it was required to get the original replica count of the workload and
+set it as output artifact. So that it can later be used to scale up the
+workload to same number of replicas.
+
+After the new changes, the ``ScaleWorkload`` function automatically sets the
```
PrasadG193, line 11: IMO it doesn't make sense or add any value mentioning history about how it used to behave earlier.

docs/tasks/scaleworkload.rst
```diff
@@ -0,0 +1,50 @@
...
+  actions:
+    backup:
+      outputArtifacts:
+        backupOutput:
+          keyValue:
+            origReplicas: "{{ .Phases.shutdownPod.Output.originalReplicaCount }}"
+      phases:
+      # before scaling replicas 0, ScaleWorkload will get the original replica count
+      # to set that as output artifact (originalReplicaCount)
+      - func: ScaleWorkload
```
PrasadG193, line 33: Ideally, we should have versioning mechanism for different varients of the function instead of overriding existing functionality. But that's separate discussion

pkg/kube/workload.go
```diff
@@ -514,3 +514,28 @@ func IsPodRunning(cli kubernetes.Interface, podName, podNamespace string) (bool,
 
 	return true, nil
 }
+
+func StatefulSetReplicas(ctx context.Context, kubeCli kubernetes.Interface, namespace, name string) (int32, error) {
+	sts, err := kubeCli.AppsV1().StatefulSets(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return 0, errors.Wrapf(err, "Could not get STS{Namespace %s, Name: %s}, to figure out replicas", namespace, name)
```
PrasadG193, line 521:
```suggestion
		return 0, errors.Wrapf(err, "Could not get StatefulSet{Namespace %s, Name: %s}, to figure out replicas", namespace, name)
```

pkg/kube/workload.go
```diff
@@ -514,3 +514,28 @@ func IsPodRunning(cli kubernetes.Interface, podName, podNamespace string) (bool,
...
+	}
+	return *d.Spec.Replicas, nil
+}
+
+func DeploymentConfigReplicas(ctx context.Context, osCli osversioned.Interface, namespace, name string) (int32, error) {
+	dc, err := osCli.AppsV1().DeploymentConfigs(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return 0, errors.Wrapf(err, "Could not get DeploymentConfig{Namespace %s, Name: %s}, to figure out replicas", namespace, name)
+	}
+
```
PrasadG193, line 539:
```suggestion
```

pkg/kube/workload.go
```diff
@@ -514,3 +514,28 @@ func IsPodRunning(cli kubernetes.Interface, podName, podNamespace string) (bool,
 
 	return true, nil
 }
+
+func StatefulSetReplicas(ctx context.Context, kubeCli kubernetes.Interface, namespace, name string) (int32, error) {
+	sts, err := kubeCli.AppsV1().StatefulSets(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return 0, errors.Wrapf(err, "Could not get STS{Namespace %s, Name: %s}, to figure out replicas", namespace, name)
+	}
+	return *sts.Spec.Replicas, nil
```
PrasadG193, line 523: Should we check if `sts.Spec.Replicas` is nil? What value do we get if workloads are not in `Running` state?

docs/tasks/scaleworkload.rst
viveksinghggits: ok, makes sense.
I wanted to mention that just so that, if people come across a blueprint or have a blueprint that is setting replica count in legacy way, they should know why that's happening.
But I can remove it from here, if you think that's better.

docs/functions.rst
viveksinghggits: No it won't break older BP.

docs/tasks/scaleworkload.rst
viveksinghggits: yeah, separate discussion and  versioning would sense, if there are no ways of consuming this function in older way and someone wants to consume the function using older way. In that case version would be useful to make sure that someone is able to specify the version and can use the older behaviour of the function.
But, in this situation people can chose to use `ScaleWorkload` in any way without the need of versioning.

pkg/kube/workload.go
viveksinghggits: I can add the check, but should that not be the responsibility of the function that is actually doing the scaling?
Also, even if the workload is not in running state, replicas count would be set.

```
k get pods -n default 
NAME                    READY   STATUS             RESTARTS   AGE
nginx-85f656985-pbdfs   0/1     ErrImagePull       0          10s
nginx-85f656985-rk8hn   0/1     ErrImagePull       0          10s
nginx-85f656985-svxcc   0/1     ImagePullBackOff   0          47s

k get deployments.apps -n default 
NAME    READY   UP-TO-DATE   AVAILABLE   AGE
nginx   0/3     3            0           90s

k get deployments.apps -n default nginx -ojsonpath='{.spec.replicas}'
3
```

PrasadG193: Got it. It is a field of `spec`. Somehow I assumed that we are reading it from `status`. Thanks

pkg/function/scale_workload.go
```diff
@@ -80,15 +82,33 @@ func (s *scaleWorkloadFunc) Exec(ctx context.Context, tp param.TemplateParams, a
...
 	switch strings.ToLower(kind) {
 	case param.StatefulSetKind:
-		return nil, kube.ScaleStatefulSet(ctx, cli, namespace, name, replicas, waitForReady)
+		count, err := kube.StatefulSetReplicas(ctx, cli, namespace, name)
+		if err != nil {
+			return nil, err
+		}
+		return map[string]interface{}{
+			outputArtifactOriginalReplicaCount: count,
+		}, kube.ScaleStatefulSet(ctx, cli, namespace, name, replicas, waitForReady)
```
e-sumin, line 91: I'd suggest the following change, since it's a best practice to return an empty value when an error occurs. It clarifies the intent and also it maintains consistency with the standard library.

Also I think the result makes no sense in case of scaling failure, am I right ?

```suggestion
		count, err := kube.StatefulSetReplicas(ctx, cli, namespace, name)
		if err != nil {
			return nil, err
		}
		
		err = kube.ScaleStatefulSet(ctx, cli, namespace, name, replicas, waitForReady)
		if err != nil {
			return nil, err
		}
		
		return map[string]interface{}{
			outputArtifactOriginalReplicaCount: count,
		}, nil
```

The same pays for following changes too.

e-sumin (approved):
LGTM, except for the suggested change, but it's up to you whether to accept it or not.

pkg/function/scale_workload.go
viveksinghggits: I can do it, but in kanister generally we have been just returning the method if the return type matches. Here are some examples

https://github.com/kanisterio/kanister/blob/master/pkg/phase.go#L51
https://github.com/kanisterio/kanister/blob/master/pkg/phase.go#L97
https://github.com/kanisterio/kanister/blob/master/pkg/app/cockroachdb.go#L298
https://github.com/kanisterio/kanister/blob/master/pkg/blockstorage/vmware/vmware.go#L338

docs/tasks/scaleworkload.rst
PrasadG193: I don't think it would matter how it was used to work earlier once we override the functionality.

PR #2534, opened 2023-12-13
Optimizations for main build workflow

hairyhum:
## Change Overview

Trying to reduce time it takes to run the CI workflow. 
Currently packing and unpacking entire directory takes a long time. It's faster to just download the go dependencies.

Current changes:
- Don't pack entire dir on gomod job, only pass the go.mod and go.sum files as outputs
- Make `lint`, `test`, `build` and `docs` steps run in parallel
- Remove double invocation of `go test` in `test.sh`, this was left over from the time when test build was separate from test run, no longer required

These changes do not include any caching proposed in https://github.com/kanisterio/kanister/pull/2352 but already cut the pipeline time more then 2 times so it make sense to merge that first and then work on caching as a follow-up

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

Run the pipeline.

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,3 +1,4 @@
+name: Build and test
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -8,37 +9,52 @@ on:
...
+          echo FILE
+        } >> "$GITHUB_OUTPUT"
+    - id: gosum
+      run: |
+        {
+          echo 'gosum<<FILE'
+          cat go.sum
+          echo
+          echo FILE
+        } >> "$GITHUB_OUTPUT"
```
viveksinghggits, line 33: earlier we used to upload the code from one step and download it on other step? How does that work now? I see we are just outputting the go.* files and getting them from other step. How is code available in other steps?

Also, it looks like we are not running `go mod download` anymore, do we not need that?

Last question :smile: , why do we have to just copy the go.* between steps and not other code.

hairyhum: We do have the `checkout` step, which updates the code. `go.` files were synced before, but I'm not sure that's strictly required TBH. We could make it a check (that all deps are up to date) and don't move the `go.` files between jobs, which would make sense, but I just didn't want to introduce new behaviours in this PR. 
Doing `go mod download` and then packing and unpacking the files is way longer than to let each step (lint, build) to download deps themselves. Also if you run `go mod tidy` it does download all dependencies and there is no need to run `go mod download` again.

denisvmedia: Let me join the questions left by @viveksinghggits . I'd also put some textual notes to the workflow to explain why we are doing that / what we are trying to achieve.

build/test.sh
```diff
@@ -78,7 +78,6 @@ check_dependencies() {
 check_dependencies
 
 echo "Running tests:"
-go test -v -installsuffix "static" ${TARGETS}
```
denisvmedia, line 81: Why is this removed?

hairyhum: Originally we were builing tests with `go test -i` and then running it (next few lines), but then the flag was deprecated and removed from this line, but that meant that we run the tests twice, which doesn't make much sense.

denisvmedia (approved):
LGTM

viveksinghggits (approved):

PR #2539, opened 2023-12-14
Update golangci-lint to v1.55, fix lint errors

hairyhum:
## Change Overview

Currently golangci-lint is built with go version older than the go1.21 This can cause weird linter false-positives (e.g. https://github.com/kanisterio/kanister/actions/runs/7202258335/job/19620056020)

Updated version of golangci-lint in the build image Fixed linter warnings:
- deprecated rand.Seed
- non-sensical return in setupKopiaRepositoryServer

## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [x] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

PR #2543, opened 2023-12-14
Limit minimal time poll.Wait* functions can wait between attempts

hairyhum:
## Change Overview

Currently calculation of sleep time can make it negative, which makes `t.C` channel return right away and does not give ctx time to finish on deadline.
This means that even if context is expired, retries still happen until there are no more retries.

Currently it's causing test failures in `snapshot_test.go`

The fix is to cap minimal sleep time at 5ms to make sure that context can perform expiration.

NOTE: we could make the minimal sleep smaller, but it's wouldn't make sense getting less than OS scheduling frequency.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

Added new test to `poll_test.go`, but it's not always catching the race condition with context.

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


hairyhum:
Seems like this one needs an update linter from https://github.com/kanisterio/kanister/pull/2539

pkg/poll/poll.go
```diff
@@ -91,7 +91,11 @@ func WaitWithBackoffWithRetries(ctx context.Context, b backoff.Backoff, numRetri
 		sleep := b.Duration()
 		if deadline, ok := ctx.Deadline(); ok {
 			ctxSleep := time.Until(deadline)
-			sleep = minDuration(sleep, ctxSleep)
+			// We want to wait for smaller of backoff sleep and context sleep
```
julio-lopez, line 94: Question: IIUC from the PR description the `<-t.C` option is (often? always?) selected in the `select` block below when `sleep` is negative, is that the case?

If the deadline has passed, it seems that a sensible behavior would be to respect the context deadline by returning an error right away and thus not executing the operation.

It seems that with this change, `ctx.Done():` is going to be selected below when the context has expired, independent of how large `sleep` is, so in that case, why not simply make `sleep` positive (strictly greater than 0)? Would that be sufficient?

julio-lopez (changes requested):

pkg/poll/poll.go
hairyhum: Essentially this is a race condition. If `ctx` already expired but did not perfom the `cancel` yet, so there's nothing in the `ctx.Done()` channel. At the same time if we have negative timer, `t.C` will always have a value in the channel.
Just making sleep positive (in nanoseconds) doesn't really lets the context to finish in my tests.
We could just return a `context.DeadlineExceeded` error when `ctxSleep <= 0`, but then execution will continue in race with termination of the context. While technically correct (deadline did exceed), I don't know which consequences this might have.

hairyhum: In the other hand additional 10ms of wait doesn't really add that much considering default backoff starts with 100ms and goes up from there (so context deadline in 10s of ms doesn't make much sense)

pkg/poll/poll.go
```diff
@@ -91,7 +91,10 @@ func WaitWithBackoffWithRetries(ctx context.Context, b backoff.Backoff, numRetri
...
+			// but it has to be over os scheduling interval
+			// otherwise context doesn't have time to finish
```
julio-lopez, line 96:
```suggestion
			// but it has to be > 0 to give ctx.Done() a chance
			// to return below
```

julio-lopez (approved):
Please update the comment. See inline suggestion.

PR #2546, opened 2023-12-15
Change  go directive in go.mod to 1.21.5

viveksinghggits:
## Change Overview

This PR to changes the go directive of go.mod to 1.21.5, this PR along with https://github.com/kanisterio/kanister/pull/2545 would make sure that the kanister is built with go version 1.21.5.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

Build the binaries 

```
make release-snapshot
```
and make sure the go version in kanctl is 1.21.5.

```
go version -m dist/kanctl_linux_amd64_v1/kanctl
dist/kanctl_linux_amd64_v1/kanctl: go1.21.5 X:boringcrypto

```

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kanisterio/kanister
 
-go 1.21.3
+go 1.21.5
 
 replace (
 	github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.1


denisvmedia:
Note, this may break dependabot. This directive won't force the build to use 1.21.5, but will fail if it's not using it. I'm not actually sure we should do that now.

viveksinghggits:
I think it would be ok, even if we don't merge this. This is the one that should be merged though https://github.com/kanisterio/kanister/pull/2545.

viveksinghggits:
> Note, this may break dependabot. This directive won't force the build to use 1.21.5, but will fail if it's not using it. I'm not actually sure we should do that now.

I think it would be ok, even if we don't merge this. This is the one that should be merged though https://github.com/kanisterio/kanister/pull/2545.

PR #2547, opened 2023-12-15
Update ubi-minimal base image to ubi-minimal:9.3-1475

infraq:
Update ubi-minimal base image to ubi-minimal:9.3-1475


pavannd1:
@hairyhum FYI: we have a pipeline that keeps track of the base image and updates the version whenever a new one is available. It's good to consume these updates as soon as possible since they usually fix security vulnerabilities.

pavannd1 (approved):

PR #2567, opened 2023-12-20
Tests optimizations

hairyhum:
## Change Overview

Follow-up to https://github.com/kanisterio/kanister/pull/2534

- Run test suites in parallel
- Handle sigterm in some test containers
- Optimize sleeps and timeouts

**NOTE:** We currently use env variables to configure kanister and that prevents some tests from running in parallel. Currently using `testmutex` mutex to block them. 
This is not ideal solution, but other options such as regex matching or build tags are just worse. 
We could rework controller and API configuration so it doesn't rely on environment variables, which can potentially make it safer.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/build/test.sh
+++ b/build/test.sh
@@ -77,9 +77,12 @@ check_dependencies() {
 
 check_dependencies
 
+echo "Listing tests:"
+go test -v ${TARGETS} -check.list .
+echo
+
 echo "Running tests:"
-go test -v ${TARGETS} -list .
-go test -v -installsuffix "static" ${TARGETS} -check.v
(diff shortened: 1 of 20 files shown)


hairyhum:
This one is stale and will require some work to update all tests. Closing for now, can be a reference if we want to attempt to improve test performance in the future.

PR #2606, opened 2024-01-16
Add Kopia SDK storage wrappers for s3 provider (SDKWrappers-1)

kale-amruta:
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kopialib/storage/s3_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+var _ = Suite(&s3StorageTestSuite{})
+
+func (s *s3StorageTestSuite) TestSetOptions(c *C) {
+	for i, tc := range []struct {
+		name            string
+		options         map[string]string
+		expectedOptions s3.Options
+		expectedErr     string
+		errChecker      Checker
```
ankitjain235, line 36:
```suggestion
		errChecker      Checker
        desc            string
```
We could add a description here and use it below in line:
```c.Check(err.Error(), Equals, tc.expectedErr, Commentf("test number: %d, desc: %s", i, tc.desc))```

**Review by ankitjain235 — COMMENTED** [review]
Added some suggestions.

**Review by ankitjain235 — COMMENTED** [review]
Added some suggestions.

File: pkg/kopialib/storage/errors.go [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+var (
+	ErrMissingRequiredFieldMsg      = "missing required field: %s for storage provider %s"
+	ErrStorageOptionsCannotBeNilMsg = "storage options cannot be nil for the provider %s"
```
viveksinghggits, line 19:
```suggestion
	MissingRequiredFieldMsg      = "missing required field: %s for storage provider %s"
	StorageOptionsCannotBeNilMsg = "storage options cannot be nil for the provider %s"
```

Should we do this?
Because these are not actually errors but just messages?

pkg/kopialib/storage/s3.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/kanisterio/kanister/pkg/utils"
+	"github.com/kopia/kopia/repo/blob"
+	"github.com/kopia/kopia/repo/blob/s3"
+)
+
+var _ Storage = &s3Storage{}
+
+var requiredS3Arguments = []string{
+	kopialib.BucketKey,
+}
```
viveksinghggits, line 30:
```suggestion
var requiredArguments = []string{
	kopialib.BucketKey,
}
```

what do you think about this?
Because we are already in s3 file and s3 impl., so it would be better if we just have requiredArguments.

pkg/kopialib/storage/s3.go
```diff
@@ -0,0 +1,67 @@
...
+
+type s3Storage struct {
+	Options *s3.Options
+	Create  bool
```
viveksinghggits, line 34: what does `Create` decide/configure here, create if not exists?
Also, connect would connect to an S3 repo?

pkg/kopialib/storage/s3.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (s *s3Storage) Connect() (blob.Storage, error) {
+	return s3.New(context.Background(), s.Options, s.Create)
```
viveksinghggits, line 38: seems a bit misleading. Connect is the method name but we are calling `s3.New`, which looks like we are creating new instance of s3.

pkg/kopialib/storage/s3.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (s *s3Storage) WithOptions(opts s3.Options) {
+	s.Options = &opts
+}
+
+func (s *s3Storage) WithCreate(create bool) {
+	s.Create = create
+}
+
+func (s *s3Storage) SetOptions(ctx context.Context, options map[string]string) error {
```
viveksinghggits, line 49: do we really need `WithOptions` and `SetOptions` separately? it looks like both of them are just trying to set `s.Options`.

pkg/kopialib/storage/storage_interface.go
```diff
@@ -0,0 +1,58 @@
...
+	TypeS3        StorageType = "S3"
+	TypeAzure     StorageType = "Azure"
+	TypeFileStore StorageType = "FileStore"
+	TypeGCP       StorageType = "GCP"
+)
+
+type Storage interface {
+	Connect() (blob.Storage, error)
+	SetOptions(context.Context, map[string]string) error
+	WithCreate(bool)
```
viveksinghggits, line 36: I think these two methods should not be part of the interface. They can be part of the actual implementation though. 
Because they are more related to the impl. of the interface, whether we want to create the impl with options with created etc.

pkg/kopialib/storage/storage_interface.go
```diff
@@ -0,0 +1,58 @@
...
+func New(storageType StorageType) Storage {
+	switch storageType {
+	case TypeS3:
+		return &s3Storage{}
+	default:
+		return nil
+	}
+}
+
+func validateCommonStorageArgs(options map[string]string, storageType StorageType, requiredArgs []string) error {
```
viveksinghggits, line 48: if this validates common args for a specific implementation, we should have this as a method in the implementation itself. So that we can call it like

```
s3.validateCommonStorageArgs()
```
or something in the same lines.

viveksinghggits (changes requested):
I have added first round of comments, will have another look after these are addressed.

pkg/kopialib/storage/errors.go
r4rajat: Done !

pkg/kopialib/storage/s3.go
r4rajat: Yeah, Agreed ! Made the changes.

pkg/utils/utils.go
```diff
@@ -89,6 +89,19 @@ func GetIntOrDefault(value string, defaultValue int) (int, error) {
 	return v, nil
 }
 
+func GetBoolOrDefault(value string, defaultValue bool) (bool, error) {
```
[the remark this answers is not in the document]
r4rajat: Done !

pkg/kopialib/storage/s3.go
kale-amruta: Yes, SetOptions is a generic way to set configuration parameters for all the storage providers and its part of interface. WithOptions is not part of the interface and is a utililty which takes input specific to a provider. Its just a utility function that can be used apart from the generic method

pkg/kopialib/storage/storage_interface.go
kale-amruta: No it is generic for all, other storage providers would also have it

pkg/kopialib/storage/s3.go
kale-amruta: Create the S3 bucket if it does not exist

pkg/kopialib/storage/s3.go
kale-amruta: Removed WithOptions for now, will add later if required

pkg/kopialib/storage/storage_interface.go
kale-amruta: I think SetOptions should be part of interface because we cannot initialize a storage without options. I can remove Create method from implementation

pkg/kopialib/storage/storage_interface.go
viveksinghggits: > No it is generic for all, other storage providers would also have it

Sorry, I think I didn't understand properly what you meant by generic here. Do you mean this utility is common and would be used by all implementations?
If not, and other providers should also have it, in that case should we move this to provider's source file. That is s3.go?

pkg/kopialib/storage/s3.go
```diff
@@ -0,0 +1,76 @@
...
+
+var _ Storage = &s3Storage{}
+
+var requiredS3Arguments = []string{
+	kopialibutils.BucketKey,
+}
+
+type s3Storage struct {
+	options *s3.Options
+	create  bool
```
viveksinghggits, line 36: should this be named `createBucketIfNotExists` to make it more readable?

pkg/kopialib/storage/storage_interface.go
```diff
@@ -0,0 +1,58 @@
...
+type StorageType string
+
+const (
+	TypeS3        StorageType = "S3"
+	TypeAzure     StorageType = "Azure"
+	TypeFileStore StorageType = "FileStore"
+	TypeGCP       StorageType = "GCP"
+)
+
+type Storage interface {
```
viveksinghggits, line 34: Should this type be named something else, because if we access it from another package, we would have to use `storage.Storage`.

pkg/kopialib/storage/storage_interface.go
```diff
@@ -0,0 +1,58 @@
...
+	New() (blob.Storage, error)
+	SetOptions(context.Context, map[string]string) error
```
viveksinghggits, line 36: what are the other things that we would want to do here?  would we also want to create repo, connect to repo etc?
If yes, are we going to implement then in future?

pkg/kopialib/storage/storage_interface.go
kale-amruta: It is common and used by all the providers.

pkg/kopialib/storage/storage_interface.go
kale-amruta: Thats not part of storage, that will be part of repository. This interface is a wrapper to perform operations on storage providers

pkg/kopialib/storage/storage_interface.go
kale-amruta: I will rename it to provider

pkg/kopialib/storage/s3.go
kale-amruta: The create flag is not even used in the upstream kopia - https://github.com/kopia/kopia/blob/master/repo/blob/s3/s3_storage.go#L329. I am removing the create attribute from struct for now. Sorry for the confusion

pkg/kopialib/storage/s3.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/pkg/errors"
+)
+
+var _ Storage = &s3Storage{}
+
+var requiredS3Arguments = []string{
+	kopialibutils.BucketKey,
+}
+
+type s3Storage struct {
```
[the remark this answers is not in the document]
kale-amruta: I can rename it to Initialize

pkg/kopialib/storage/s3.go
kale-amruta: https://github.com/kanisterio/kanister/pull/2606#discussion_r1483910217

ankitjain235:
@kale-amruta Just checking, do we need this PR or should we close it?

PR #2641, opened 2024-01-31
fix: switch to bufio.Reader for KubeTask output parsing

### LINKED ISSUE #2612: [BUG] "Split lines failed: bufio.Scanner: token too long" for phase with a lot of output
**Describe the bug**
One of our phases increased its output by a huge amount of lines from one day to another.
From 450 lines (40Kb) to 2530 lines (169Kb). This caused Kanister to fail the actionset with the error
```
Split lines failed: bufio.Scanner: token too long
```
https://github.com/kanisterio/kanister/blame/master/pkg/output/stream.go#L50

**To Reproduce**
1. Create blueprint
2. Add phase that produces a lot of output: more than 64Kb
3. Run actionset

**Expected behavior**
Output is parsed.

**Screenshots**
n/a

**Environment**
Kubernetes Version/Provider: v1.27.7-gke.1121001
Storage Provider: n/a
Cluster Size (#nodes): n/a
Data Size: n/a
Kanister: 0.99.0

**Additional context**
The default buffer of `bufio` is 64Kb: https://pkg.go.dev/bufio#pkg-constants
That explains, why the error occurs when crossing this boundary.
```
const (
	// MaxScanTokenSize is the maximum size used to buffer a token
	// unless the user provides an explicit buffer with Scanner.Buffer.
	// The actual maximum token size may be smaller as the buffer
	// may need to include, for instance, a newline.
	MaxScanTokenSize = 64 * 1024
)
```

I'm not quite sure how to handle this. Just increasing the buffer blindly to some higher value does not seem like a good solution.

Thread in Slack: https://kanisterio.slack.com/archives/C85C8V22J/p1705490134135299

hairyhum:
## Change Overview

Use `bufio.Reader` and `ReadLine` instead of `bufio.Scanner`
`bufio.Scanner` has buffer limit and can fail the function if output is too big

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2612

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


viveksinghggits (approved):

e-sumin:
@hairyhum The `bufio.Scanner` had a problem - its buffer was limited and could be overwhelmed by some output. Your solution does fix this problem, but it introduces another issue - there is no protection against complete memory exhaustion. This is what I meant during the Community meeting when I said that I couldn't see a simple solution.

hairyhum:
@e-sumin I agree that there's a risk of higher memory usage

> 1.2 - If a line does not fit into the buffer, log it in chunks of 65kb
> 2 - Perform output parsing as we are currently doing (expecting the lines with ###Phase-output### to be small and fit into 65kb)

So the assumption there would be that the output line bigger than 65kb would not have `###Phase-output###` in there?
Currently (and it can be not a good thing) the phase output assumptions are:
- Line shorter than 65kb
- Line contains `###Phase-output###`
- Everything  after `###Phase-output###` until the end of the line is the output

If line is > 65kb right now - there's an error.

As I understand, you propose to keep the assumptions with a difference that if the line is >65kb - it does not cause an error and just getting logged.

The issue I see with it is that we introduce a hidden assumption which may silently ignore some lines with `###Phase-output###`

Alternative things we can do:
- Bump supported line length. Then the question is now much considering potential memory issues
- Require `###Phase-output###` to be in the beginning of the line (e.g. first 100 bytes), this is sort of a breaking change because we currently don't require it. And then we know that the line is an output and we may or may not restrict the size of output value. It would make sense to do that.
- Scan long lines until we reach `###Phase-output###` and then scan until the end of the line. Again we could add a limit for the output itself. This preserves support for current format while removing the error.

WDYT?

hairyhum:
New change follows the `Scan long lines until we reach ###Phase-output### and then scan until the end of the line. ` approach.
This allows to support existing behaviour and long lines both for non-output and output.
Bytes are read in chunks which should have lower memory usage.

**TODO:** add unit tests.

e-sumin:
@hairyhum I've added some draft for unit test, which is able to generate random input.

pkg/output/stream_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (s *OutputTestSuite) TestHugeStreamsWithoutPhaseOutput(c *C) {
+	done := make(chan struct{})
+	defer func() { close(done) }()
+
+	// e-sumin: Here I'm generating test case, when we have just random string around 10000 runes
+	// I expect that it will be logged as one line as it was before
+	// But in fact it is logged by chunks of ±4kb
+	// When we will fix the code behavior, numOfLines has to be set to 10, and avgPrefix len has to be set to 500000
```
e-sumin, line 150: @hairyhum could you please check this case ? I believe we should try to keep old behaviour as soon as it possible.

hairyhum: Fixed with replacing `NewReader` with `NewReaderSize`

pkg/output/output_test.go
```diff
@@ -45,13 +48,27 @@ func (s *OutputSuite) TestValidateKey(c *C) {
 	}
 }
 
+// FIXME: replace this with TestLogAndParse
```
e-sumin, line 51: Do you want to fix this in this PR ?

pkg/output/stream.go
```diff
@@ -35,31 +42,166 @@ func splitLines(ctx context.Context, r io.ReadCloser, f func(context.Context, st
...
+		outputContent := state.outputBuf
+		outputContent = append(outputContent, line...)
+
+		if err := f(ctx, outputContent); err != nil {
+			return state, err
+		}
+
+		// Transition out of readingOutput state
+		return InitState(), nil
 	}
```
e-sumin, line 119:
```suggestion
	}
	
	// Reached the end of the line while reading phase output
	if err := f(ctx, append(state.outputBuf, line...)); err != nil {
		return state, err
	}

	// Transition out of readingOutput state
	return InitState(), nil
```

pkg/output/stream.go
```diff
@@ -35,31 +42,166 @@ func splitLines(ctx context.Context, r io.ReadCloser, f func(context.Context, st
...
-	return errors.Wrap(s.Err(), "Split lines failed")
 }
 
+func handleSeparatorSuffix(state scanState, line []byte, isPrefix bool, ctx context.Context, f func(context.Context, []byte) error) (scanState, error) {
+	if bytes.Index(line, state.separatorSuffix) == 0 {
+		return captureOutputContent(line, isPrefix, len(state.separatorSuffix), ctx, f)
+	} else {
+		// Read log like normal
+		return handleLog(line, isPrefix, ctx, f)
+	}
```
e-sumin, line 128:
```suggestion
	}

	// Read log like normal
	return handleLog(line, isPrefix, ctx, f)
```

pkg/output/stream.go
```diff
@@ -35,31 +42,166 @@ func splitLines(ctx context.Context, r io.ReadCloser, f func(context.Context, st
...
+		return InitState(), nil
+	} else {
+		// Log everything before separator as plain output
+		prefix := line[0:indexOfPOString]
+		if len(prefix) > 0 {
+			logOutput(ctx, prefix)
+		}
+
+		return captureOutputContent(line, isPrefix, indexOfPOString+len(PhaseOpString), ctx, f)
+	}
```
e-sumin, line 153:
```suggestion
	}

	// Log everything before separator as plain output
	prefix := line[0:indexOfPOString]
	if len(prefix) > 0 {
		logOutput(ctx, prefix)
	}

	return captureOutputContent(line, isPrefix, indexOfPOString+len(PhaseOpString), ctx, f)
```

pkg/output/stream.go
```diff
@@ -35,31 +42,166 @@ func splitLines(ctx context.Context, r io.ReadCloser, f func(context.Context, st
...
+	outputContent := line[startIndex:]
+	if !isPrefix {
+		if err := f(ctx, outputContent); err != nil {
+			return InitState(), err
+		}
+		return InitState(), nil
+	} else {
+		return ReadPhaseOutputState(append([]byte(nil), outputContent...)), nil
+	}
+}
```
e-sumin, line 166:
```suggestion
func captureOutputContent(
	line []byte,
	isPrefix bool,
	startIndex int,
	ctx context.Context,
	f func(context.Context, []byte) error,
) (scanState, error) {
	outputContent := line[startIndex:]
	if !isPrefix {
		if err := f(ctx, outputContent); err != nil {
			return InitState(), err
		}
		return InitState(), nil
	}

	return ReadPhaseOutputState(append([]byte(nil), outputContent...)), nil
}
```

pkg/output/stream.go
```diff
@@ -35,31 +42,166 @@ func splitLines(ctx context.Context, r io.ReadCloser, f func(context.Context, st
...
+func handleSeparatorSuffix(state scanState, line []byte, isPrefix bool, ctx context.Context, f func(context.Context, []byte) error) (scanState, error) {
+	if bytes.Index(line, state.separatorSuffix) == 0 {
+		return captureOutputContent(line, isPrefix, len(state.separatorSuffix), ctx, f)
+	} else {
+		// Read log like normal
+		return handleLog(line, isPrefix, ctx, f)
+	}
+}
+
+func handleLog(line []byte, isPrefix bool, ctx context.Context, f func(context.Context, []byte) error) (scanState, error) {
```
e-sumin, line 131:
```suggestion
func handleLog(
	line []byte,
	isPrefix bool,
	ctx context.Context,
	f func(context.Context, []byte) error,
) (scanState, error) {
```

pkg/output/stream.go
```diff
@@ -35,31 +42,166 @@ func splitLines(ctx context.Context, r io.ReadCloser, f func(context.Context, st
...
+			return state, err
+		}
+
+		// Transition out of readingOutput state
+		return InitState(), nil
 	}
-	return errors.Wrap(s.Err(), "Split lines failed")
 }
 
+func handleSeparatorSuffix(state scanState, line []byte, isPrefix bool, ctx context.Context, f func(context.Context, []byte) error) (scanState, error) {
```
e-sumin, line 122:
```suggestion
func handleSeparatorSuffix(
	state scanState,
	line []byte,
	isPrefix bool,
	ctx context.Context,
	f func(context.Context, []byte) error,
) (scanState, error) {
```

pkg/output/output_test.go
hairyhum: I think this is addressed in `stream_test.go`

e-sumin (approved):
LGTM

PR #2645, opened 2024-02-01
docs(readme): describe commit conventions

hairyhum:
## Change Overview

Add commit conventions discussed in https://github.com/kanisterio/kanister/discussions/2582

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,21 +84,26 @@ If your changes involve the Kanister API types, generate the API documentation u
 ### Commit messages
 
 The basic idea is that we ask all contributors to practice
-[good git commit hygiene](https://www.futurelearn.com/info/blog/telling-stories-with-your-git-history)
+[good git commit hygiene](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/)
 to make reviews and retrospection easy. Use your git commits to provide context
 for the reviewers, and the folks who will be reading the codebase in the months
 and years to come.
 
-Finalized commit messages should look similar to the following format:
+We're trying to keep all commits in `master` to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format.
+See [conventions](#commit-conventions) for more info on types and scopes.
+We are using squash and merge approach to PRs which means that commit descriptions are generated from PR titles.
 
-```text
-Short one line title
-
-An explanation of the problem, providing context, and why the change is being
-made.
-```
+It's recommended to use conventional commits when strarting a PR, but follow-up commits in the PR don't have to follow the convention.
 
 ### Submitting Pull Requests
+
+**PR titles should be in following format:**
+
+```text
+<type>[optional scope]: <description>
+```
+
+See [conventions](#commit-conventions) for more info on types and scopes.
 
 When submitting a pull request, it's important that you communicate your intent,
 by clearly:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -89,17 +89,76 @@ to make reviews and retrospection easy. Use your git commits to provide context
...
-Finalized commit messages should look similar to the following format:
+Kanister recommends using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format:
 
 ```text
-Short one line title
+<type>[optional scope]: <description>
 
-An explanation of the problem, providing context, and why the change is being
-made.
+[optional body]
```
viveksinghggits, line 97: Hi @hairyhum ,
Sorry, if I am mistaken. But I thought we discussed that this format should be followed in PR titles and desc.
What I had in mind for commits was a rule set like this


Commits should be of format

```
Title
Description
```

- It's totally fine if you are able to convey the change in title itself, you don't necessarily have to provide desc.
- Commit titles should not start with lower case. So `add go mod check while init` -> `Add go mod check while init`, same applies to PR titles as well
- Title should imperative so use `Add go lint` instead of `Added go lint`, same applies to PR titles as well
- Do not end title with period
- Desc (if present) should specify why did you make that change
- Don't assume reviewer understand the problem

This is very rough, but I think for commits we should have this kind of rules and for PR titles we can have the type and scope etc.

hairyhum: So your proposal is to have PR title/description convention **instead of** commit convention?

hairyhum:
@viveksinghggits i've tried to reword the requirements and give more context on the flow. Does that make sense?

CONTRIBUTING.md
```diff
@@ -89,16 +89,21 @@ to make reviews and retrospection easy. Use your git commits to provide context
 for the reviewers, and the folks who will be reading the codebase in the months
```
viveksinghggits, line 89: this futurelearn link is not working anymore we should remove that.

CONTRIBUTING.md
```diff
@@ -89,16 +89,21 @@ to make reviews and retrospection easy. Use your git commits to provide context
...
 and years to come.
 
-Finalized commit messages should look similar to the following format:
+We're trying to keep all commits in `master` to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format.
+See [conventions](#commit-conventions) for more info on types and scopes.
+We are using squash and merge approach to PRs which means that commit descriptions are generated from PR descriptions.
```
viveksinghggits, line 94: I am not really sure if the desc for the commit, that is created in master when a PR is merged is, is created from PR desc. Instead I have seen that commit desc just has the other commits/their desc from the PR.
I was looking into this PR for example https://github.com/kanisterio/kanister/pull/2664 below is how this PR is merged
image

As we can see the commit message is PR title but the commit desc is the commits of the PR.

CONTRIBUTING.md
```diff
@@ -89,16 +89,21 @@ to make reviews and retrospection easy. Use your git commits to provide context
...
 
-```text
-Short one line title
+It's recommended to use conventional commits when strarting a PR, but follow-up commits in the PR don't have to follow the convention.
```
viveksinghggits, line 96: I think we should also mention the points under `5 Steps to Write Better Commit Messages` from this blog post https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/ somewhere here.
I really like the points where imperative mood is recommended and others.

viveksinghggits (approved):
LGTM

CONTRIBUTING.md
hairyhum: I meant pr title. Updated.

julio-lopez (commented):
🎉 🎊 🚀 🥳

PR #2654, opened 2024-02-06
Implement Kopia common args and opts

plar:
## Change Overview

This PR is the 2nd in the series of PRs that will add a new way to build Kopia CLI commands.   

PR Train:

1. [PR-2653](https://github.com/kanisterio/kanister/pull/2653) Introducing a new approach to building Kopia CLI commands using safecli v0.0.6
1. **[PR-2654](https://github.com/kanisterio/kanister/pull/2654) Implement Kopia common args and opts**
1. [PR-2655](https://github.com/kanisterio/kanister/pull/2655) Implement Kopia storage helpers
1. [PR-2656](https://github.com/kanisterio/kanister/pull/2656) Add Kopia Filesystem storage opts
1. [PR-2657](https://github.com/kanisterio/kanister/pull/2657) Add Kopia GCS storage opts
1. [PR-2658](https://github.com/kanisterio/kanister/pull/2658) Add Kopia Azure storage opts
1. [PR-2659](https://github.com/kanisterio/kanister/pull/2659) Add Kopia S3 and S3 compliant storage opts
1. [PR-2661](https://github.com/kanisterio/kanister/pull/2661) Add Kopia repository create command
1. [PR-2662](https://github.com/kanisterio/kanister/pull/2662) Add Kopia repository connect command
1. [PR-2663](https://github.com/kanisterio/kanister/pull/2663) Add Kopia repository connect server command
1. [PR-2664](https://github.com/kanisterio/kanister/pull/2664) Add Kopia repository set-parameters command
1. [PR-2665](https://github.com/kanisterio/kanister/pull/2665) Add Kopia repository status command
This PR introduces common argument structures and flags for Kopia CLI, along with extended error definitions:

* `pkg/kopia/cli/args/Common` and `pkg/kopia/cli/args/CacheA` structs encapsulate common command-line arguments and cache-related arguments for the Kopia CLI.
* `pkg/kopia/cli/internal/opts` package presents common and cache-related options for the Kopia CLI.
* `pkg/kopia/cli/internal/args` package presents common arguments for the Kopia CLI.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kopia/cli/internal/flag/common/common_flags.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func (f checkForUpdates) Flag() string {
+	if f.CheckForUpdates {
+		return "--check-for-updates"
+	}
+	return "--no-check-for-updates"
+}
+
+func (f checkForUpdates) Apply(cli safecli.CommandAppender) error {
```
hairyhum, line 91: Same as for `checkForUpdates`: is there a reason this isn't just a bool flag?

pkg/kopia/cli/internal/flag/common/common_flags.go
```diff
@@ -0,0 +1,205 @@
...
+		// ContentCacheSizeMB(f.ContentCacheSizeMB),
+		// MetadataCacheSizeMB(f.MetadataCacheSizeMB),
+	)
+}
+
+// Cache creates a new cache flag.
+// If no arguments are provided, the default cache flags are used.
+// If one argument is provided, the cache flags are used.
+// If more than one argument is provided, ErrInvalidCacheArgs is returned.
+func Cache(args ...cli.CacheArgs) flag.Applier {
```
hairyhum, line 175: Same as for `Common`.

hairyhum (commented):
Just some comments about code size. It feels like we can shed some code in some places here.

pkg/kopia/cli/internal/flag/common/common_flags.go
```diff
@@ -0,0 +1,205 @@
...
+var (
+	All           = flag.NewBoolFlag("--all", true)
+	Delta         = flag.NewBoolFlag("--delta", true)
+	ShowIdentical = flag.NewBoolFlag("--show-identical", true)
+	NoGRPC        = flag.NewBoolFlag("--no-grpc", true)
+)
+
+// predefined flags
+var (
+	CheckForUpdates   = checkForUpdates{CheckForUpdates: true}
```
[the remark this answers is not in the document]
ankitjain235: Right, all the above flags above(`--all`, `--grpc`) are supported in kopia in this format only.
All(we can double check) such flags are of this format only: `--[no-]flag`.
If for some flag default is true and we need to disable it, we are passing `--no-grpc`.
Similarly, if for a flag default is false and we need to enable is, we pass `--all`.
The handling for these could be common?

ankitjain235 (commented):
Added one suggestion, looks good otherwise.

pkg/kopia/cli/internal/opts/opts.go
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+package opts
+
+import "github.com/kanisterio/safecli/command"
+
+var (
+	All           = command.NewOption("--all", true)
+	Delta         = command.NewOption("--delta", true)
+	ShowIdentical = command.NewOption("--show-identical", true)
```
ankitjain235, line 22: These can also be moved to functions below right? They are also of the same format? WDYT?

ankitjain235 (commented):
LGTM besides a minor comment.

pkg/kopia/cli/internal/opts/cache_opts.go
```diff
@@ -0,0 +1,68 @@
...
+	defaultCacheDirectory = "/tmp/kopia-cache"
+)
+
+// CacheDirectory creates a new cache directory option with a given directory.
+// If the directory is empty, the default cache directory is used.
+func CacheDirectory(dir string) command.Applier {
+	if dir == "" {
+		dir = defaultCacheDirectory
+	}
+	return command.NewOptionWithArgument("--cache-directory", dir)
```
hairyhum, line 34: `NewOptionWithArgument` would return noop if the value of the argument is `""`.
I wonder if we have any places in the CLI where an empty argument makes logical sense to supply?
Maybe not in this particular case, but there could be some argument somewhere.

pkg/kopia/cli/internal/opts/opts.go
plar: The reason I did it is because we are using the `--all` flag exclusively in the `kopia snapshot list...` command in our current implementation, and it is non-conditional. However, I agree with you and think that we need to convert them into functions for consistency. On the command side, we will simply write `opts.All(true)`, etc.

pkg/kopia/cli/internal/opts/cache_opts.go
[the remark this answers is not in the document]
plar: I prefer to leave it as a constant, since using the `defaultCacheDirectory` here improves code readability and maintainability by centralizing the default cache directory path. It helps to make any future updates simpler and reducing the risk of typos.

plar: I haven't seen such a case. An option without an argument is just an option, and we should use `NewOption`.

hairyhum: I was talking about deliberately passing an empty string. Since in shell missing argument and argument with an empty string as a value is not the same thing, but `NewOptionWithArgument` has an implicit assumption that it is.
Maybe we could make it explicit somehow?

plar: If we remove the specified code from [https://github.com/kanisterio/safecli/command/option_arg.go](https://github.com/kanisterio/safecli/blob/38b6544f3590512f0da53778d152d7bbee8b7c4e/command/option_arg.go#L44-L46) 

```go
...
	} else if arg == "" {
		return noopArgument{}
	}
...
```
then we will generate an option as `--option=` without any value on the right side and this is usually considered a bad argument. We can simulate this approach by just using `NewOption("--option=")` for the empty argument.

I just checked `argparse` (python argument parser) and this function allows you to do it quite well. The decision about whether this is an error should be made on the user side.

```python
# args.py
import argparse

parser = argparse.ArgumentParser(description='cli tool')
parser.add_argument('--option', type=str, default="", nargs='?', help='optional arg with default empty string')

args = parser.parse_args()
print(f"Option value: '{args.option}'")
```

```bash
$ python args.py --option=
Option value: ''
```

You're probably right, we shouldn't prohibit the user from stepping on the rake if he wants to. :)

Let's remove [noopArgument code](https://github.com/kanisterio/safecli/blob/38b6544f3590512f0da53778d152d7bbee8b7c4e/command/option_arg.go#L44-L46)? WDYT?

hairyhum: The problem I see there is that it will add `--option=` for non-initialized strings, because for golang an empty string and non-initialized string is the same thing. This is essentially an inconsistency between golang and shell arguments. So if we keep it implicit we still have a problem.
We can make it explicit by maybe providing an option to `NewOptionWithArgument` or making something like `NewOptionWithNonEmptyArgument` to make it clear if you use it you can't pass `""` as an argument.

plar: If we remove the [noopArgument code](https://github.com/kanisterio/safecli/blob/38b6544f3590512f0da53778d152d7bbee8b7c4e/command/option_arg.go#L44-L46), we make it **explicit**. The user will receive exactly what they pass in, and we leave the decision about how to handle the empty argument up to the user, similar to the `CacheDirectory` option above.

Adding `isArgRequired` to `NewOptionWithArgument` or creating a new `NewOptionWithNonEmptyArgument` will  pollute `safecli` (we would have to add a similar method for redacted NewXXX functions) IMHO. 

It is also unclear what to do if `isArgRequired` is `true` and the `arg` is empty. Should we ignore the argument or return an error? There's too much logic involved. :thinking:

hairyhum: Ok. Fair enough. This means if we want to treat non-initialized strings as no argument we need to move noop logic a level higher.

plar: Right, and we already have `NewNoopArgument`, which can help build this logic.

```go
// OptWithOptionalArgAndDefault("") => --opt=default
// OptWithOptionalArgAndDefault("custom") => --opt=custom
func OptWithOptionalArgAndDefault(arg string) command.Applier {
	if arg == "" {
		arg = "default"
	}
	return command.NewOptionWithArgument("--opt", dir)
}

// OptWithOptionalArgAndNoop("") => nothing will be generated
// OptWithOptionalArgAndNoop("custom") => --opt=custom
func OptWithOptionalArgAndNoop(opt string) command.Applier {
	if opt == "" {
		return command.NewNoopArgument()
	}
	return command.NewOptionWithArgument("--opt", opt)
}

// OptWithRequiredArgAndError("") => error
// OptWithRequiredArgAndError("custom") => --opt=custom
func OptWithRequiredArgAndError(opt string) command.Applier {
	if opt == "" {
		return command.NewErrorArgument(fmt.Errorf("opt is required"))
	}
	return command.NewOptionWithArgument("--opt", opt)
}
```

hairyhum (approved):

pkg/kopia/cli/internal/args/args_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+package args_test
+
+import (
+	"testing"
+
+	"github.com/kanisterio/kanister/pkg/kopia/cli"
+	"github.com/kanisterio/kanister/pkg/kopia/cli/internal/args"
+	"github.com/kanisterio/safecli/test"
+	"gopkg.in/check.v1"
```
PrasadG193, line 23:
```suggestion
	"testing"

	"gopkg.in/check.v1"
	"github.com/kanisterio/safecli/test"

	"github.com/kanisterio/kanister/pkg/kopia/cli"
	"github.com/kanisterio/kanister/pkg/kopia/cli/internal/args"
```
Please place external packages to a separate group

pkg/kopia/cli/internal/args/args_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Name:        "Invalid ID",
+		Argument:    args.ID(""),
+		ExpectedErr: cli.ErrInvalidID,
+	},
+	{
+		Name:        "ID",
+		Argument:    args.ID("id12345"),
+		ExpectedCLI: []string{"cmd", "id12345"},
+	},
+}})
```
PrasadG193, line 39: Not test to run defined? Am I missing something?

pkg/kopia/cli/internal/opts/cache_opts_test.go
```diff
@@ -0,0 +1,54 @@
...
+package opts_test
+
+import (
+	"testing"
+
+	"github.com/kanisterio/kanister/pkg/kopia/cli/args"
+	"github.com/kanisterio/kanister/pkg/kopia/cli/internal/opts"
+	"github.com/kanisterio/safecli/command"
+	"github.com/kanisterio/safecli/test"
+	"gopkg.in/check.v1"
```
PrasadG193, line 24: These packages are external packages to Kanister, so should be kept in separate group.

pkg/kopia/cli/internal/opts/cache_opts.go
```diff
@@ -0,0 +1,68 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package opts
+
+import (
+	"strconv"
+
+	"github.com/kanisterio/kanister/pkg/kopia/cli/args"
+	"github.com/kanisterio/safecli/command"
```
PrasadG193, line 21: `safecli` package is extenal package to the Kanister, so should be kept in the separate group

pkg/kopia/cli/internal/opts/common_opts.go
```diff
@@ -0,0 +1,70 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package opts
+
+import (
+	"github.com/kanisterio/kanister/pkg/kopia/cli/args"
+	"github.com/kanisterio/safecli/command"
```
PrasadG193, line 19: same here

pkg/kopia/cli/internal/opts/common_opts_test.go
```diff
@@ -0,0 +1,60 @@
...
+package opts_test
+
+import (
+	"testing"
+
+	"github.com/kanisterio/kanister/pkg/kopia/cli/args"
+	"github.com/kanisterio/kanister/pkg/kopia/cli/internal/opts"
+	"github.com/kanisterio/safecli/command"
+	"github.com/kanisterio/safecli/test"
+	"gopkg.in/check.v1"
```
PrasadG193, line 24: ditto

pkg/kopia/cli/internal/opts/opts_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+package opts_test
+
+import (
+	"testing"
+
+	"github.com/kanisterio/kanister/pkg/kopia/cli/internal/opts"
+	"github.com/kanisterio/safecli/command"
+	"github.com/kanisterio/safecli/test"
+	"gopkg.in/check.v1"
```
PrasadG193, line 23: same here

PrasadG193 (approved):
Needs minor formating fixes, rest looks good to me

pkg/kopia/cli/internal/args/args_test.go
plar: What do you mean? We have 2 tests here:

```
func TestArgs(t *testing.T) { check.TestingT(t) }
// Go test/bridge func for `go test` and delegate the actual testing work to the `check` framework.

// register `test.ArgumentSuite` with the `check` framework.
var _ = check.Suite(&test.ArgumentSuite{Cmd: "cmd", Arguments: []test.ArgumentTest{
	{
		Name:        "Invalid ID",
		Argument:    args.ID(""),
		ExpectedErr: cli.ErrInvalidID,
	},
	{
		Name:        "ID",
		Argument:    args.ID("id12345"),
		ExpectedCLI: []string{"cmd", "id12345"},
	},
}})
```

We're using [`github.com/kanisterio/safecli/test.ArgumentSuite`](https://github.com/kanisterio/safecli/blob/main/test/arg_suite.go#L28) to run table tests.

pkg/kopia/cli/args/cache_args.go
```diff
@@ -0,0 +1,26 @@
...
+
+// Cache provides the cache arguments for Kopia CLI.
+type Cache struct {
+	CacheDirectory           string // the directory where cache is stored. Default is "/tmp/kopia-cache".
+	ContentCacheSizeLimitMB  int    // the maximum size of the content cache in MB.
+	MetadataCacheSizeLimitMB int    // the maximum size of the metadata cache in MB.
+
+	// unused?
+	ContentCacheSizeMB  int // the size of the content cache in MB.
+	MetadataCacheSizeMB int // the size of the metadata cache in MB.
```
julio-lopez, line 25: @plar If these are not referenced anywhere, can you remove them please?

PR #2705, opened 2024-02-28
Add `provisioned-by` annotations to PVs created by `CreateVolumeFromSnapshot` kanister function

viveksinghggits:
## Change Overview

If PVs and PVCs are created by kanister using provider based provisioners and proper annotations are not set, when we delete the PVC, respective PV gets to `Failed` state and is not cleaned up. Because of that the volume in storage provider is not deleted. This happens because of the migration of provider based provisioner's migration to CSI based provisioners. And if proper annotation is not added provider based PVs are not deleted because CSI would not know what to do.

This PR makes sure that when kanister creates the PV in `CreateVolumeFromSnapshot` kanister function, we add proper annotation so that when PVCs are deleted, PVs are released properly and deleted.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

```
/kanister/pkg/function (provby-pv-annotations*) » go test -check.f "CreateVolumeFromSnapshotTestSuite.TestAddPVProvisionedByAnnotation"
OK: 1 passed
PASS
ok  	github.com/kanisterio/kanister/pkg/function	0.671s
```

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/function/create_volume_from_snapshot_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"gopkg.in/check.v1"
```
pavannd1, line 22:
```suggestion
```
Already imported in the next line

pkg/function/create_volume_from_snapshot.go
```diff
@@ -115,7 +116,9 @@ func createVolumeFromSnapshot(ctx context.Context, cli kubernetes.Interface, nam
 		if err != nil {
 			return nil, errors.Wrapf(err, "Unable to create PVC for volume %v", *vol)
 		}
-		pv, err := kubevolume.CreatePV(ctx, cli, vol, vol.Type, annotations, nil, nil)
+
+		pvAnnotations := addPVProvisionedByAnnotation(nil, provider)
```
pavannd1, line 120: Do we need the `annotations` from above to be passed in here? Does `kubevolume.CreatePVC` add values to the annotations map?

pkg/function/create_volume_from_snapshot_test.go
```diff
@@ -133,3 +137,51 @@ func (s *CreateVolumeFromSnapshotTestSuite) TestCreateVolumeFromSnapshot(c *C) {
...
+			},
+			expectedAnnotations: map[string]string{
+				"keyone":                         "valueone",
+				"keytwo":                         "valuetwo",
+				consts.PVProvisionedByAnnotation: consts.AWSEBSProvisionerInTree,
+			},
+		},
+	} {
+		op := addPVProvisionedByAnnotation(tc.annotations, tc.st)
+		c.Assert(op, check.DeepEquals, tc.expectedAnnotations)
```
pavannd1, line 185:
```suggestion
		c.Assert(op, DeepEquals, tc.expectedAnnotations)
```
`gocheck` is imported as `.` above

pkg/function/create_volume_from_snapshot.go
viveksinghggits: `kubevolume.CreatePVC` doesn't add values to the annotations. But I think its better idea to have separate annotation values for PV and PVC.

viveksinghggits: and the reason is, as we know these two resources can  have different set of labels.

pkg/function/create_volume_from_snapshot_test.go
viveksinghggits: ohh.
go_lint passed somehow. I will fix this.

pkg/function/create_volume_from_snapshot.go
```diff
@@ -125,6 +128,22 @@ func createVolumeFromSnapshot(ctx context.Context, cli kubernetes.Interface, nam
 	return providerList, nil
 }
 
+func addPVProvisionedByAnnotation(annotations map[string]string, provider blockstorage.Provider) map[string]string {
+	if annotations == nil {
```
ewhamilton, line 132: Would we rather defer this check until after storageType check so that we don't change a nil annotations into an empty map for other providers?

Just asking, though it might not matter in this code.

ewhamilton:
LGTM. I'd approve but I don't recall whether that triggers merge on kanister or not. Your option on my feedback. Otherwise I approve and ask Pavan for +1. (I'm AFK all weekend).

pkg/function/create_volume_from_snapshot.go
```diff
@@ -125,6 +128,22 @@ func createVolumeFromSnapshot(ctx context.Context, cli kubernetes.Interface, nam
 	return providerList, nil
 }
 
+func addPVProvisionedByAnnotation(annotations map[string]string, provider blockstorage.Provider) map[string]string {
```
PrasadG193, line 131: Do we really need `annotations` arg in the function? The only reference I found passes `nil` value to this argument. Can we get rid of it?

viveksinghggits: I added annotations just for the cases where we have a PV's annotations and want to set this annotation. Not being used right now, but might be used in future.

pkg/function/create_volume_from_snapshot.go
viveksinghggits: yeah i also think if we are here it would mean that we are going to add those annotations, that is the reason I didn't have that check before. can add it, if we think thats better.

pavannd1 (approved):

pavannd1:
@viveksinghggits are we waiting on any responses for this or is it ready to merge?

viveksinghggits:
> @viveksinghggits are we waiting on any responses for this or is it ready to merge?

this is not ready to merge. there were some other bugs that I figured out while working on this and those are being merged as part of 
https://github.com/kanisterio/kanister/pull/2732
https://github.com/kanisterio/kanister/pull/2731

once those are merged. We will be able to merge this.

viveksinghggits:
As soon as they are are merged I will add kueue to this.

PR #2731, opened 2024-03-08
Fix a bug in `zoneToRegion` which didn't work if multiple zones where passed separated by `__`

viveksinghggits:
## Change Overview

There was a bug in `zoneToRegion` function because of which the PV resources were not being created with correct `nodeAffinity` set. `zoneToRegion` tried to get regions from  passed zone just by removing `-[onechar]` from the zone. For example if we passed the zone `us-west1-b` this function would return `us-west1` by removing `-b`. This would work properly in the cases where just one zone is given to the function.

But if more than one zones are given to the function (separated by `__`), this function didn't work properly. Because for the input `us-west1-a__us-west2-a` it returned `us-west1-a__us-west2` (using existing logic), which is incorrect. As we can see the `-a` was not reomved from first zone. This commit doesn't introduce the separator `__`, it was already part of code and zones were being passed to this funciton separated by `__`.

This commit fixes above problem by making sure that we split the zones and then remove `-char` from the zone and add the regions together again using `__` sep.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

```
go test -check.f "TestVolSuite.TestZoneToRegion" -count 10
OK: 1 passed
OK: 1 passed
OK: 1 passed
OK: 1 passed
OK: 1 passed
OK: 1 passed
OK: 1 passed
OK: 1 passed
OK: 1 passed
OK: 1 passed
PASS
ok  	github.com/kanisterio/kanister/pkg/kube/volume	0.520s
```

### DIFF
--- a/pkg/blockstorage/blockstorage_test.go
+++ b/pkg/blockstorage/blockstorage_test.go
@@ -28,6 +28,7 @@ import (
 	ktags "github.com/kanisterio/kanister/pkg/blockstorage/tags"
 	envconfig "github.com/kanisterio/kanister/pkg/config"
 	"github.com/kanisterio/kanister/pkg/field"
+	"github.com/kanisterio/kanister/pkg/kube/volume"
 	"github.com/kanisterio/kanister/pkg/log"
 )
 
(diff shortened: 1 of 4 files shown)


hairyhum (approved):

e-sumin (commented):
LGTM

PR #2733, opened 2024-03-08
refactor: Satisfy linter in `zonesToRegion`

e-sumin:
## Change Overview

There was annotation which disables `gocritic`. This linter could be enabled now, since we do have test which will detect panic.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/volume/volume.go
+++ b/pkg/kube/volume/volume.go
@@ -348,16 +348,16 @@ func zoneToRegion(zone string) string {
 }
 
 func zonesToRegions(zone string) []string {
-	reg := map[string]struct{}{}
+	reg := map[string]bool{}
+	var regions []string
 	r := regexp.MustCompile("-?[a-z]$")
 	for _, z := range strings.Split(zone, RegionZoneSeparator) {
 		zone = r.ReplaceAllString(z, "")
-		reg[zone] = struct{}{}
+		if _, ok := reg[zone]; !ok {
+			reg[zone] = true
+			regions = append(regions, zone)
+		}
 	}
 
-	var regions []string
-	for k := range reg {
-		regions = append(regions, k)
-	}
 	return regions
 }
--- a/pkg/kube/volume/volume.go
+++ b/pkg/kube/volume/volume.go
@@ -349,10 +349,7 @@ func zoneToRegion(zone string) string {
 
 func zonesToRegions(zone string) []string {
 	reg := map[string]struct{}{}
-	// TODO: gocritic rule below suggests to use regexp.MustCompile but it
-	// panics if regex cannot be compiled. We should add proper test before
-	// enabling this below so that no change to this regex results in a panic
-	r, _ := regexp.Compile("-?[a-z]$") //nolint:gocritic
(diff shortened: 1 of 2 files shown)


pkg/kube/volume/volume_test.go
```diff
@@ -237,27 +237,6 @@ func (s *TestVolSuite) TestZoneToRegion(c *C) {
 		},
 	} {
 		reg := zonesToRegions(tc.zone)
-		c.Assert(slicesEqual(reg, tc.expectedRegion), Equals, true)
+		c.Assert(reg, DeepEquals, tc.expectedRegion)
```
viveksinghggits, line 240: @e-sumin do we know if deep equal returns true if two slices have same elements but in diff order? because that the scenario here and that's the reason I wrote that custom slice checker.

pkg/kube/volume/volume.go
```diff
@@ -349,10 +349,7 @@ func zoneToRegion(zone string) string {
 
 func zonesToRegions(zone string) []string {
 	reg := map[string]struct{}{}
-	// TODO: gocritic rule below suggests to use regexp.MustCompile but it
-	// panics if regex cannot be compiled. We should add proper test before
-	// enabling this below so that no change to this regex results in a panic
-	r, _ := regexp.Compile("-?[a-z]$") //nolint:gocritic
```
e-sumin, line 355: This code will panic in any case, because error is ignored, and `r` will be `nil` in case of wrong regex. So, disabling linter and usage of `Compile` did not solve the issue with possible panic.

Anyway, now we do have unit test which will detect panic.

pkg/kube/volume/volume_test.go
viveksinghggits, line 240: @e-sumin do we know if deep equal returns true if two slices have same elements but in diff order? because that the scenario here and that's the reason I wrote that custom slice checker.

viveksinghggits (changes requested):

pkg/kube/volume/volume_test.go
viveksinghggits: you might not find the failure when you just run the test once, try to run it >10 times using `-count` and you might see the problem.

pkg/kube/volume/volume_test.go
```diff
@@ -237,27 +237,6 @@ func (s *TestVolSuite) TestZoneToRegion(c *C) {
...
-			if o == t {
-				found = true
-			}
-		}
-		if !found {
-			return false
-		}
-	}
-	return true
-}
```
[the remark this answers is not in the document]
viveksinghggits: `DeepEqual` didn't seem to work for me. for the same slices having elems in diff order. 

https://go.dev/play/p/jVKm-nt90ZW

`DeepEquals` eventually calls `reflect.DeepEqual`.

PR #2735, opened 2024-03-08
refactor: Satisfy linter in zonesToRegion

e-sumin:
## Change Overview

There was annotation which disables `gocritic`. This linter could be enabled now, since we do have test which will detect panic.

Also it fixes order of regions returned by `zonesToRegions`

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/kube/volume/volume.go
+++ b/pkg/kube/volume/volume.go
@@ -348,16 +348,16 @@ func zoneToRegion(zone string) string {
 }
 
 func zonesToRegions(zone string) []string {
-	reg := map[string]struct{}{}
+	reg := map[string]bool{}
+	var regions []string
 	r := regexp.MustCompile("-?[a-z]$")
 	for _, z := range strings.Split(zone, RegionZoneSeparator) {
 		zone = r.ReplaceAllString(z, "")
-		reg[zone] = struct{}{}
+		if _, ok := reg[zone]; !ok {
+			reg[zone] = true
+			regions = append(regions, zone)
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/kube/volume/volume.go
```diff
@@ -348,19 +348,16 @@ func zoneToRegion(zone string) string {
 }
 
 func zonesToRegions(zone string) []string {
-	reg := map[string]struct{}{}
-	// TODO: gocritic rule below suggests to use regexp.MustCompile but it
-	// panics if regex cannot be compiled. We should add proper test before
-	// enabling this below so that no change to this regex results in a panic
-	r, _ := regexp.Compile("-?[a-z]$") //nolint:gocritic
```
e-sumin, line 355: This code will panic in any case, because error is ignored, and `r` will be `nil` in case of wrong regex. So, disabling linter and usage of `Compile` did not solve the issue with possible panic.

Anyway, now we do have unit test which will detect panic.

e-sumin:
I've accidentally pushed original PR under wrong branch name. After renaming branch via GH, it has automatically closed existing PR, so I had to create new one.

viveksinghggits (approved):
thank you @e-sumin :pray: .

PR #2754, opened 2024-03-15
build(ci): tag kanisterio/build image as latest or dev_latest tag

hairyhum:
## Change Overview

Currently we have to update the build image tag every time we build a new one. This is because we sometimes build "dev" versions of this image for testing.

Tagging 'latest' and 'dev_latest' makes it more convenient to use.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

Run the workflow to build `dev_latest` and `latest` tags. 

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


.github/workflows/kanister-image-build.yaml
```diff
@@ -7,6 +7,14 @@ on:
         description: 'Image tag in the format vx.x.x'
         required: true
         type: string
+      latest_tag:
```
viveksinghggits, line 10: this workflow is run automatically right?
how would someone decide if they want to tag the image to latest or something else?

hairyhum: No, this pipeline does not run automatically. `workflow_dispatch` trigger can only be triggered manually and the inputs have to be set by the person triggering it.

- `dev_latest` should be set when building the test image (e.g. to test changes in the image)
- `latest` should be set when building the image to be used for downstream users, such as `run-container.sh` and CI pipelines

viveksinghggits:
PR looks good, but just for my understanding 
> Currently we have to update the build image tag every time we build a new one.

where do we update the image tag? are you talking about updating it in the `Makefile`, if that is he case is that not required after this PR?

hairyhum:
> where do we update the image tag? are you talking about updating it in the `Makefile`, if that is he case is that not required after this PR?

@viveksinghggits it needs to be updated in `run_container.sh` and in the release pipeline (currently not in this repo). We don't always do that and that can randomly cause pipelines to fail if we forget.

viveksinghggits (approved):

PR #2758, opened 2024-03-18
Remove `cluster_name` field from kanister controller logs

viveksinghggits:
## Change Overview

When we use kanister log package `pkg/log` as part of the init function we try to set some variables that are used while logging a message. `cluster_name` is one of those variables, and to figure out `cluster_name` we get the `default` namespace and then use its `UID` field.
If someone just uses kando to let's say
- Set an output artifact
- Or to upload and object to object store

And they see that kando is trying to communicate with k8s apiserver, they might get confused because those operations don't involve communication with k8s apiserver. And if they have some security measures (network policies) in place to not let application namespace communicate with k8s apiserver, they would see a lot of warning messages about connection being dropped.

This PR fixes that by not logging the `cluster_name` field in kanister logs and this would result into log package not talking k8s apiserver.

*Note:*
We have been printing the cluster_name in the logs from quite some time, even though we don't expect this to have any side effects, if we think that the cluster_name field is important part of logs, I will figure out other solution.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

Build new `kando` binary and run it in a pod and use `tcpdump` in the same pod to make sure no connections are being made to apiserver.

```
kanister/pkg/log (unset_cluster_name_env*) » go test -check.f "LogSuite"            
{"level":"info","msg":"Test message","time":"2024-03-18T13:41:58.69816758+01:00"}
["hostname":"host" "service_name":"sservice" "version":"v0.0.1"]
OK: 13 passed
PASS
ok  	github.com/kanisterio/kanister/pkg/log	0.003s


```

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -12,7 +12,6 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/kanisterio/kanister/pkg/caller"
-	"github.com/kanisterio/kanister/pkg/config"
 	"github.com/kanisterio/kanister/pkg/field"
 )
 
@@ -92,10 +91,6 @@ func initEnvVarFields() {
 		if ev, ok := os.LookupEnv(e); ok {
 			envVarFields = field.Add(envVarFields, strings.ToLower(e), ev)
 		}
-	}
-
-	if clsName, err := config.GetClusterName(nil); err == nil {
-		envVarFields = field.Add(envVarFields, "cluster_name", clsName)
 	}
 }
 
--- a/pkg/log/log_test.go
+++ b/pkg/log/log_test.go
@@ -43,7 +43,11 @@ func (s *LogSuite) TestWithNilContext(c *C) {
 }
 
 func (s *LogSuite) TestInitEnvVarFields(c *C) {
-	os.Setenv("HOSTNAME", "host")
+	// HOSTNAME is set automatically when the test is run from CI
+	host := os.Getenv("HOSTNAME")
+	if host == "" {
+		os.Setenv("HOSTNAME", "host")
+	}
 	os.Setenv("SERVICE_NAME", "sservice")
 	os.Setenv("VERSION", "v0.0.1")
 
--- a/pkg/log/log_test.go
+++ b/pkg/log/log_test.go
@@ -42,6 +42,21 @@ func (s *LogSuite) TestWithNilContext(c *C) {
 	WithContext(nil).Print("Message")
 }
 
+func (s *LogSuite) TestInitEnvVarFields(c *C) {
+	os.Setenv("HOSTNAME", "host")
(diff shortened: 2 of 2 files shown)


pkg/log/log.go
```diff
@@ -93,10 +92,6 @@ func initEnvVarFields() {
 			envVarFields = field.Add(envVarFields, strings.ToLower(e), ev)
 		}
 	}
-
-	if clsName, err := config.GetClusterName(nil); err == nil {
-		envVarFields = field.Add(envVarFields, "cluster_name", clsName)
-	}
```
PrasadG193, line 99: Is it possible to remove init() only from kando command execution path? Can we keep that in the controller path?

viveksinghggits: I think what we need to decide is, is it ok to remove cluster_name from log fields or not. And then we can decide on the solution.

viveksinghggits: @pavannd1, @hairyhum what do you think?

hairyhum: I also think removing information from the logs where it's needed is not a good thing. If we can keep cluster_name for controller logs - that would be better.

pkg/log/log_test.go
```diff
@@ -42,6 +42,27 @@ func (s *LogSuite) TestWithNilContext(c *C) {
...
+	os.Setenv("VERSION", "v0.0.1")
+
+	initEnvVarFields()
+	fields := envVarFields.Fields()
+
+	c.Assert(len(fields), Equals, 3)
+
+	c.Assert(fields[0].Key(), Equals, "hostname")
+	c.Assert(fields[1].Key(), Equals, "service_name")
+	c.Assert(fields[1].Value(), Equals, "sservice")
```
e-sumin, line 61: I've added `Value()` test, I think it makes sense.

pkg/log/log_test.go
```diff
@@ -42,6 +42,27 @@ func (s *LogSuite) TestWithNilContext(c *C) {
...
+	if host == "" {
+		os.Setenv("HOSTNAME", "host")
+	}
+	os.Setenv("SERVICE_NAME", "sservice")
+	os.Setenv("VERSION", "v0.0.1")
+
+	initEnvVarFields()
+	fields := envVarFields.Fields()
+
+	c.Assert(len(fields), Equals, 3)
```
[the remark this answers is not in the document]
viveksinghggits: I will have a look.

viveksinghggits:
Closing this in favour of https://github.com/kanisterio/kanister/pull/2768

PR #2777, opened 2024-03-21
Set `cluster_name` var in logs vars separately and not in `init`

viveksinghggits:
## Change Overview

If we set the `cluster_name` field in log variables (extran fields that are printed in logs), inside the `init` function of `log` package, we would un necessarily be calling k8s api even from the places where we don't have to do anything with the k8s api.
To fix this, this commit introduces a new function that can be used to setup that field in the logs vars and that function will only be called from the places where we need the logs to have to the `cluster_name` field, for example `kanctl`, `controller` and `repository-controller`.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

Detailed steps are part of a comment below.


viveksinghggits:
In current version `kando` just run `kando` or `kando output a b` command and we would
see in the output of `tcpdump` that a connection is being made.

```
11:38:35.721885 IP ubukando-755d98cc67-wxgwp.33736 > kubernetes.default.svc.cluster.local.443: Flags [S], seq 626173369, win 64240, options [mss 1460,sackOK,TS val 17765482 ecr 0,nop,wscale 7], length 0
```

Build new binary and test these `kando` commands

## `kando`

no comms being made to apiserver, verified by output of `tcpdump`.

## `kando output ab cd`

no comms being made to apiserver, verified by looking at output of `tcpdump`.

## `kando location push`

```
kando location push --profile 'profilejson' hello.txt --path test-path
```

we see a lot of messages in the output of `tcpdump` but none of those
are to `kubernetes.default` but they are to s3 and that is obvious.

`tcpdump` logs are [here](https://gist.github.com/viveksinghggits/3e50121aa925ebd9cd38a35bd7e6c207).

## `kando location pull`

```
kando location pull --profile 'profilejson' --path test-path hello
```

and tcpdump logs can be verified to make sure no connections are made to apiserver.

`tcpdump` logs are [here](https://gist.github.com/viveksinghggits/d736d9bacac5b378693b247a4028f507).

hairyhum (approved):

pavannd1 (approved):
LGTM

PR #2778, opened 2024-03-21
Mark ephemeral pods created as part of kanister functions with JobID label

mabhi:
## Change Overview
Due to an ongoing work regarding identifying ephemeral pods those get spawned during various operations as part of some of the kanister functions, we need to mark these pod with some labels. One such pre-decided label for these pods is  `kanister.io/JobID`. This label will help:

-  Developers to verfiy whether required pod-specs were applied to these pods.
-  Identify pods, due to erroroneous cases stay lingering in memory without ever completing.
-  If multiple such type of pods are instantiated and we want to target / debug one out of these.

Right now the flow of this label information is, with the instantiation of actionset against a blueprint that services various kanister functions is triggered from an external application, and let's say if this label populated with a JobID, then this label [key: value] pair is added in entirerity to the temporary pod, so we can track it.

This PR handle cases when the `kube-task` labelled pods invoke kanister function that has a context.Context instance passed that contains a key-value pair corresponding to the pre-decided label, then it will make its way to the pods label selector map in addition to others if present already.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

1. Create Deployment -
```cat <<EOF | kubectl create -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  name: time-logger
  namespace: test
  labels:
    app: time-logger-dep
spec:
  replicas: 1
  selector:
    matchLabels:
      app: time-logger
  template:
    metadata:
      labels:
        app: time-logger
    spec:
      containers:
      - name: test-container
        image: containerlabs/aws-sdk
        command: ["sh", "-c"]
        args: ["while true; do for x in $(seq 1200); do date >> /var/log/time.log; sleep 1; done; truncate /var/log/time.log --size 0; done"]
EOF
```

2. Create Blueprint -
```
cat <<EOF | kubectl create -f -
kind: Blueprint
metadata:
  name: time-log-bp
  namespace: kanister
actions:
  backup:
    phases:
    - func: KubeTask
      name: backupToS3
      args:
        namespace: "{{ .Deployment.Namespace }}"
        image: busybox
        podOverride:
          containers:
          - name: container
            imagePullPolicy: IfNotPresent
        command: ["sh", "-c", "counter=1; while [ $counter -lt 20 ]; do echo $counter; ((counter++)); sleep 1; done"]
EOF
```
3. Create Actionset -
```
cat <<EOF | kubectl create -f -
kind: ActionSet
metadata:
  generateName: s3backup-
  namespace: kanister
  labels:
    kanister.io/JobID: hello-world

spec:
  actions:
  - name: backup
    blueprint: time-log-bp
    object:
      kind: Deployment
      name: time-logger
      namespace: test
EOF
```
4.  Watch ephemeral pod and check its label -
```
NAME                     READY   STATUS    RESTARTS   AGE   LABELS
kanister-job-j298f   1/1     Running   0          0s    createdBy=kanister,kanister.io/JobID=hello-world
```


pkg/function/kube_task.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/kanisterio/kanister/pkg/output"
 	"github.com/kanisterio/kanister/pkg/param"
 	"github.com/kanisterio/kanister/pkg/progress"
+	"strings"
```
pavannd1, line 33: This should be a part of the standard package group at the top.
@mabhi if you use VSCode, you can set it up to auto-correct this on save.

pkg/function/kube_task.go
```diff
@@ -64,6 +65,18 @@ func kubeTask(ctx context.Context, cli kubernetes.Interface, namespace, image st
 		Command:      command,
 		PodOverride:  podOverride,
 	}
+	// Mark labels to pods with prefix `jobPrefix`. Add the jobID as reference to the origin for the pod.
```
pavannd1, line 68: I don't see `jobPrefix` used. Is it `consts.LabelPrefix+"JobID"`?

mabhi: Yes, I corrected it.

pkg/function/kube_task.go
mabhi: Done

viveksinghggits:
@mabhi are we just going to handle `KubeTask` in this PR?

pkg/function/kube_task.go
```diff
@@ -64,6 +65,18 @@ func kubeTask(ctx context.Context, cli kubernetes.Interface, namespace, image st
...
+		for _, f := range fields.Fields() {
+			if strings.HasPrefix(f.Key(), consts.LabelPrefix) {
+				jobID := f.Value().(string)
+				if options.Labels == nil {
+					options.Labels = make(map[string]string)
+				}
+				options.Labels[consts.LabelPrefix+"JobID"] = jobID
+			}
+		}
+	}
```
viveksinghggits, line 79: is exporting this to another util helpful? my understanding is we would use this in other kanister functions as well.

mabhi: Yes, very true. I will move this.

mabhi:
> @mabhi are we just going to handle `KubeTask` in this PR?

Yes, accounted additional pods as suggested

pkg/kube/utils.go
```diff
@@ -171,3 +174,20 @@ func PVCContainsReadOnlyAccessMode(pvc *corev1.PersistentVolumeClaim) bool {
 
 	return false
 }
+
+// AddDebugLabelsToPodOptions adds additional label selector to `PodOptions`,
+// provided the context has a key starting with `keyPrefix`.
+func AddDebugLabelsToPodOptions(ctx context.Context, options *PodOptions, keyPrefix, keySuffix string) {
```
PrasadG193, line 180: 1. Since the values are extracted from context, the function name should suggest that. The function name should be self-explanatory.
2. Instead of passing prefix and suffix as a parameter to the function, its better to just pass key string. The key construction should be done by caller.
```suggestion
func AddLabelsToPodOptionFromContext(ctx context.Context, options *PodOptions, labelKey string) {
```

pkg/kube/utils.go
```diff
@@ -171,3 +174,20 @@ func PVCContainsReadOnlyAccessMode(pvc *corev1.PersistentVolumeClaim) bool {
...
+		for _, f := range fields.Fields() {
+			if strings.HasPrefix(f.Key(), keyPrefix) {
+				value := f.Value().(string)
+				if options.Labels == nil {
+					options.Labels = make(map[string]string)
+				}
+				options.Labels[keyPrefix+keySuffix] = value
+			}
+		}
+	}
```
PrasadG193, line 192:
```suggestion
	if fields == nil {
	        return
	}
	for _, f := range fields.Fields() {
		if !strings.HasPrefix(f.Key(), keyPrefix) {
		        continue
		}
		value := f.Value().(string)
		if options.Labels == nil {
			options.Labels = make(map[string]string)
		}
		options.Labels[labelKey] = value
	}
```
Its best practice to reduce nesting in Go - https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

mabhi: Done

pkg/kube/utils.go
PrasadG193: Right, we shouldn't add this check in this utility function. The function should just do one thing - add labels in the option with the given key. The callers should do the validation on keys if required (maybe separate function for validateLabelKeys()).

pkg/function/kube_task.go
```diff
@@ -64,6 +64,8 @@ func kubeTask(ctx context.Context, cli kubernetes.Interface, namespace, image st
 		Command:      command,
 		PodOverride:  podOverride,
 	}
+	// Mark labels to pods with prefix `kanister.io`. Add the jobID as reference to the origin for the pod.
+	kube.AddLabelsToPodOptionsFromContext(ctx, options, consts.LabelPrefix, "JobID")
```
PrasadG193, line 68: Lets declare `JobID` label key as const

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +113,66 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
 	cancel()
 }
 
+// TestPodRunnerWithJobIDDebugLabelForSuccessCase: This test adds a debug entry (kanister.io/JobID) into the context and verifies the
+// pod got created with corresponding label using the entry or not.
+func (s *PodRunnerTestSuite) TestPodRunnerWithJobIDDebugLabelForSuccessCase(c *C) {
+	ctx, cancel := context.WithCancel(context.Background())
+	randomUUID := "xyz123"
+	ctx = field.Context(ctx, consts.LabelPrefix+"JobID", randomUUID)
```
PrasadG193, line 121:
```suggestion
	ctx = field.Context(ctx, path.Join(consts.LabelPrefix, "JobID"), randomUUID)
```

pkg/kube/utils.go
mabhi: Done

pkg/function/kube_task.go
```diff
@@ -64,6 +68,9 @@ func kubeTask(ctx context.Context, cli kubernetes.Interface, namespace, image st
 		Command:      command,
 		PodOverride:  podOverride,
 	}
+	// Mark labels to pods with prefix `kanister.io`. Add the jobID as reference to the origin for the pod.
+	validateFunc := validateLabelKeyIsPresentFunc(consts.LabelPrefix)
+	kube.AddLabelsToPodOptionsFromContext(ctx, options, path.Join(consts.LabelPrefix, jobIDSuffix), validateFunc)
```
PrasadG193, line 73: @mabhi I didn't mean passing validateFunc as an arg to the function. The `kube.AddLabelsToPodOptionsFromContext()` util function should just do one thing, add labels to the options from context. The validation part should be done separately, before calling that function.

mabhi: Simplied the validation and add-label-to-pod-options functions.

pkg/kube/utils.go
```diff
@@ -171,3 +174,30 @@ func PVCContainsReadOnlyAccessMode(pvc *corev1.PersistentVolumeClaim) bool {
...
+func ValidateLabelKeyIsPresentFromContext(ctx context.Context, keyPrefix string) (bool, string) {
+	fields := field.FromContext(ctx)
+	if fields == nil {
+		return false, ""
+	}
+	for _, f := range fields.Fields() {
+		if strings.HasPrefix(f.Key(), keyPrefix) {
+			return true, f.Value().(string)
+		}
+	}
```
PrasadG193, line 201: I am sorry, I think I am missing something and that is creating confusion. Why do we need these checks? Do we not know the complete key name that is present in the context? Why do we need to match the prefix only?

mabhi: Correct. It should match the entire key and not prefix. 
Fixed.

pkg/kube/utils.go
```diff
@@ -171,3 +173,30 @@ func PVCContainsReadOnlyAccessMode(pvc *corev1.PersistentVolumeClaim) bool {
 
 	return false
 }
+
+// AddLabelsToPodOptions adds additional label selector to `PodOptions`. It extracts the value from the context
```
PrasadG193, line 177:
```suggestion
// AddLabelsToPodOptionsFromContext adds an additional label to `PodOptions`. It extracts the value from the context
```

pkg/kube/utils.go
```diff
@@ -171,3 +173,30 @@ func PVCContainsReadOnlyAccessMode(pvc *corev1.PersistentVolumeClaim) bool {
...
+		}
+	}
+	// Not found
+	if targetValue == "" {
+		return
+	}
+	if options.Labels == nil {
+		options.Labels = make(map[string]string)
+	}
+	options.Labels[targetKey] = targetValue
```
PrasadG193, line 201:
```suggestion
	if options.Labels == nil {
		options.Labels = make(map[string]string)
	}
	for _, f := range fields.Fields() {
		if f.Key() == targetKey {
			options.Labels[targetKey] = f.Value().(string)
			return
		}
	}
```

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,88 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
+			name:        "target key present",
+			targetKey:   path.Join(consts.LabelPrefix, "JobID"),
+			targetValue: randomUUID,
+			hasError:    false,
+		},
+		{
+			name:        "target key not present",
+			targetKey:   path.Join(consts.LabelPrefix, "NonJobID"),
+			targetValue: "some-other-value",
+			hasError:    true,
```
PrasadG193, line 137: The field is declared as `hasError` but its passed to the function as `ignoreError`. I also couldn't figure out why we need this since we don't return any error from the AddLabelsToPodOptionsFromContext(). It doesn't seem to be. used in any validation checks.

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,88 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
+	randomUUID := "xyz123"
+	for _, tc := range []struct {
+		name        string
+		targetKey   string
+		targetValue string
+		hasError    bool
+	}{
+		{
+			name:        "target key present",
+			targetKey:   path.Join(consts.LabelPrefix, "JobID"),
```
PrasadG193, line 129:
```suggestion
			targetKey:   path.Join(consts.LabelPrefix, jobIDSuffix),
```

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,88 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
+			Name:      podName,
+			Command:   []string{"sh", "-c", "tail -f /dev/null"},
+		}
+		deleted := make(chan struct{})
+		cli.PrependReactor("delete", "pods", func(action testing.Action) (handled bool, ret runtime.Object, err error) {
+			c.Log("Pod deleted due to Context Cancelled")
+			close(deleted)
+			return true, nil, nil
+		})
+		var targetKey = path.Join(consts.LabelPrefix, "JobID")
```
PrasadG193, line 165:
```suggestion
		var targetKey = path.Join(consts.LabelPrefix, jobIDSuffix)
```

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,88 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
 		return nil, nil
 	}
 }
+
+func afterPodRunTestKeyPresentFunc(labelKey, labelValue string, ignoreError bool, ch chan struct{}) func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
+	return func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
+		<-ch
+		if ignoreError {
+			return nil, nil
+		}
```
PrasadG193, line 191: THis can be removed since we made the function generic to support any labels.

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,88 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
+		hasError    bool
+	}{
+		{
+			name:        "target key present",
+			targetKey:   path.Join(consts.LabelPrefix, "JobID"),
+			targetValue: randomUUID,
+			hasError:    false,
+		},
+		{
+			name:        "target key not present",
```
PrasadG193, line 134: We need to refactor tests since we made the function generic one.
This test case seems redundant.
We should add a test case where the key is missing from context and perform validation.

pkg/kube/utils.go
mabhi: With this suggestion, I am thinking even if the key is not present, the option.Labels is initialized. I did not want to do this initialization in case my targetKey is absent. wdyt ?

pkg/kube/pod_runner_test.go
mabhi: `jobIDSuffix` is in different package. Do you meant, that I should make the const public like in consts.

pkg/kube/pod_runner_test.go
mabhi: Same reasoning as above.

pkg/kube/pod_runner_test.go
mabhi: Removed

pkg/kube/utils.go
mabhi: Done

pkg/kube/utils.go
mabhi: Done

pkg/function/kube_task.go
```diff
@@ -64,7 +67,8 @@ func kubeTask(ctx context.Context, cli kubernetes.Interface, namespace, image st
 		Command:      command,
 		PodOverride:  podOverride,
 	}
-
+	// Mark labels to pods with prefix `kanister.io`. Add the jobID as reference to the origin for the pod.
```
PrasadG193, line 70: This comment is no longer valid, either remove or correct the comment.

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,88 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
 	cancel()
 }
 
+// TestPodRunnerWithDebugLabelForSuccessCase: This test adds a debug entry into the context and verifies the
```
PrasadG193, line 117:
```suggestion
// TestPodRunnerWithDebugLabelForSuccessCase adds a debug entry into the context and verifies the
```
Let's keep the consistency function comment formats across the project. Let's not use `:`.

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,88 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
 	}
 }
+
+func afterPodRunTestKeyPresentFunc(labelKey, validationErrorMsg string, expectedLabel bool, ch chan struct{}) func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
+	return func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
+		<-ch
+
+		_, got := pc.Pod().Labels[labelKey]
+		if got != expectedLabel {
+			return nil, errors.New(validationErrorMsg)
```
PrasadG193, line 197:
```suggestion
			return nil, errors.New("Got different label than expected")
```
We don't need to declare `validationErrorMsg` per test case. We can use a generic error message.

PrasadG193 (approved):
LGTM, minor suggestions about formatting and optimization in the tests.

pkg/kube/pod_runner_test.go
mabhi: Done

pkg/function/kube_task.go
mabhi: Done

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,84 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
+func afterPodRunTestKeyPresentFunc(labelKey string, expectedLabel bool, ch chan struct{}) func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
+	return func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
+		<-ch
+		_, got := pc.Pod().Labels[labelKey]
+		if got != expectedLabel {
+			return nil, errors.New("Got different label than expected")
+		}
+		return nil, nil
+	}
+}
```
PrasadG193, line 197: We can also add an additional validation to check the label value
```suggestion
func afterPodRunTestKeyPresentFunc(labelKey string, isLabelExpected bool, expectedLabelValue string, ch chan struct{}) func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
	return func(ctx context.Context, pc PodController) (map[string]interface{}, error) {
		<-ch
		labelValue, found := pc.Pod().Labels[labelKey]
		if found != isLabelExpected {
			return nil, errors.New("Found unexpected label")
		}
		if labelValue != expectedLabelValue {
		        return nil, errors.New("Found label doesn't match with expected label")
		}
		return nil, nil
	}
}
```

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,84 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
+		cli.PrependReactor("delete", "pods", func(action testing.Action) (handled bool, ret runtime.Object, err error) {
+			c.Log("Pod deleted due to Context Cancelled")
+			close(deleted)
+			return true, nil, nil
+		})
+		AddLabelsToPodOptionsFromContext(ctx, po, tc.targetKey)
+		pr := NewPodRunner(cli, po)
+		errorCh := make(chan error)
+		go func() {
+			_, err := pr.Run(ctx, afterPodRunTestKeyPresentFunc(tc.targetKey, tc.expectedLabel, deleted))
```
PrasadG193, line 172: We can pass expected label value and validate if we are receiving the same 
```suggestion
			_, err := pr.Run(ctx, afterPodRunTestKeyPresentFunc(tc.targetKey, tc.isLabelExpected, tc.targetLabel, deleted))
```

PrasadG193 (commented):
Added suggestions about additional validation we can add in the test.

pkg/kube/pod_runner_test.go
```diff
@@ -109,9 +114,84 @@ func (s *PodRunnerTestSuite) TestPodRunnerForSuccessCase(c *C) {
...
+		contextKey    string
+		contextValue  string
+		expectedLabel bool
+	}{
+		{
+			name:          "target key (kanister.io/JobID) present in pod labels",
+			targetKey:     path.Join(consts.LabelPrefix, jobIDSuffix),
+			contextKey:    path.Join(consts.LabelPrefix, jobIDSuffix),
+			contextValue:  "xyz123",
+			expectedLabel: true,
```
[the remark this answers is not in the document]
mabhi: Done

pkg/kube/pod_runner_test.go
mabhi: Done

pkg/kube/pod_runner_test.go
mabhi: Done

PrasadG193 (approved):

PR #2792, opened 2024-03-27
build(security): add nightly run of govulncheck with `continue-on-error: false`

### LINKED ISSUE #2782: Nightly runs for govulnscan

Already merged code should be scanned with `govulnscan` as well. E.g. when a new vulnerability emerges in the mean-time.

(Github Action for `govulnscan` was added in https://github.com/kanisterio/kanister/pull/2752)

Nightly runs should be added. Possibly in a separate pipeline and with `continue-on-error: false`

**Describe alternatives you've considered**

Automatically creating Github issues as reminders on failures might be useful. It could also be used for notifications when a Github Team is tagged in the text. Probably a separate issue should be created though if needed.

**Environment**

**Additional context**

hairyhum:
## Change Overview

Add nightly trigger to govulncheck

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2782 

## Test Plan

Make sure action run on PR has continue-on-error: true
Check action run nightly after merge

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


julio-lopez (commented):
This seem to be overloading the 'dependency review' workflow. 

The dependency review workflow is meant to be run on PRs. AFAIK, it only makes sense to run `actions/dependency-review-action` in PRs, or anything that results in changing the dependencies, but a PR workflow would be the gate for that.

It may make sense to split this into 2 separate workflows in either of the following ways (there may be others)

- [probably preferred] Either have a separate workflow for the nightly job that **only** runs `govuln` and have the name of the workflow reflect that is a nightly.
- Or maybe include `govuln` in one of the existing periodic/nightly workflows.
- Alternatively, have a separate workflow for govuln only, that runs both on PRs and nightly as well.

hairyhum:
> Alternatively, have a separate workflow for govuln only, that runs both on PRs and nightly as well.

That's exactly what this workflow is. It triggers on PRs and nightlies, but it's not a part of any other pipeline.

hairyhum:
> Alternatively, have a separate workflow for govuln only, that runs both on PRs and nightly as well.

Right. This makes more sense. Moved govulncheck to a separate workflow.

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+      - id: govulncheck
+        name: 'Govulncheck'
+        uses: golang/govulncheck-action@3a32958c2706f7048305d5a2e53633d7e37e97d0 # v1.0.2
+        continue-on-error: ${{ github.event_name == 'pull_request' }}
+        with:
+          repo-checkout: false
+          cache: false
```
julio-lopez, line 24: nit: add a newline
```suggestion
          cache: false

```

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,24 @@
+
+name: 'govulncheck'
+on:
+  ## Nightly
+  schedule:
+    - cron: '0 0 * * *'
```
julio-lopez, line 6: This is UTC IIUC.
It may be a good idea to add some fudge factor here. It does not necessarily need to run exactly at midnight UTC.

julio-lopez (approved):
Comments can be addressed separately

.github/workflows/govulncheck.yaml
hairyhum: It can be anything. 00:00 UTC is as good a time as any.

PR #2825, opened 2024-04-17
support Azure sovereign clouds

KastenMike:
## Change Overview

Currently it's not possible to create these clients in sovereign clouds e.g. AzureChina because the cloud isn't specified and defaults to the Public cloud. This leads to errors like `adal: Failed to execute the refresh request. Error = 'Post \https://login.microsoftonline.com/****/oauth2/token` when the URL for AzureChina is `https://login.chinacloudapi.cn/`

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/blockstorage/azure/auth.go
+++ b/pkg/blockstorage/azure/auth.go
@@ -7,6 +7,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/kanisterio/kanister/pkg/blockstorage"
+	"github.com/kanisterio/kanister/pkg/consts"
 )
 
 const ActiveDirectory = "activeDirectory"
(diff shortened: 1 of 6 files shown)


pkg/blockstorage/azure/auth.go
```diff
@@ -55,12 +56,26 @@ type AzureAuthenticator interface {
...
 
-func NewAzureAuthenticator(config map[string]string) (AzureAuthenticator, error) {
+type NewManagedIdentityCredentialFunc func(opts *azidentity.ManagedIdentityCredentialOptions) (*azidentity.ManagedIdentityCredential, error)
+type NewClientSecretCredentialFunc func(tenantID, clientID, clientSecret string, opts *azidentity.ClientSecretCredentialOptions) (*azidentity.ClientSecretCredential, error)
+type AzIdentityType struct {
+	NewManagedIdentityCredential NewManagedIdentityCredentialFunc
+	NewClientSecretCredential    NewClientSecretCredentialFunc
+}
+
+func NewAzureAuthenticator(config map[string]string, identity *AzIdentityType) (AzureAuthenticator, error) {
```
hairyhum, line 66: This function is only called with `nil` as a second argument.

pkg/blockstorage/azure/client.go
```diff
@@ -49,7 +51,20 @@ var (
...
 
-func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
+type ComputeNewClientFactoryFunc func(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*armcompute.ClientFactory, error)
+type SubscriptionsNewClientFactoryFunc func(credential azcore.TokenCredential, options *arm.ClientOptions) (*armsubscriptions.ClientFactory, error)
+type AzClientFactoriesType struct {
+	ComputeNewClientFactory       ComputeNewClientFactoryFunc
+	SubscriptionsNewClientFactory SubscriptionsNewClientFactoryFunc
+}
+
+func NewClient(ctx context.Context, config map[string]string, azClientFactories *AzClientFactoriesType) (*Client, error) {
```
hairyhum, line 61: This function is only called as `nil` as a third argument

pkg/blockstorage/azure/client.go
```diff
@@ -82,12 +97,14 @@ func NewClient(ctx context.Context, config map[string]string) (*Client, error) {
 		return nil, err
 	}
 	cred := authenticator.GetAuthorizer()
-	computeClientFactory, err = armcompute.NewClientFactory(subscriptionID, cred, nil)
+	opts := &arm.ClientOptions{ClientOptions: policy.ClientOptions{}}
+	opts.ClientOptions.Cloud = GetCloudConfig(config[blockstorage.AzureCloudEnvironmentID])
```
hairyhum, line 101: I believe this code makes any difference because it's not possible to pass `blockstorage.AzureCloudEnvironmentID` from the https://github.com/kanisterio/kanister/blob/master/pkg/function/create_volume_snapshot.go#L298

hairyhum (commented):
Since we're moving away from using this code as a library and also plan to remove it completely in favour of CSI, I don't see how this change would work, because internal usage of azure provider doesn't have the configuration to select the clouds.
If that's something we need to do, then we need to extend the profile configuration to support new config keys.

pkg/blockstorage/azure/auth.go
KastenMike: It is overridden in unit tests

pkg/blockstorage/azure/client.go
KastenMike: It is overridden in unit tests

pkg/blockstorage/azure/client.go
KastenMike: ok seems it's dead code / too much effort to change. Closing the PR

PR #2842, opened 2024-04-23
Take control over Kopia log level

e-sumin:
## Change Overview

When building Kopia command line take LogLevel and FileLogLevel from env variables if it not specified in parameters

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pavannd1 (commented):
Is it possible to add these flags to one of the command unit tests we have?

e-sumin:
> Is it possible to add these flags to one of the command unit tests we have?

Good idea. I've added new unit test for common args builder.

hairyhum (approved):
LGTM

pkg/kopia/command/common.go
```diff
@@ -15,16 +15,34 @@
...
+)
+
+func LogLevel() string {
+	return os.Getenv(LogLevelVarName)
+}
+
+func FileLogLevel() string {
+	return os.Getenv(FileLogLevelVarName)
+}
+
```
plar, line 39: Hi Eugen,

First off, thanks for log flags! I've missed them recently (and hard-coded locally for tests) :).

What do you think about encapsulating the defaulting logic within envvar utility funcs?
For example, we could use a generic function like `nonEmptyOrDefault` to handle defaults more elegantly.

Here’s an example of what this might look like:

```
const (
...
    DefaultLogLevel     = "info"
    DefaultFileLogLevel = "debug"
)

func nonEmptyOrDefault[T comparable](t T, def T) T {
    var empty T
    if t != empty {
        return t
    }
    return def
}

func getEnv(name, def string) string {
    return nonEmptyOrDefault(os.Getenv(name), def)
}

func LogLevel() string {
    return getEnv(LogLevelVarName, DefaultLogLevel)
}

func FileLogLevel() string {
    return getEnv(FileLogLevelVarName, DefaultFileLogLevel)
}
```

and after we can refactor `commonArgs` as:

OLD
```
        ...
	logLevel := cmdArgs.LogLevel
	if logLevel == "" {
		logLevel = LogLevel()
	}

	if logLevel != "" {
		c = c.AppendLoggableKV(logLevelFlag, logLevel)
	} else {
		c = c.AppendLoggableKV(logLevelFlag, LogLevelError)
	}

	fileLogLevel := cmdArgs.FileLogLevel
	if fileLogLevel == "" {
		fileLogLevel = FileLogLevel()
	}

	if fileLogLevel != "" {
		c = c.AppendLoggableKV(fileLogLevelFlag, fileLogLevel)
	}
	...
```

NEW
```
        ...
	logLevel := nonEmptyOrDefault(cmdArgs.LogLevel, LogLevel())
	c = c.AppendLoggableKV(logLevelFlag, logLevel)

	fileLogLevel := nonEmptyOrDefault(cmdArgs.FileLogLevel, FileLogLevel())
	c = c.AppendLoggableKV(fileLogLevelFlag, fileLogLevel)
	...
```

pkg/kopia/command/common.go
```diff
@@ -15,16 +15,34 @@
...
 	"github.com/kanisterio/kanister/pkg/field"
 	"github.com/kanisterio/kanister/pkg/log"
 	"github.com/kanisterio/kanister/pkg/logsafe"
 )
 
+const (
+	// LogLevelVarName is the environment variable that controls Kopia log level.
+	LogLevelVarName = "KOPIA_LOG_LEVEL"
+	// FileLogLevelVarName is the environment variable that controls Kopia file log level.
+	FileLogLevelVarName = "KOPIA_FILE_LOG_LEVEL"
```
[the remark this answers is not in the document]
viveksinghggits: yes, I remember we discussed something like this.

viveksinghggits, line 29: Just for my understand Eugen, can you please very briefly talk about what is the difference between Log Level and File Log Level?

pkg/kopia/command/common.go
```diff
@@ -15,16 +15,34 @@
...
+)
+
+func LogLevel() string {
+	return os.Getenv(LogLevelVarName)
```
viveksinghggits, line 33: how is this  environment variable set? should we default to something if this env var is not set?

pkg/kopia/command/common.go
```diff
@@ -40,12 +58,26 @@ func stringSliceCommand(args logsafe.Cmd) []string {
 func commonArgs(cmdArgs *CommandArgs) logsafe.Cmd {
 	c := logsafe.NewLoggable(kopiaCommand)
 
-	if cmdArgs.LogLevel != "" {
-		c = c.AppendLoggableKV(logLevelFlag, cmdArgs.LogLevel)
+	logLevel := cmdArgs.LogLevel
```
viveksinghggits, line 61: do you think this requires change in the consumers of this function for example `SnapshotCreate` etc? Are we planning to do that in other PR?

pkg/kopia/command/common.go
hairyhum: https://kopia.io/docs/advanced/logging/#controlling-log-level

pkg/kopia/command/common.go
e-sumin: accepted, done.

pkg/kopia/command/common.go
e-sumin: Thanks for @hairyhum for link.

Just to have an answer right here:
```
You can control how much data is written to console and log files by using flags:

--log-level - sets log level for console output (defaults to info)
--file-log-level - sets log level for file output (defaults to debug)
```

pkg/kopia/command/common.go
e-sumin: In updated code, default value is provided.

pkg/kopia/command/common.go
e-sumin: I think it is not required. 

Behavior before change:
Either level from var OR error

Behavior after change:
Either level from var **OR level from env** OR error

Taking level from env is needed only when debugging. 
I went through the invokers and LogLevel is not set there. But if it will be set there to some value, I believe, decision of priority between env and their value should be done on their side.

pkg/kopia/command/common.go
```diff
@@ -15,16 +15,48 @@
...
+}
+
+// LogLevel will return either value from env or "error" as default value
+func LogLevel() string {
+	return GetEnvOrDefault(LogLevelVarName, LogLevelError)
+}
+
+// FileLogLevel will return value from env
+func FileLogLevel() string {
+	return os.Getenv(FileLogLevelVarName)
```
viveksinghggits, line 51: what does this default to?

e-sumin: empty string, and if it is empty, we won't pass this flag at all (as it was before my changes)
```
	if fileLogLevel != "" {
		c = c.AppendLoggableKV(fileLogLevelFlag, fileLogLevel)
	}
```

viveksinghggits (approved):

pkg/kopia/command/common_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"os"
+
+	"gopkg.in/check.v1"
+)
+
+type CommonUtilsSuite struct{}
+
+var _ = check.Suite(&CommonUtilsSuite{})
+
+func (s *CommonUtilsSuite) TestCommonArgs(c *check.C) {
```
julio-lopez, line 27: These tests do not need to use `check.v1` and could be built using the support in the plain Go `testing` package.

Since this is (was) a new test, wouldn't it make sense to avoid `check.v1` here?

julio-lopez (commented):
@e-sumin 
+1: LG 👍 Thanks for doing this.

A couple of minor comments / questions.

PR #2857, opened 2024-05-03
Add ephemeral package with environment variable registration

### LINKED ISSUE #2856: Add environment variable registration mechanism for ephemeral pods
Ephemeral pods launched by Kanister need to have environment variables injected based on the outer OS environment variables (parent pod that launched these ephemeral pods) or based on static environment variables that need to be added. 

The possible set of conditional environment variables needs to be registered and if found in the outer OS then should be injected. The static environment variables, when registered, should always be added to the container environment variables.

Add an ephemeral package which has a registration scheme for environment variables which can be used outside Kanister to register which environment variables you'd like injected.

**Additional context**
This helps to get the correct FIPS required environment variables into the ephemeral pods to ensure the crypto backend is started in FIPS mode.

kidgilson:
## Change Overview

This PR introduces the `ephemeral` package meant to handle registering and applying ephemeral pod specific fields. 

The first use-case was to introduce FIPS related environment variables that need to be set on all ephemeral pods to ensure the crypto backend (OpenSSL in this case) is started with the FIPS provider enabled and fails otherwise. 

The registration mechanism leaves it open to register conditional or static environment variables with the appropriate helper functions to help apply them. The only conditional environment variable helper at this point is the `OSEnvVar` which looks to see if the environment variable is set on the OS and applies it to the ephemeral pod if so. This usage is the main need for setting the FIPS environment variables since they're already present on the parent pod launching the ephemeral pods.

All places that used the `kube.PodOptions` type had the `ephemeral.GlobalEnvVars()` function added which adds all registered environment variables to the pod.

Two convenience functions `RegisterOSEnvVar` and `RegisterStaticEnvVar` were added to help facilitate the registering of environment variables from packages which import this.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [X] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2856 

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/controllers/repositoryserver/utils.go
+++ b/pkg/controllers/repositoryserver/utils.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/kanisterio/kanister/pkg/consts"
+	"github.com/kanisterio/kanister/pkg/ephemeral"
 	"github.com/kanisterio/kanister/pkg/format"
 	"github.com/kanisterio/kanister/pkg/kopia/command/storage"
 	"github.com/kanisterio/kanister/pkg/kube"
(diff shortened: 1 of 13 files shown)


kidgilson:
Closing in favor of https://github.com/kanisterio/kanister/pull/2874 which generalizes the `ephemeral` package for more uses.

PR #2874, opened 2024-05-08
Add new ephemeral package for registering and applying PodOptions changes

### LINKED ISSUE #2856: Add environment variable registration mechanism for ephemeral pods

kidgilson:
## Change Overview

This PR introduces the ephemeral package meant to handle registering and applying ephemeral pod specific fields.

The first use-case was to introduce FIPS related environment variables that need to be set on all ephemeral pods to ensure the crypto backend (OpenSSL in this case) is started with the FIPS provider enabled and fails otherwise.

The registration mechanism is made to be generic by using an `Applier` interface which is used to register different ways of applying changes to a `*kube.PodOptions` or a `*corev1.Container` .

For now convenience functions related to registering environment variable Applier's are added. The conditional environment variable helper is the OSEnvVar function which looks to see if the environment variable is set on the OS and applies it to the ephemeral pod if so. This usage is the main need for setting the FIPS environment variables since they're already present on the parent pod launching the ephemeral pods.

All places that used the `kube.PodOptions` and `corev1.Container` types have the ephemeral.Options.Apply() function added which adds all registered Appliers (only environment variables at this point) to the pod.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [X] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2856 

## Test Plan

- [ ] :muscle: Manual
- [X] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/ephemeral/envvar.go
+++ b/pkg/ephemeral/envvar.go
@@ -1,13 +1,28 @@
+// Copyright 2024 The Kanister Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package ephemeral
 
 import (
 	"os"
 
-	"github.com/kanisterio/kanister/pkg/kube"
 	corev1 "k8s.io/api/core/v1"
+
+	"github.com/kanisterio/kanister/pkg/kube"
 )
 
-// OSEnvVar creates an ApplierSet to set an environment variable if its present
+// OSEnvVar creates an ApplierSet to set an environment variable if it's present
 // in the current environment.
 func OSEnvVar(name string) ApplierSet {
 	return ApplierSet{
--- a/pkg/ephemeral/envvar.go
+++ b/pkg/ephemeral/envvar.go
@@ -7,52 +7,55 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// OSEnvVar registers an environment variable if its present on the current OS to be applied to the PodOptions.
-func OSEnvVar(name string) Applier {
-	return ApplierFunc(func(options any) {
-		if val, present := os.LookupEnv(name); present {
-			switch v := options.(type) {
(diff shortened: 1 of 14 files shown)


bathina2 (approved):
LGTM

shwetha-97 (approved):
LGTM

pkg/ephemeral/ephemeral.go
```diff
@@ -0,0 +1,115 @@
...
+
+// FiltererFunc is a function which implements the Filterer interface and can be
+// used to generically filter PodOptions to manipulate using the ApplierList.
+type FiltererFunc[T Constraint] func(*T) bool
+
+func (f FiltererFunc[T]) Filter(options *T) bool {
+	return f(options)
+}
+
+// Filter applies the Applier's if the Filterer criterion is met.
```
pavannd1, line 90:
```suggestion
// Filter applies the Appliers if the Filterer criterion is met.
```

pkg/ephemeral/ephemeral.go
```diff
@@ -0,0 +1,115 @@
...
+			return
+		}
+
+		for _, applier := range appliers {
+			applier.Apply(options)
+		}
+	})
+}
+
+// PodOptionsNameFilter is a Filterer that filters based on the PodOptions.Name.
```
pavannd1, line 103: `PodOptions.Name` is the pod name. Wonder if we should explicitly state that in the comment

pkg/ephemeral/envvar.go
```diff
@@ -0,0 +1,61 @@
+package ephemeral
+
+import (
+	"os"
+
+	"github.com/kanisterio/kanister/pkg/kube"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// OSEnvVar creates an ApplierSet to set an environment variable if its present
```
pavannd1, line 10:
```suggestion
// OSEnvVar creates an ApplierSet to set an environment variable if it's present
```

pkg/ephemeral/ephemeral.go
kidgilson: Sounds good. I'll add more to the comment.

pavannd1 (approved):

Issue #2882, opened 2024-05-10
[BUG] ActionSet artifacts fail to populate in multi-action actionsets


hairyhum:
**Describe the bug**
When using actionsets with multiple actions, some actions fail to populate artifacts in the action status.
This is happening because actionset reconciliation expects the entire actionset to be complete when doing that. If there are multiple actions it's not the case.

**To Reproduce**
1. Create an actionset referencing multiple actions. E.g. example blueprint from timelog and example blueprint from mysql.
1. Monitor the action status
1. See that timelog action did not publish the artifacts, but instead has the template in the artifacts field
1. See error in the logs showing "ActionSet cannot be complete if any actions are not complete"

**Expected behavior**
All actions in actionsets have their artifacts 

**Additional context**
The call tree is:
```
pkg/controller/controller.go::renderActionsetArtifacts which is an action-specific 
    pkg/reconcile/reconcile.go::ActionSet which is not action-specific
      pkg/validate/validate.go::actionSetStatus
```

**Related issues**
https://github.com/kanisterio/kanister/issues/2270
https://github.com/kanisterio/kanister/issues/1364

PR #2883, opened 2024-05-10
fix: do not try to update actionset state when publishing artifacts

hairyhum:
## Change Overview

If there are multiple actions running in parallel, they will populate artifacts while another action is still running.
This causes a conflict in update of actionset state.

Actionset state should be updated only when all actions complete by onUpdateActionSet

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- Fixes https://github.com/kanisterio/kanister/issues/2882

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E


hairyhum:
TOOD: add tests

pkg/controller/controller.go
```diff
@@ -705,6 +684,53 @@ func (c *Controller) renderActionsetArtifacts(ctx context.Context,
...
+		if coreErr != nil || deferErr != nil {
+			ras.Status.State = crv1alpha1.StateFailed
+			return nil
+		}
+
+		for _, as := range ras.Status.Actions {
+			for _, p := range as.Phases {
+				if p.State != crv1alpha1.StateComplete {
+					log.WithContext(ctx).Print(
+						"Finished action, but other action's phase is still running. Not setting state to complete.",
```
viveksinghggits, line 711: I have a question about this, in the PR desc also it's mentioned that `Actionset state should be updated only when all actions complete by onUpdateActionSet`. 

What I am not able to understand is, the relation between two actions. 
`Finished action, but other action's phase is still running.` if backup action is complete should it not be marked complete? Which other action would be wait for before marking backup action to be complete?

hairyhum: I was hoping we can make it work on `onUpdateActionSet`, but there are unresolved races there. I will change the description.

hairyhum: > if backup action is complete should it not be marked complete

There is no status per action, adding which would be a bigger change than a bugfix. We have status per phase and per actionset currently. So we can't mark an action to be complete (which would make things a bit easier TBH).

PrasadG193 (approved):
LGTM

PR #2888, opened 2024-05-16
Improve progress tracking

e-sumin:
## Change Overview

Currently we are tracking only amounts of data uploaded during action or phase. This PR brings new counters which allows us to track both, uploaded and downloaded data. 

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/apis/cr/v1alpha1/types.go
+++ b/pkg/apis/cr/v1alpha1/types.go
@@ -202,7 +202,7 @@ type PhaseProgress struct {
 	// EstimatedUploadSizeB represents the total estimated size of data in Bytes
 	// that will be uploaded during the phase execution.
 	// This field will be empty for phases which do not involve data movement.
-	EstimatedUploadSizeB int64 `json:"estinatedUploadSizeB,omitempty"`
+	EstimatedUploadSizeB int64 `json:"estimatedUploadSizeB,omitempty"`
 	// EstimatedTimeSeconds is the estimated time required in seconds to upload the
 	// remaining data estimated with EstimatedUploadSizeB.
 	// This field will be empty for phases which do not involve data movement.
--- a/pkg/apis/cr/v1alpha1/types.go
+++ b/pkg/apis/cr/v1alpha1/types.go
@@ -155,6 +155,20 @@ type ActionProgress struct {
 	// PercentCompleted is computed by assessing the number of completed phases
 	// against the total number of phases.
 	PercentCompleted string `json:"percentCompleted,omitempty"`
+	// SizeDownloadedB represents the size of data downloaded in Bytes at a given time during action execution.
+	// This field will be empty for actions which do not involve data movement.
+	SizeDownloadedB int64 `json:"sizeDownloadedB,omitempty"`
+	// SizeUploadedB represents the size of data uploaded in Bytes at a given time during action execution.
+	// This field will be empty for actions which do not involve data movement.
+	SizeUploadedB int64 `json:"sizeUploadedB,omitempty"`
+	// EstimatedDownloadSizeB represents the total estimated size of data in Bytes
+	// that will be downloaded during the action execution.
+	// This field will be empty for actions which do not involve data movement.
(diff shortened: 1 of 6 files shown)


pkg/customresource/actionset.yaml
```diff
@@ -325,6 +341,22 @@ spec:
           type: string
           description: Progress of completion in percentage
           jsonPath: .status.progress.percentCompleted
+        - name: SizeDownloadedB
+          type: integer
+          description: Amount of data downloaded during action execution
```
viveksinghggits, line 346:
```suggestion
          description: Amount of data that has been downloaded
```
what do you think about this?

viveksinghggits: if you agree, we can change other columns as well.

pkg/progress/action.go
```diff
@@ -144,6 +144,26 @@ func completedOrFailed(aIDX int, actionSet *crv1alpha1.ActionSet, phaseName stri
 	return false
 }
 
+func isPhaseProgressDiffer(a, b crv1alpha1.PhaseProgress) bool {
```
viveksinghggits, line 147: I am not able to figure out what exactly this function is, by reading the name. is it trying to figure out if phase progress is different?

viveksinghggits (commented):
LGTM

pkg/progress/action.go
e-sumin: `a` and `b` both are phase progress, the function returns true if they are different, and false if they are similar.

e-sumin: I can't think of a better name, do you have any ideas?

pkg/apis/cr/v1alpha1/types.go
```diff
@@ -196,15 +210,22 @@ type Phase struct {
...
+	EstimatedDownloadSizeB int64 `json:"estimatedDownloadSizeB,omitempty"`
 	// EstimatedUploadSizeB represents the total estimated size of data in Bytes
 	// that will be uploaded during the phase execution.
 	// This field will be empty for phases which do not involve data movement.
-	EstimatedUploadSizeB int64 `json:"estinatedUploadSizeB,omitempty"`
-	// EstimatedTimeSeconds is the estimated time required in seconds to upload the
-	// remaining data estimated with EstimatedUploadSizeB.
+	EstimatedUploadSizeB int64 `json:"estimatedUploadSizeB,omitempty"`
+	// EstimatedTimeSeconds is the estimated time required in seconds to transfer the
+	// remaining data estimated with EstimatedUploadSizeB / EstimatedDownloadSizeB.
```
viveksinghggits, line 228:
```suggestion
	// remaining data estimated with EstimatedUploadSizeB/EstimatedDownloadSizeB.
```

pkg/progress/action.go
viveksinghggits: maybe having different instead of differ would be a good idea? Or if we want to shorten we can just use `Diff` instead of differ.

e-sumin:
All notes were accepted.

viveksinghggits (approved):

pkg/progress/action.go
```diff
@@ -144,6 +144,26 @@ func completedOrFailed(aIDX int, actionSet *crv1alpha1.ActionSet, phaseName stri
 	return false
 }
 
+func isPhaseProgressDifferent(a, b crv1alpha1.PhaseProgress) bool {
```
viveksinghggits, line 147: just one small nit please feel free to ignore it. because we have two PhaseProgresses here, should we rename this to 
`arePhaseProgressesDifferent`?

e-sumin: I thought about this yesterday but my conclusion was - probably it's not so important. But since you have the same thoughts, probably that makes sense. Renamed.

Setyania (commented):
Location

PR #2941, opened 2024-06-14
fix: Make pod writer exec wait for cat command to finish

hairyhum:
## Change Overview

PodWriteSuite.TestPodWriter fails with error because written file is empty while it's being written by the pod writer cat command.

Pod writer Write returns before the file is written because it doesn't wait for cat command to complete.

The change makes it wait for cat command and removes the race condition.

Also bump timeout in exec_test.go to fix another flake

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E


hairyhum:
Maybe there's a better way to make sure `cat` finishes, I'm open to ideas.

e-sumin (approved):
LGTM
I don't have any other ideas how we can ensure that `Write` is done.

pkg/kube/pod_writer.go
```diff
@@ -53,7 +53,7 @@ func NewPodWriter(cli kubernetes.Interface, path string, content io.Reader) PodW
 
 // Write will create a new file(if not present) and write the provided content to the file
 func (p *podWriter) Write(ctx context.Context, namespace, podName, containerName string) error {
-	cmd := []string{"sh", "-c", "cat - > " + p.path}
+	cmd := []string{"sh", "-c", "cat - > " + p.path + " && :"}
```
viveksinghggits, line 56: can this result into a problem, if someone wrote something and immediately after it returns they tried to check the content?

viveksinghggits:
> Maybe there's a better way to make sure `cat` finishes, I'm open to ideas.

do we really have to run `cat` in background?

pkg/kube/pod_writer.go
hairyhum: That's the issue currently. In `pod_writer_test.go` we have a flake because we execute `Write` and then check the content right away, which gets us an empty file. Even worse that adding sleep in the test didn't fix it sometimes, suggesting that exec finished before it could process the input data (that's how I understood it).
My understanding is with the fix `cat` waits for the input and `Write` only returns when the data is written to the file.

hairyhum:
> do we really have to run cat in background?

I don't think we run it in background in any case, the fix adds `&&`, not `&`.

viveksinghggits:
> > do we really have to run cat in background?
> 
> I don't think we run it in background in any case, the fix adds `&&`, not `&`.

Got it. Thank you Daniil, I misread.

PR #2959, opened 2024-06-25
refactor: Pass parallelism flag to snapshot restore command

e-sumin:
## Change Overview

Currently when building snapshot restore command we are not using parallelism flag, and default value is used. This PR brings control over this parameter.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


hairyhum (approved):
LGTM. But I wonder if we can start moving to a new CLI tooling https://github.com/kanisterio/kanister/tree/master/pkg/kopia/cli
It's probably more work, but we have to do that eventually.

hairyhum:
Sorry @e-sumin I intended to leave a comment, please dismiss the approval.

e-sumin:
> LGTM. But I wonder if we can start moving to a new CLI tooling 
I believe we can, but not in this PR. I think it's better to do that separately.

helm/kanister-operator/templates/deployment.yaml
```diff
@@ -41,6 +41,8 @@ spec:
           value: {{ .Values.controller.logLevel }}
         - name: DATA_STORE_PARALLEL_UPLOAD
           value: {{ .Values.controller.parallelism | quote }}
+        - name: DATA_STORE_PARALLEL_DOWNLOAD
+          value: {{ .Values.controller.parallelism | quote }}
```
PrasadG193, line 45: We are referring to the same flag `controller.parallelism` for upload and download parallelism. Is it expected? Or should we have separate flags for both?

pkg/function/restore_data_using_kopia_server.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/kanisterio/kanister/pkg/utils"
```
viveksinghggits, line 23: This doesn't seem to be grouped together correctly.

pkg/kopia/const.go
```diff
@@ -26,6 +26,11 @@ const (
 	DataStoreParallelUploadName = "DATA_STORE_PARALLEL_UPLOAD"
 	// DefaultDataStoreParallelUpload is the Default Value of Parallelism
 	DefaultDataStoreParallelUpload = 8
+	// DataStoreParallelDownloadName is the Environmental Variable set in Kanister
+	// For Parallelism to be used by Kopia for restore action
```
viveksinghggits, line 30:
```suggestion
	// for Parallelism to be used by Kopia for restore action
```

helm/kanister-operator/templates/deployment.yaml
e-sumin: I've intentionally took it from the same value, cause this is very specific settings. But since you both think that it's good to have them separated, not a big deal. I can introduce new value.

The question is, how new value should be named ? I'd go for `.Values.controller.parallelDownloads` but in such case the next question will follow up - what to do with old one. Should it be renamed to `parallelUploads` ? If yes - should we notify users about that change (parallelism is not mentioned in documentation, and as far as I understand it's undocumented feature) ?

PrasadG193: I was thinking if it makes sense to have separate section for datastore related config? Having it at controller level seems bit confusing. Controller also perform other operations like actionset action executions. 
Maybe something like 
```
dataStore:
  parallelism:
    upload:
    download:
```

PrasadG193: > If yes - should we notify users about that change (parallelism is not mentioned in documentation, and as far as I understand it's undocumented feature)

This feature is not usable right now. It doesn't work with Kanister without Kopia server setup. Which we haven't documented yet. So we have the privilege to change it without announcing it.

e-sumin: Thank you ! Great idea, done.

PrasadG193 (approved):
LGTM

PR #2986, opened 2024-07-17
Support annotation in VolumeSnapshot and VolumeSnapshotContent resources

saima-s:
## Change Overview

This PR is about adding support for annotation in some methods of Snapshotter interface. Methods are listed below:
1. Create
2. Clone
3. CreateFromSource
4. CreateContentFromSource

## Pull request type

- [x] :sunflower: Feature

## Issues 

- fixes #issue-number

#### Test case-1
Created snapshot and restored those snapshots in another namespace.

#### Result
Result can be found [here](https://gist.github.com/saima-s/74aac90cba31ab365ab765000d358597)

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kube/snapshot/snapshot.go
```diff
@@ -62,7 +62,8 @@ type Snapshotter interface {
 	// 'waitForReady' will block the caller until the snapshot status is 'ReadyToUse'.
 	// or 'ctx.Done()' is signalled. Otherwise it will return immediately after the snapshot is cut.
 	// 'labels' can also be addded to the volume snapshot.
-	Create(ctx context.Context, name, namespace, pvcName string, snapshotClass *string, waitForReady bool, labels map[string]string) error
+	// 'annotations' can also be addded to the volume snapshot.
+	Create(ctx context.Context, name, namespace, pvcName string, snapshotClass *string, waitForReady bool, labels map[string]string, annotations map[string]string) error
```
viveksinghggits, line 66: I was thinking if we should have a type called `SnapshotMeta` that should be passed to this method.
```
// has fields that can be set for created volume snapshot
type SnapshotMeta struct{
Name string
Namespace string
Labels map[string]string
Annotations map[string]string
}
```

viveksinghggits: let's see what @PrasadG193 thinks.

pkg/kube/snapshot/snapshot.go
```diff
@@ -88,7 +89,8 @@ type Snapshotter interface {
 	// 'cloneNamespace' is the namespace where the clone will be created.
 	// 'waitForReady' will make the function blocks until the clone's status is ready to use.
 	// 'labels' is the labels to set on the created VSC
-	Clone(ctx context.Context, name, namespace, cloneName, cloneNamespace string, waitForReady bool, labels map[string]string) error
+	// 'annotations' is the annotation to set on the created VS and VSC
+	Clone(ctx context.Context, name, namespace, cloneName, cloneNamespace string, waitForReady bool, labels map[string]string, annotations map[string]interface{}) error
```
viveksinghggits, line 93: Same comment here about creating a new type and then passing that around.

pkg/kube/snapshot/snapshot.go
```diff
@@ -101,15 +103,17 @@ type Snapshotter interface {
 	// 'namespace' is the namespace of the snapshot.
 	// 'waitForReady' blocks the caller until snapshot is ready to use or context is cancelled.
 	// 'labels' is the labels to set on the created VSC
-	CreateFromSource(ctx context.Context, source *Source, snapshotName, namespace string, waitForReady bool, labels map[string]string) error
+	// 'annotations' is the annotation to set on the created VS and VSC
+	CreateFromSource(ctx context.Context, source *Source, snapshotName, namespace string, waitForReady bool, labels map[string]string, annotations map[string]interface{}) error
```
viveksinghggits, line 107: here as well.

pkg/kube/snapshot/snapshot.go
```diff
@@ -101,15 +103,17 @@ type Snapshotter interface {
...
 	// CreateContentFromSource will create a 'VolumesnaphotContent' for the underlying snapshot source.
 	//
 	// 'source' contains information about CSI snapshot.
 	// 'contentName' is the name of the VSC that will be created
 	// 'snapshotName' is the name of the snapshot that will be reference the VSC
 	// 'namespace' is the namespace of the snapshot.
 	// 'deletionPolicy' is the deletion policy to set on the created VSC
-	CreateContentFromSource(ctx context.Context, source *Source, contentName, snapshotName, namespace, deletionPolicy string) error
+	// 'annotations' is the annotation to set on the created VSC
+	CreateContentFromSource(ctx context.Context, source *Source, contentName, snapshotName, namespace, deletionPolicy string, annotations map[string]string) error
```
viveksinghggits, line 116: and here too.

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -259,14 +261,16 @@ func (sna *SnapshotAlpha) GetSource(ctx context.Context, snapshotName, namespace
...
 	deletionPolicy, err := sna.getDeletionPolicyFromClass(source.VolumeSnapshotClassName)
 	if err != nil {
 		return errkit.Wrap(err, "Failed to get DeletionPolicy from VolumeSnapshotClass")
 	}
 	contentName := snapshotName + "-content-" + string(uuid.NewUUID())
-	snap := UnstructuredVolumeSnapshotAlpha(snapshotName, namespace, "", contentName, source.VolumeSnapshotClassName, blockstorage.SanitizeTags(labels))
-	if err := sna.CreateContentFromSource(ctx, source, contentName, snapshotName, namespace, deletionPolicy); err != nil {
+	snapshotAnnotation := getAnnotation(annotations, SnapshotAnnotation)
+	snap := UnstructuredVolumeSnapshotAlpha(snapshotName, namespace, "", contentName, source.VolumeSnapshotClassName, blockstorage.SanitizeTags(labels), snapshotAnnotation)
+	contentAnnotation := getAnnotation(annotations, SnapshotContentAnnotation)
```
viveksinghggits, line 272: are we getting volumesnapshot annotations and the volumesnapshotcontent annotations from the same annotations map that is passed to this method?
how are we differentiating between the two?
I think it will become confusing for the callers.

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -278,6 +282,14 @@ func (sna *SnapshotAlpha) CreateFromSource(ctx context.Context, source *Source,
 	return sna.WaitOnReadyToUse(ctx, snapshotName, namespace)
 }
 
+func getAnnotation(annotations map[string]interface{}, key string) map[string]string {
+	if val, ok := annotations[key]; ok {
+		return val.(map[string]string)
+	}
+
+	return nil
+}
```
viveksinghggits, line 291: This is confusing for me, are annotations `,` separated? Because we are getting annotation value for either 
```
SnapshotContentAnnotation         = "snapshotcontent"
	SnapshotAnnotation                = "snapshot"
```
and then just returning the value? 🤔

pkg/kube/snapshot/snapshot_alpha.go
saima-s: Yes, here we are creating map[string]interface{} in the case where string can be `snapshot` and `snapshotcontent` and value is interface which can again be map[string]string for each snapshot and snapshot content annotation.
E.g: 
```
testAnnotation := map[string]interface{}{
		SnapshotAnnotation : map[string]string{
			"testannotation" : "value",
			"testannotation1": "value1",
		},
		SnapshotContentAnnotation : map[string]string{
			"testannotation2" : "value2",
			"testannotation3": "value3",
		},
	}
```

pkg/kube/snapshot/snapshot_alpha.go
saima-s: Since the annotation will be in below format:
```
testAnnotation := map[string]interface{}{
		SnapshotAnnotation : map[string]string{
			"testannotation" : "value",
			"testannotation1": "value1",
		},
		SnapshotContentAnnotation : map[string]string{
			"testannotation2" : "value2",
			"testannotation3": "value3",
		},
	}
```
`getAnnotation` is function to get interface value on the basis of key `SnapshotAnnotation` or `SnapshotContentAnnotation`.

pkg/function/create_csi_snapshot.go
```diff
@@ -143,7 +143,7 @@ func (*createCSISnapshotFunc) Arguments() []string {
 }
 
 func createCSISnapshot(ctx context.Context, snapshotter snapshot.Snapshotter, name, namespace, pvc, snapshotClass string, wait bool, labels map[string]string) (*v1.VolumeSnapshot, error) {
-	if err := snapshotter.Create(ctx, name, namespace, pvc, &snapshotClass, wait, labels); err != nil {
+	if err := snapshotter.Create(ctx, name, namespace, pvc, &snapshotClass, wait, labels, nil); err != nil {
```
viveksinghggits, line 146: Instead of passing `nil` should we pass empty annotation `map[string]string{}`.

saima-s: Passed `nil` by getting the idea from how labels are implemented as below:
```
if labels != nil {
		snap.SetLabels(labels)
	}
	if annotation != nil {
		snap.SetAnnotations(annotation)
	}
```

pkg/function/create_csi_snapshot.go
```diff
@@ -143,7 +143,12 @@ func (*createCSISnapshotFunc) Arguments() []string {
 }
 
 func createCSISnapshot(ctx context.Context, snapshotter snapshot.Snapshotter, name, namespace, pvc, snapshotClass string, wait bool, labels map[string]string) (*v1.VolumeSnapshot, error) {
-	if err := snapshotter.Create(ctx, name, namespace, pvc, &snapshotClass, wait, labels, nil); err != nil {
+	if err := snapshotter.Create(ctx, pvc, &snapshotClass, wait, metav1.ObjectMeta{
+		Name:        name,
+		Namespace:   namespace,
+		Labels:      labels,
+		Annotations: nil,
+	}); err != nil {
```
viveksinghggits, line 151: Let's create the `ObjectMeta` separately and then pass that to `Create`.

viveksinghggits: Instead of creating that inline.

pkg/function/create_csi_snapshot_static.go
```diff
@@ -162,7 +162,7 @@ func createCSISnapshotStatic(
 		Driver:                  driver,
 		VolumeSnapshotClassName: snapshotClass,
 	}
-	if err := snapshotter.CreateFromSource(ctx, source, name, namespace, wait, nil, nil); err != nil {
+	if err := snapshotter.CreateFromSource(ctx, source, wait, metav1.ObjectMeta{Name: name, Namespace: namespace}, metav1.ObjectMeta{}); err != nil {
```
viveksinghggits, line 165: I think we should create `ObjectMeta` before this line, like requested above.

pkg/function/delete_csi_snapshot_content_test.go
```diff
@@ -83,10 +83,13 @@ func (testSuite *DeleteCSISnapshotContentTestSuite) TestDeleteCSISnapshotContent
...
-			snapshotName,
-			snapshotNamespace,
-			deletionPolicy, nil)
+			deletionPolicy, metav1.ObjectMeta{
+				Name:      snapshotName,
+				Namespace: snapshotNamespace,
+			},
+			metav1.ObjectMeta{
+				Name: snapshotContentName,
+			})
```
viveksinghggits, line 92: same here, let's create `snapshotMeta` and `contentMeta` before this line and pass that. It would make sure that reader understands what exactly is being passed.

pkg/function/delete_csi_snapshot_test.go
```diff
@@ -105,7 +105,11 @@ func (testSuite *DeleteCSISnapshotTestSuite) TestDeleteCSISnapshot(c *C) {
 		_, err = fakeCli.CoreV1().PersistentVolumeClaims(testSuite.namespace).Create(ctx, originalPVC, metav1.CreateOptions{})
 		c.Assert(err, IsNil)
 
-		err = fakeSnapshotter.Create(ctx, testSuite.snapName, testSuite.namespace, testSuite.pvcName, &testSuite.volumeSnapshotClass, false, nil, nil)
+		err = fakeSnapshotter.Create(ctx, testSuite.pvcName, &testSuite.volumeSnapshotClass, false,
+			metav1.ObjectMeta{
+				Name:      testSuite.snapName,
+				Namespace: testSuite.namespace,
+			})
```
viveksinghggits, line 112:
```suggestion
                 snapshotMeta := metav1.ObjectMeta{
				Name:      testSuite.snapName,
				Namespace: testSuite.namespace,
			}
		err = fakeSnapshotter.Create(ctx, testSuite.pvcName, &testSuite.volumeSnapshotClass, false,
			snapshotMeta)
```

Let's do this, like discussed above.

pkg/function/restore_csi_snapshot_test.go
```diff
@@ -95,7 +95,7 @@ func (testSuite *RestoreCSISnapshotTestSuite) TestRestoreCSISnapshot(c *C) {
 		originalPVC := getOriginalPVCManifest(testSuite.pvcName, testSuite.storageClass)
 		createPVC(c, testSuite.namespace, originalPVC, fakeCli)
 
-		err = fakeSnapshotter.Create(ctx, testSuite.snapName, testSuite.namespace, testSuite.pvcName, &testSuite.volumeSnapshotClass, false, nil, nil)
+		err = fakeSnapshotter.Create(ctx, testSuite.pvcName, &testSuite.volumeSnapshotClass, false, metav1.ObjectMeta{Name: testSuite.snapName, Namespace: testSuite.namespace})
```
viveksinghggits, line 98: Here as well, let's create the snapshotMeta first and then pass that to this method call.

viveksinghggits (changes requested):

pkg/kube/snapshot/snapshot.go
PrasadG193: Sound good, we can use [meta/v1.ObjectMeta](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) as well

pkg/kube/snapshot/mocks/mock_snapshotter.go
```diff
@@ -1,8 +1,8 @@
 // Code generated by MockGen. DO NOT EDIT.
-// Source: github.com/kanisterio/kanister/pkg/kube/snapshot (interfaces: Snapshotter)
+// Source: pkg/kube/snapshot/snapshot.go
 
-// Package mocks is a generated GoMock package.
-package mocks
+// Package mock_snapshot is a generated GoMock package.
+package mock_snapshot
```
PrasadG193, line 5: How is it generated? Can we keep the package name same?

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -46,6 +46,8 @@ const (
 	DeletionPolicyDelete              = "Delete"
 	DeletionPolicyRetain              = "Retain"
 	CloneVolumeSnapshotClassLabelName = "kanister-cloned-from"
+	SnapshotContentAnnotation         = "snapshotcontent"
+	SnapshotAnnotation                = "snapshot"
```
PrasadG193, line 50: I don't see its used anywhere

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -365,14 +357,14 @@ func (sna *SnapshotAlpha) getDeletionPolicyFromClass(snapClassName string) (stri
 	return vsc.DeletionPolicy, nil
 }
 
-func UnstructuredVolumeSnapshotAlpha(name, namespace, pvcName, contentName, snapClassName string, labels, annotation map[string]string) *unstructured.Unstructured {
+func UnstructuredVolumeSnapshotAlpha(pvcName, snapClassName string, snapshotMeta, contentMeta metav1.ObjectMeta) *unstructured.Unstructured {
```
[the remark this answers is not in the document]
PrasadG193: Same for other functions where meta is passed for different resources

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -402,14 +409,18 @@ func UnstructuredVolumeSnapshotContentAlpha(name, snapshotName, snapshotNs, dele
...
-					"namespace": snapshotNs,
+					"name":      snapshotMeta.Name,
+					"namespace": snapshotMeta.Namespace,
 				},
 				"snapshotClassName": snapClassName,
 				"deletionPolicy":    deletionPolicy,
 			},
 		},
 	}
+	if contentMeta.Annotations != nil {
```
PrasadG193, line 420: Should we also set labels?

pkg/kube/snapshot/snapshot.go
PrasadG193: @viveksinghggits @saima-s I was giving second thought on using `metav1.ObjectMeta`. It has a lot of other fields that we don't need. It might confuse consumer e.g they might set `OwnerReferences` in objectMeta but since we don't set it, it might lead to confusion. I think its better to declare a struct with only fields we allow to be set.

pkg/kube/snapshot/snapshot_alpha.go
saima-s: Initially labels was not set, that's why didn't added as apart of this PR. We can take it up on another PR.

pkg/kube/snapshot/snapshot.go
saima-s: Yes, I agree. But we can also inform about what field is to set or not in function definition as:
```
// snapshotMeta object has 'name' as name of the VolumeSnapshot.
// snapshotMeta object has 'namespace' as namespace of the PVC. VolumeSnapshot will be crated in the same namespace.
// snapshotMeta object has 'labels' that can also be addded to the volume snapshot.
// snapshotMeta object has 'annotations' that can also be addded to the volume snapshot.

```

viveksinghggits: Yes, I also think it's better to not use `metav1.ObjectMeta` and have a new type/s like we discussed before.

viveksinghggits: > Yes, I agree. But we can also inform about what field is to set or not in function definition as:

This just talks about the fields are the used, what about the other fields that are not used. We will have to mentioned that only these fields from `ObjectMeta` are used and other are ignored, which again, would confuse developers.

saima-s: Here, we can have 2 different struct for snapshot and snapshot content. Initially we will have similar set of fields as, later on we can separately include other fields if we want other to be configurable:
```
Name string
Namespace string
Labels map[string]string
Annotations map[string]string
```
Or we can have only 1 struct, as of now these 4 fields will suffice what we need. whdyt?

PrasadG193: I would prefer only 1 struct as I don't see any difference in the fields

pkg/kube/snapshot/mocks/mock_snapshotter.go
```diff
@@ -1,18 +1,17 @@
 // Code generated by MockGen. DO NOT EDIT.
-// Source: github.com/kanisterio/kanister/pkg/kube/snapshot (interfaces: Snapshotter)
+// Source: snapshot.go
 
-// Package mocks is a generated GoMock package.
+// Package mock_snapshot is a generated GoMock package.
```
PrasadG193, line 4: Please correct the package name, also the the `Source:` changes seems to be incorrect/unnecessary

pkg/kube/snapshot/types.go
```diff
@@ -0,0 +1,35 @@
+package snapshot
+
+// SnapshotMeta is metadata of snapshot which is configurable.
+type SnapshotMeta struct {
```
PrasadG193, line 4: I would prefer `ObjectMeta`, so that on import it becomes `snapshot.ObjectMeta`.

pkg/kube/snapshot/types.go
```diff
@@ -0,0 +1,35 @@
...
+	// Name must be unique within a namespace. Is required when creating resources, although
+	// some resources may allow a client to request the generation of an appropriate name
+	// automatically. Name is primarily intended for creation idempotence and configuration
+	// definition.
+	// Cannot be updated.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names#names
+	// +optional
+	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
```
PrasadG193, line 12: @saima-s looks like you have copies the comments from `meta/v1.ObjectMeta` struct. We don't have to elaborate this much here. Let's add simple and short comments on fields of the struct. Also, no need to protobuf tags, they are irrelevant in our case.

saima-s: ok, sure.

pkg/kube/snapshot/mocks/mock_snapshotter.go
saima-s: This was generated using `mockgen`. Updated the package name.

pkg/kube/snapshot/snapshot.go
```diff
@@ -84,32 +82,30 @@ type Snapshotter interface {
...
+	// 'snapshotMeta' has metadata of the Volumesnaphot resource that is going to get created.
+	// 'contentMeta' has metadata of the VolumesnapshotContent content resource that is going to get created.
+	Clone(ctx context.Context, name, namespace string, waitForReady bool, snapshotMeta, contentMeta ObjectMeta) error
 	// GetSource will return the CSI source that backs the volume snapshot.
 	//
 	// 'snapshotName' is the name of the Volumesnapshot.
 	// 'namespace' is the namespace of the Volumesnapshot.
 	GetSource(ctx context.Context, snapshotName, namespace string) (*Source, error)
-	// CreateFromSource will create a 'Volumesnapshot' and 'VolumesnaphotContent' pair for the underlying snapshot source.
+	// CreateFromSource will create a 'Volumesnapshot' and 'VolumesnapshotContent' pair for the underlying snapshot source.
```
PrasadG193, line 94:
```suggestion
	// CreateFromSource will create a 'VolumeSnapshot' and 'VolumeSnapshotContent' pair for the underlying snapshot source.
```
[nit]

pkg/kube/snapshot/snapshot.go
```diff
@@ -84,32 +82,30 @@ type Snapshotter interface {
...
 	//
 	// 'source' contains information about CSI snapshot.
-	// 'snapshotName' is the name of the snapshot that will be created.
-	// 'namespace' is the namespace of the snapshot.
 	// 'waitForReady' blocks the caller until snapshot is ready to use or context is cancelled.
-	// 'labels' is the labels to set on the created VSC
-	CreateFromSource(ctx context.Context, source *Source, snapshotName, namespace string, waitForReady bool, labels map[string]string) error
-	// CreateContentFromSource will create a 'VolumesnaphotContent' for the underlying snapshot source.
+	// 'annotations' is the annotation to set on the created VS and VSC
```
PrasadG193, line 98:
```suggestion
```

pkg/kube/snapshot/snapshot.go
```diff
@@ -84,32 +82,30 @@ type Snapshotter interface {
...
 	// 'name' is the name of the VolumeSnapshot that will be cloned.
 	// 'namespace' is the namespace of the VolumeSnapshot that will be cloned.
-	// 'cloneName' is name of the clone.
-	// 'cloneNamespace' is the namespace where the clone will be created.
 	// 'waitForReady' will make the function blocks until the clone's status is ready to use.
-	// 'labels' is the labels to set on the created VSC
-	Clone(ctx context.Context, name, namespace, cloneName, cloneNamespace string, waitForReady bool, labels map[string]string) error
+	// 'snapshotMeta' has metadata of the Volumesnaphot resource that is going to get created.
+	// 'contentMeta' has metadata of the VolumesnapshotContent content resource that is going to get created.
+	Clone(ctx context.Context, name, namespace string, waitForReady bool, snapshotMeta, contentMeta ObjectMeta) error
```
PrasadG193, line 88:
```suggestion
	// 'cloneSnapshotMeta' has metadata of the Volumesnaphot resource that is going to get created.
	// 'contentMeta' has metadata of the VolumesnapshotContent content resource that is going to get created.
	Clone(ctx context.Context, name, namespace string, waitForReady bool, cloneSnapshotMeta, snapshotContentMeta ObjectMeta) error
```
Let's rename `contentMeta` -> `snapshotContentMeta` in all the occurrences

pkg/kube/snapshot/snapshot.go
```diff
@@ -84,32 +82,30 @@ type Snapshotter interface {
...
 	// 'source' contains information about CSI snapshot.
-	// 'snapshotName' is the name of the snapshot that will be created.
-	// 'namespace' is the namespace of the snapshot.
 	// 'waitForReady' blocks the caller until snapshot is ready to use or context is cancelled.
-	// 'labels' is the labels to set on the created VSC
-	CreateFromSource(ctx context.Context, source *Source, snapshotName, namespace string, waitForReady bool, labels map[string]string) error
-	// CreateContentFromSource will create a 'VolumesnaphotContent' for the underlying snapshot source.
+	// 'annotations' is the annotation to set on the created VS and VSC
+	// 'snapshotMeta' has metadata of the Volumesnapshot resource that is going to get created.
+	// 'contentMeta' has metadata of the VolumesnapshotContent resource that is going to get created.
```
PrasadG193, line 100:
```suggestion
	// 'snapshotMeta' has metadata of the VolumeSnapshot resource that is going to get created.
	// 'contentMeta' has metadata of the VolumeSnapshotContent resource that is going to get created.
```
Lets fix this for all the occurrences

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -216,20 +217,20 @@ func (sna *SnapshotAlpha) DeleteContent(ctx context.Context, name string) error
 
 // Clone will clone the VolumeSnapshot to namespace 'cloneNamespace'.
 // Underlying VolumeSnapshotContent will be cloned with a different name.
-func (sna *SnapshotAlpha) Clone(ctx context.Context, name, namespace, cloneName, cloneNamespace string, waitForReady bool, labels map[string]string) error {
-	_, err := sna.Get(ctx, cloneName, cloneNamespace)
+func (sna *SnapshotAlpha) Clone(ctx context.Context, name, namespace string, waitForReady bool, snapshotMeta, contentMeta ObjectMeta) error {
```
PrasadG193, line 220:
```suggestion
func (sna *SnapshotAlpha) Clone(ctx context.Context, name, namespace string, waitForReady bool, cloneSnapshotMeta, snapshotContentMeta ObjectMeta) error {
```

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -353,14 +355,16 @@ func (sna *SnapshotAlpha) getDeletionPolicyFromClass(snapClassName string) (stri
 	return vsc.DeletionPolicy, nil
 }
 
-func UnstructuredVolumeSnapshotAlpha(name, namespace, pvcName, contentName, snapClassName string, labels map[string]string) *unstructured.Unstructured {
+// UnstructuredVolumeSnapshotAlpha has contentMeta has metadata of VolumeSnapshotContent.
+// If contentMeta has name value set, UnstructuredVolumeSnapshotAlpha will create both VolumeSnapshot and VolumeSnapshotContent resource.
```
PrasadG193, line 359:
```suggestion
// UnstructuredVolumeSnapshotAlpha creates Unstructured object for the VolumeSnapshot resource.
```
Comment needs correction, `UnstructuredVolumeSnapshotAlpha` doesn't create any object, it just returns unstructured  object for the given resource

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -369,31 +373,35 @@ func UnstructuredVolumeSnapshotAlpha(name, namespace, pvcName, contentName, snap
...
+		snap.SetAnnotations(snapshotMeta.Annotations)
 	}
 	return snap
 }
 
-func UnstructuredVolumeSnapshotContentAlpha(name, snapshotName, snapshotNs, deletionPolicy, driver, handle, snapClassName string) *unstructured.Unstructured {
-	return &unstructured.Unstructured{
+// UnstructuredVolumeSnapshotContentAlpha has snapshotMeta param which is used as volumeSnapshotRef when VolumeSnapshotContent is created.
+func UnstructuredVolumeSnapshotContentAlpha(deletionPolicy, driver, handle, snapClassName string, snapshotMeta, contentMeta ObjectMeta) *unstructured.Unstructured {
+	snaphotContent := unstructured.Unstructured{
```
PrasadG193, line 399:
```suggestion
// UnstructuredVolumeSnapshotContentAlpha returns Unstructured object for the VolumeSnapshot resource.  
func UnstructuredVolumeSnapshotContentAlpha(deletionPolicy, driver, handle, snapClassName string, volSnapshotRef, snapshotContentMeta ObjectMeta) *unstructured.Unstructured {
	snaphotContent := unstructured.Unstructured{
```

pkg/kube/snapshot/snapshot_alpha.go
PrasadG193: Same for beta implementation of the function

pkg/kube/snapshot/snapshot_alpha.go
PrasadG193: Same for beta implementation of the function

pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -343,14 +341,16 @@ func getDeletionPolicyFromClass(dynCli dynamic.Interface, snapClassGVR schema.Gr
 	return vsc.DeletionPolicy, nil
 }
 
-func UnstructuredVolumeSnapshot(gvr schema.GroupVersionResource, name, namespace, pvcName, contentName, snapClassName string, labels map[string]string) *unstructured.Unstructured {
+// UnstructuredVolumeSnapshot has contentMeta has metadata of VolumeSnapshotContent.
+// If contentMeta has name value set, UnstructuredVolumeSnapshot will create both VolumeSnapshot and VolumeSnapshotContent resource.
+func UnstructuredVolumeSnapshot(gvr schema.GroupVersionResource, pvcName, snapClassName string, snapShotMeta, contentMeta ObjectMeta) *unstructured.Unstructured {
```
PrasadG193, line 346: Same changes required as suggested on `UnstructuredVolumeSnapshotAlpha`

pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -362,33 +362,37 @@ func UnstructuredVolumeSnapshot(gvr schema.GroupVersionResource, name, namespace
...
+		snap.SetAnnotations(snapShotMeta.Annotations)
 	}
 	return snap
 }
 
-func UnstructuredVolumeSnapshotContent(gvr schema.GroupVersionResource, name, snapshotName, snapshotNs, deletionPolicy, driver, handle, snapClassName string) *unstructured.Unstructured {
-	return &unstructured.Unstructured{
+// UnstructuredVolumeSnapshotContent has snapshotMeta param which is used as volumeSnapshotRef when VolumeSnapshotContent is created.
+func UnstructuredVolumeSnapshotContent(gvr schema.GroupVersionResource, deletionPolicy, driver, handle, snapClassName string, snapshotMeta, contentMeta ObjectMeta) *unstructured.Unstructured {
+	snapshotContent := unstructured.Unstructured{
```
PrasadG193, line 384: Same feedback here

PrasadG193 (commented):
Need to correct few code comment, overall looks in good shape now

pkg/kube/snapshot/snapshot_alpha.go
saima-s: done

pkg/function/create_csi_snapshot_static.go
```diff
@@ -162,7 +162,11 @@ func createCSISnapshotStatic(
 		Driver:                  driver,
 		VolumeSnapshotClassName: snapshotClass,
 	}
-	if err := snapshotter.CreateFromSource(ctx, source, name, namespace, wait, nil); err != nil {
+	snapshotMeta := snapshot.ObjectMeta{
+		Name:      name,
+		Namespace: namespace,
+	}
+	if err := snapshotter.CreateFromSource(ctx, source, wait, snapshotMeta, snapshot.ObjectMeta{}); err != nil {
```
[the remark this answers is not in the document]
saima-s: yes we can do this.

pkg/kube/snapshot/mocks/mock_snapshotter.go
```diff
@@ -39,188 +39,188 @@ func (m *MockSnapshotter) EXPECT() *MockSnapshotterMockRecorder {
 }
 
 // Clone mocks base method.
-func (m *MockSnapshotter) Clone(arg0 context.Context, arg1, arg2, arg3, arg4 string, arg5 bool, arg6 map[string]string) error {
+func (m *MockSnapshotter) Clone(ctx context.Context, name, namespace string, waitForReady bool, cloneSnapshotMeta, snapshotContentMeta snapshot.ObjectMeta) error {
```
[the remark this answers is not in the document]
saima-s: This change came from the generated code only. After the change made to interface, generated this using `mockgen`.

pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -191,20 +189,20 @@ func (sna *SnapshotBeta) DeleteContent(ctx context.Context, name string) error {
 
 // Clone will clone the VolumeSnapshot to namespace 'cloneNamespace'.
 // Underlying VolumeSnapshotContent will be cloned with a different name.
-func (sna *SnapshotBeta) Clone(ctx context.Context, name, namespace, cloneName, cloneNamespace string, waitForReady bool, labels map[string]string) error {
-	_, err := sna.Get(ctx, cloneName, cloneNamespace)
+func (sna *SnapshotBeta) Clone(ctx context.Context, name, namespace string, waitForReady bool, cloneSnapshotMeta, snapshotContentMeta ObjectMeta) error {
```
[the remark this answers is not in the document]
saima-s: @PrasadG193 wdyt?

pkg/kube/snapshot/snapshot.go
```diff
@@ -84,36 +82,33 @@ type Snapshotter interface {
...
 	// 'source' contains information about CSI snapshot.
-	// 'contentName' is the name of the VSC that will be created
-	// 'snapshotName' is the name of the snapshot that will be reference the VSC
-	// 'namespace' is the namespace of the snapshot.
 	// 'deletionPolicy' is the deletion policy to set on the created VSC
-	CreateContentFromSource(ctx context.Context, source *Source, contentName, snapshotName, namespace, deletionPolicy string) error
-	// WaitOnReadyToUse will block until the Volumesnapshot in namespace 'namespace' with name 'snapshotName'
+	// 'snapshotMeta' has metadata of the VolumeSnapshot resource that is source reference for VSC created.
+	// 'snapshotContentMeta' has metadata of the VolumeSnapshotContent resource that is going to get created.
+	CreateContentFromSource(ctx context.Context, source *Source, deletionPolicy string, snapshotMeta, snapshotContentMeta ObjectMeta) error
```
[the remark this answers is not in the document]
saima-s: agreed. Made changes.

pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -77,43 +77,41 @@ func (sna *SnapshotBeta) GetVolumeSnapshotClass(ctx context.Context, annotationK
...
 		}
-		return errkit.Wrap(err, "Failed to query PVC", "pvc", volumeName, "namespace", namespace)
+		return errkit.Wrap(err, "Failed to query PVC", "pvc", volumeName, "namespace", snapshotMeta.Namespace)
 	}
-
-	snap := UnstructuredVolumeSnapshot(snapGVR, name, namespace, volumeName, "", *snapshotClass, blockstorage.SanitizeTags(labels))
-	if _, err := dynCli.Resource(snapGVR).Namespace(namespace).Create(ctx, snap, metav1.CreateOptions{}); err != nil {
-		return errkit.Wrap(err, "Failed to create snapshot resource", "name", name, "namespace", namespace)
+	snapshotMeta.Labels = blockstorage.SanitizeTags(snapshotMeta.Labels)
+	snap := UnstructuredVolumeSnapshot(snapGVR, volumeName, *snapshotClass, snapshotMeta, ObjectMeta{})
```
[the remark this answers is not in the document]
saima-s: done

pkg/kube/snapshot/snapshot_beta.go
```diff
@@ -399,6 +403,10 @@ func UnstructuredVolumeSnapshotContent(gvr schema.GroupVersionResource, name, sn
 			},
 		},
 	}
+	if snapshotContentMeta.Annotations != nil {
+		snapshotContent.SetAnnotations(snapshotContentMeta.Annotations)
+	}
```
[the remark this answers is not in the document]
saima-s: I have added it.

pkg/kube/snapshot/snapshot.go
```diff
@@ -104,7 +104,7 @@ type Snapshotter interface {
 	// 'deletionPolicy' is the deletion policy to set on the created VSC
 	// 'snapshotMeta' has metadata of the VolumeSnapshot resource that is source reference for VSC created.
```
viveksinghggits, line 105: we are no longer accepting `snapshotMeta` as argument, is this comment still valid?

pkg/kube/snapshot/snapshot_alpha.go
```diff
@@ -395,7 +395,7 @@ func UnstructuredVolumeSnapshotAlpha(pvcName, snapClassName string, snapshotMeta
 }
 
 // UnstructuredVolumeSnapshotContentAlpha returns Unstructured object for the VolumeSnapshotContent resource.
-func UnstructuredVolumeSnapshotContentAlpha(deletionPolicy, driver, handle, snapClassName string, snapshotMeta, snapshotContentMeta ObjectMeta) *unstructured.Unstructured {
+func UnstructuredVolumeSnapshotContentAlpha(snapshotName, snapshotNs, deletionPolicy, driver, handle, snapClassName string, snapshotContentMeta ObjectMeta) *unstructured.Unstructured {
```
viveksinghggits, line 398:
```suggestion
func UnstructuredVolumeSnapshotContentAlpha(snapshotName, snapshotNS, deletionPolicy, driver, handle, snapClassName string, snapshotContentMeta ObjectMeta) *unstructured.Unstructured {
```

releasenotes/notes/support-annotation-on-snapshotter-function-ff9b7ba2daf10427.yaml
```diff
@@ -0,0 +1,2 @@
+---
+features: Support custom annotations for VolumeSnapshot and VolumeSnapshotContent resource when it gets created.
```
viveksinghggits, line 2: do we really need release note for this? I remember @hairyhum asked for it, but just wanted to confirm because this release note doesn't seem to be useful to the users.

viveksinghggits (commented):
Almost there.

pkg/kube/snapshot/snapshot.go
saima-s: changed.

releasenotes/notes/support-annotation-on-snapshotter-function-ff9b7ba2daf10427.yaml
hairyhum: I understand that this change only affects internal golang API.
I think it still changes the API, so makes sense to mention even if it's not exposed to the function args (maybe it should be).

viveksinghggits: ok.

viveksinghggits, line 2: Maybe we should phrase it something like this
```
Add support to pass labels and annotations to the methods that create/clone VolumeSnapshot and VolumeSnapshotContent resources.
```

pkg/kube/snapshot/snapshot_test.go
```diff
@@ -1458,7 +1607,83 @@ func (s *SnapshotLocalTestSuite) TestLabels(c *C) {
...
+			var err error
+			var vs *snapv1.VolumeSnapshot
+			snapshotMeta := snapshot.ObjectMeta{
+				Name:        snapName,
+				Namespace:   ns,
+				Annotations: tc.createAnnotations,
+			}
+			err = fakeSs.Create(ctx, volName, &snapClass, false, snapshotMeta)
+			if err == nil {
+				vs, err = fakeSs.Get(ctx, snapName, ns)
```
viveksinghggits, line 1674: should we check the labels as well?

viveksinghggits: it's ok if labels are tested somewhere else because they were already part of the code.

pkg/kube/snapshot/snapshot_test.go
```diff
@@ -1458,7 +1607,83 @@ func (s *SnapshotLocalTestSuite) TestLabels(c *C) {
...
+	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "snapshot.storage.k8s.io", Version: "v1", Kind: "VolumeSnapshotList"}, &unstructured.UnstructuredList{})
+	ns := "namespace"
+	volName := "vol1"
+	snapName := "snap1"
+	snapClass := "snapClass"
+	fakeCli := fake.NewSimpleClientset(fakePVC(volName, ns))
+	for _, tc := range []struct {
+		dynCli            dynamic.Interface
+		createAnnotations map[string]string
+		listAnnotations   map[string]string
```
viveksinghggits, line 1624:
```suggestion
		snapshotAnnotations map[string]string
		expectedAnnotations   map[string]string
```

pkg/kube/snapshot/types.go
```diff
@@ -0,0 +1,13 @@
+package snapshot
+
+// ObjectMeta is metadata of snapshot which is configurable.
```
viveksinghggits, line 3: it's not just the metadata of the snapshot right? It can basically be used for anything.

viveksinghggits, line 3:
```suggestion
// ObjectMeta has common metadata fields that can be used to pass details or a resource around.
```

pkg/kube/snapshot/types.go
```diff
@@ -0,0 +1,13 @@
...
+type ObjectMeta struct {
+	// Name must be unique within a namespace. Is required when creating resources.
+	Name string `json:"name,omitempty"`
+	// Namespace defines the space within which a resource is created.
+	Namespace string `json:"namespace,omitempty"`
+	// Labels are set to group a resource. This can be used to filter certain resource.
+	Labels map[string]string `json:"labels,omitempty"`
+	// Annotations is an unstructured key value map stored. This can be used to retrieve arbitrary metadata.
+	Annotations map[string]string `json:"annotations,omitempty"`
```
viveksinghggits, line 12:
```suggestion
	Name string `json:"name,omitempty"`
	Namespace string `json:"namespace,omitempty"`
	Labels map[string]string `json:"labels,omitempty"`
	Annotations map[string]string `json:"annotations,omitempty"`
```

viveksinghggits (commented):
LGTM

pkg/kube/snapshot/snapshot_test.go
saima-s: Lables are tested as part of `TestLabels`

pkg/kube/snapshot/types.go
```diff
@@ -0,0 +1,9 @@
+package snapshot
+
+// ObjectMeta has common metadata fields that can be used to pass details or a resource around.
```
viveksinghggits, line 3:
```suggestion
// ObjectMeta has common metadata fields that can be used to pass meta data of a resource around.
```

viveksinghggits (approved):

pkg/kube/snapshot/types.go
viveksinghggits: the suggestion that I made earlier had a typo, this fixes that.

pkg/kube/snapshot/snapshot_test.go
```diff
@@ -1458,7 +1607,83 @@ func (s *SnapshotLocalTestSuite) TestLabels(c *C) {
...
+			err = fakeSs.Create(ctx, volName, &snapClass, false, snapshotMeta)
+			if err == nil {
+				vs, err = fakeSs.Get(ctx, snapName, ns)
+				annotation := vs.GetAnnotations()
+				c.Assert(len(annotation), Equals, len(tc.expectedAnnotations))
+				for k := range tc.expectedAnnotations {
+					if _, ok := annotation[k]; !ok {
+						c.Assert(false, Equals, true)
+					}
+				}
```
PrasadG193, line 1681: [nit] Can we use `DeepEquals` check?
```suggestion
				c.Assert(annotation, DeepEquals, tc.expectedAnnotations)
```

pkg/kube/snapshot/snapshot_test.go
```diff
@@ -1458,7 +1607,83 @@ func (s *SnapshotLocalTestSuite) TestLabels(c *C) {
...
+		errChecker          Checker
+	}{
+		{
+			dynCli: dynfake.NewSimpleDynamicClient(scheme),
+			snapshotAnnotations: map[string]string{
+				"annotationtest": "true",
+			},
+			expectedAnnotations: map[string]string{
+				"annotationtest": "true",
+			},
```
PrasadG193, line 1634: [nit] We might not need `expectedAnnotations` as they are always same as `snapshotAnnotations`. We can use `snapshotAnnotations` in assert checker

PrasadG193 (approved):
LGTM. Have minor suggests in unit tests

pkg/kube/snapshot/snapshot_test.go
saima-s: done

pkg/kube/snapshot/snapshot_test.go
saima-s: done

PR #3055, opened 2024-08-21
build(dependabot): group dockerfile updates in dependabot config

hairyhum:
## Change Overview

Directory groups were not enabled before and each dockerfile had to have separate config.
Now we can group multiple directories updates into same config and PR

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

Wait and see what dependabot produces.

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


.github/dependabot.yml
```diff
@@ -51,151 +51,50 @@ updates:
...
+
   # Update pinned pip packages via requiements.txt
   - package-ecosystem: "pip"
-    directory: "/docker/postgres-kanister-tools"
-    schedule:
-      interval: weekly
+    schedule: weekly
     commit-message:
       prefix: "deps(docker,pip):"
+    directories:
```
julio-lopez, line 60: You may still want to create a group so a single PR with all the updates gets created. I assume that is the intent.
I have not checked whether or not the `directories:` key/stanza works inside a group (in `groups:`). Please verify.

```suggestion
    groups:
        docker-pip:
            directories:
```

hairyhum: No, you can't use `directories` key inside the `groups` key. `groups` key is there to select specific dependencies.
When using `directories` dependabot is supposed to group them in a single PR as I understand from the docs.

julio-lopez (approved):

PR #3060, opened 2024-08-23
Add Support for Cache Size Limits for kopia server start

r4rajat:
## Change Overview

Add Support for Cache Size Limits for command `kopia server start`
This PR will enable support for following flags for `kopia server start` command
- `--cache-directory`
- `--content-cache-size-limit-mb`
- `--metadata-cache-size-limit-mb`

without these flags kopia server sometimes takes a lot of storage for caching in cases when we have large data to backup or restore.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


pkg/kopia/command/server_test.go
```diff
@@ -41,6 +45,8 @@ func (kServer *KopiaServerTestSuite) TestServerCommands(c *C) {
 			f: func() []string {
 				args := ServerStartCommandArgs{
 					CommandArgs:          commandArgs,
+					CacheArgs:            cacheArgs,
+					CacheDirectory:       "cache/dir",
```
viveksinghggits, line 49: we are passing values of these fields, same everywhere. We should change that in every test case with diff possible values.

viveksinghggits (commented):
LGTM

pkg/kopia/command/server_test.go
r4rajat: @viveksinghggits But on other hand, we're also keeping the `commandArgs` same for all tests, which contains details like `RepoPassword`, `ConfigFilePath` & `LogDirectory`. So thats's why I also kept the `cacheArgs` same for all tests. But if you want we can keep them different. WDYT ?

viveksinghggits: I think it should be ok, even if we kept them the same.

pkg/kopia/command/server.go
```diff
@@ -60,6 +62,8 @@ func ServerStart(cmdArgs ServerStartCommandArgs) []string {
 		args = args.AppendLoggable(readOnlyFlag)
 	}
 
+	args = cmdArgs.kopiaCacheArgs(args, cmdArgs.CacheDirectory)
+
```
PrasadG193, line 66: [nit] Let's move this after `args := commonArgs()` on line no 41.

PrasadG193 (approved):
LGTM

pkg/kopia/command/server.go
r4rajat: Done !

PrasadG193:
@r4rajat can you take a look at CI failure?

PR #3063, opened 2024-08-23
deps(docker): bump kopia version for kanister-tools

hairyhum:
## Change Overview

~~Use current master of kopia to build kanister-tools image~~

Not using the current master because it messes up with the go toolchain at the moment.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [x] :green_heart: E2E


julio-lopez (changes requested):
Unpinning may have undesired side-effects, such as
- inadvertently upgrading and picking up a different version of kopia on every build (or invocation of goreleaser).
- not being able reprodude a released version with the same version of kopia that was used when it got released.
- also, there is a specific kopia dependency in `go.mod`, which is kept in sync with the one here. The version used here should be pulled from `go.mod` at build time.

hairyhum:
> also, there is a specific kopia dependency in go.mod, which is kept in sync with the one here. The version used here should be pulled from go.mod at build time.

That's a good point, but this means we didn't keep kopia version in `go.mod` up to date. Pulling version from `go.mod` should be possible.

hairyhum:
Bumped kopia commit in go.mod and goreleaser.
Not using the latest version because we have issue updating the build image right now https://github.com/kanisterio/kanister/actions/runs/11295843718

julio-lopez (approved):

PR #3075, opened 2024-08-29
Added the security context at pod and container level for Kainster operator

### LINKED ISSUE #2979: Expose securityContext on kanister-operator deployment to improve security

As for now the Deployment object for kanister-operator does not allow securityContext modifications.

Modify Deployment https://github.com/kanisterio/kanister/blob/master/helm/kanister-operator/templates/deployment.yaml to allow Pod SecurityContext and containerSecurity context customization.

anishbista60:
## Change Overview

Expose securityContext on kanister-operator deployment to improve security
## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #2979 

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


anishbista60:
@pavannd1 could you please look at this PR ?

helm/kanister-operator/values.yaml
```diff
@@ -89,3 +89,22 @@ tolerations: []
...
+
+# Container-level security context settings
+containerSecurityContext:
+  capabilities:
+    drop:
+    - ALL
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  # AllowPrivilegeEscalation should be disabled for security
+  allowPrivilegeEscalation: false
```
pavannd1, line 110:
```suggestion
  allowPrivilegeEscalation: false
  
```

anishbista60: @pavannd1 suggested changes are same as before. Should i change it to true ?

julio-lopez: @anishbista60 the change adds a new line at the end, before the file did not end in EOL.

anishbista60:
@julio-lopez please test it .

viveksinghggits:
> @viveksinghggits could you please look into this PR sir ?

Can you please add the test plan. We have a helm test written here `pkg/testing/helm_test.go`. Let's improve that to verify this change as well.

Also, you don't have to sir, just Vivek is fine.

viveksinghggits:
> @julio-lopez please test it .

Usually when someone raises a PR it's their responsibility to add proper test plan and get help to test the changes. Like suggested above, let's add some tests and we can try to merge.

helm/kanister-operator/values.yaml
```diff
@@ -89,3 +89,23 @@ tolerations: []
...
+
```
viveksinghggits, line 111: Something seems to be off here. A newline is not added properly in the file. Can you please fix this.

anishbista60:
> > @viveksinghggits could you please look into this PR sir ?
> 
> Can you please add the test plan. We have a helm test written here `pkg/testing/helm_test.go`. Let's improve that to verify this change as well.
> 
> Also, you don't have to say sir, just Vivek is fine.

yes sure. Thank you

anishbista60:
@viveksinghggits could you please take a look ?

anishbista60:
@shuguet here also.

helm/kanister-operator/templates/deployment.yaml
```diff
@@ -14,6 +14,12 @@ spec:
       labels:
 {{ include "kanister-operator.helmLabels" . | indent 8}}
     spec:
+{{- if .Values.podSecurityContext }}
+      securityContext: 
+        runAsUser: {{ .Values.podSecurityContext.runAsUser }}
+        fsGroup: {{ .Values.podSecurityContext.fsGroup }}
+        runAsNonRoot: {{ .Values.podSecurityContext.runAsNonRoot }}
```
viveksinghggits, line 21: is there a reason because of which we would just want to support these 3 fields under `securityContext` or should we support configuring all the fields under securityContext?

helm/kanister-operator/values.yaml
```diff
@@ -89,3 +89,20 @@ tolerations: []
...
+# Container-level security context settings
+containerSecurityContext:
+#  capabilities:
+#    drop:
+#    - ALL
+#  readOnlyRootFilesystem: true
+#  runAsNonRoot: true
+# AllowPrivilegeEscalation should be disabled for security
+#  allowPrivilegeEscalation: false
```
viveksinghggits, line 107: I think these fields should go under the controller field that we have above. Since we are configuring the the security context of the controller pod. what do you think?

pkg/testing/helm/helm_test.go
```diff
@@ -199,6 +199,101 @@ func (h *HelmTestSuite) TestSelectedDeploymentAttrFromKanisterHelmDryRunInstall(
...
+		testApp, err := NewHelmApp(tc.helmValues, kanisterName, "../../../helm/kanister-operator", kanisterName, "", true)
+		c.Assert(err, IsNil)
+
+		out, err := testApp.Install()
+		c.Assert(err, IsNil)
+
+		resources := helm.ResourcesFromRenderedManifest(out, func(kind helm.K8sObjectType) bool {
+			return kind == helm.K8sObjectTypeDeployment
+		})
+		c.Assert(len(resources) > 0, Equals, true)
```
viveksinghggits, line 276: Instead of checking `> 0` let's check the exact number of resources.

viveksinghggits (changes requested):

anishbista60:
@viveksinghggits Thanks for your review. I think now it's look fine according to your recommendation. 
please let me know if any changes is needed.

helm/kanister-operator/values.yaml
```diff
@@ -36,6 +36,20 @@ controller:
     # false : kanister-prometheus framework has been disabled
     # true: kanister-prometheus framework has been enabled
     enabled: false
+  # Pod-level security context settings
+  podSecurityContext:
+  # runAsUser: 1000
+  # fsGroup: 2000
+  # runAsNonRoot: true
```
viveksinghggits, line 43: I think we should remove these comments. Same for `containerSecurityContext`.

helm/kanister-operator/templates/deployment.yaml
```diff
@@ -53,6 +61,10 @@ spec:
       - name: {{ template "repository-server-controller.name" . }}
         image: {{ .Values.repositoryServerControllerImage.registry }}/{{ .Values.repositoryServerControllerImage.name }}:{{ .Values.repositoryServerControllerImage.tag }}
         imagePullPolicy: {{ .Values.repositoryServerControllerImage.pullPolicy }}
+{{- if .Values.controller.containerSecurityContext }}
+        securityContext: 
+{{ toYaml .Values.controller.containerSecurityContext | indent 10 }}
+{{- end }}
```
viveksinghggits, line 67: Let's create a function is `_helpers.tpl` and use that at both the places for containers. We can do the same for deployment as well.

helm/kanister-operator/values.yaml
```diff
@@ -89,3 +103,4 @@ tolerations: []
 #
 # node labels for pod assignment. Evaluated as template
 nodeSelector: {}
+
```
viveksinghggits, line 106:
```suggestion

```

helm/kanister-operator/values.yaml
anishbista60: @viveksinghggits if i do this. Then, i think there is not need to test for `Only Container security context is set ` and `Only pod security context is set.

viveksinghggits: sorry, I didn't understand this. By removing the comments are are not actually changing anything right? because it's (the helm fields) already commented out. So it it's commented out just removing them should not make any difference.

anishbista60: The second test case throws the following error when running `make helm-test`:

```
Test name: Only Container security context is set
helm_test.go:284:
    c.Assert(obj.Spec.Template.Spec.SecurityContext, check.DeepEquals, tc.expectedPodSecurity)
... obtained *v1.PodSecurityContext = &v1.PodSecurityContext{SELinuxOptions:(*v1.SELinuxOptions)(nil), WindowsOptions:(*v1.WindowsSecurityContextOptions)(nil), RunAsUser:(*int64)(0xc0002b5958), RunAsGroup:(*int64)(nil), RunAsNonRoot:(*bool)(0xc0002b59a0), SupplementalGroups:[]int64(nil), FSGroup:(*int64)(0xc0002b5988), Sysctls:[]v1.Sysctl(nil), FSGroupChangePolicy:(*v1.PodFSGroupChangePolicy)(nil), SeccompProfile:(*v1.SeccompProfile)(nil)} ("&PodSecurityContext{SELinuxOptions:nil,RunAsUser:*1000,RunAsNonRoot:*true,SupplementalGroups:[],FSGroup:*2000,RunAsGroup:nil,Sysctls:[]Sysctl{},WindowsOptions:nil,FSGroupChangePolicy:nil,SeccompProfile:nil,}")
... expected *v1.PodSecurityContext = (*v1.PodSecurityContext)(nil) ("nil")

FAIL: helm_test.go:203: HelmTestSuite.TestSecurityContextInHelmChart
```

- The `expectedPodSecurity` in this test case is `nil`, which works fine when I comment out the `podSecurityContext` in `values.yaml`  because the values for `podsecuritycontext` becomes`nil` and match the expected value.
- However, once the `podSecurityContext` is uncommented, the  value for `podsecuritycontext` changes, but the test case still expects `nil`. 
- If I modify the expected value to match what's in `values.yaml`, it ends up being the same as the first test case.
- Similar for third test case.

viveksinghggits: hmm, I think we we are doing something wrong. do you agree that removing the commented fields from `values.yaml` should not have any affect anywhere?

anishbista60: yes , it affect . The value that we have taken in `test file` would take precendence over the value that are present in` value.yaml`. if the values on our test file is nil or empty then values will be taken from `values.yaml`.

viveksinghggits: > over the value that are present in value.yaml

my understanding is we are not configuring any values in values.yaml right? because we have just commented them out.

anishbista60: yes, we are not using whenever it is commented.

anishbista60: If we actually want to run operator  with security context , then we have to uncomment the changes. And as I told earlier test 2 and test 3 is no more needed because we have enabled both security context.

anishbista60: Any suggestion you would add here ? Thank you

anishbista60: > Any suggestion you would like to add here ? Thank you

anishbista60:
@viveksinghggits shall i go for removing that 2nd and 3 rd  test cases ?

helm/kanister-operator/values.yaml
```diff
@@ -36,6 +36,20 @@ controller:
...
+    runAsNonRoot: true
+
+  # Container-level security context settings
+  containerSecurityContext:
+    capabilities:
+      drop:
+      - ALL
+    readOnlyRootFilesystem: true
+    runAsNonRoot: true
+    allowPrivilegeEscalation: false
```
viveksinghggits, line 52: I don't think this is required right? We just want to add the capability to configure the security context we don't necessarily want to run the containers/pods with specific security context at least in this PR.

viveksinghggits (commented):
@anishbista60 so where we commit/push files to github, we should have a newline at the end of the file. It's general feedback for the files in this PR. Can you please make sure that they are committed properly so that they have the new line at the end.

helm/kanister-operator/templates/deployment.yaml
viveksinghggits: @anishbista60 I see you have resolved this comment, but I don't see the changes for the deployment security context.

helm/kanister-operator/values.yaml
viveksinghggits: I see this is resolved as well, have we moved these fields under the controller field?

helm/kanister-operator/values.yaml
anishbista60: Ok, then i will comment out these field.

helm/kanister-operator/templates/deployment.yaml
anishbista60: In previous feedback you told me to create `_helpers.tpl` for `containersecuritycontext`  which i already did. Now, the question is should i configure for `podsecuritylevel` right ?

helm/kanister-operator/values.yaml
viveksinghggits: I was just wondering why do we have to comment them? Can we not just remove them from here? This is what I wanted to request in my first comment in other discussion as well.

anishbista60: got that. That also we can do.

anishbista60:
now , i think looks good .

anishbista60:
Anyone here please take a look and try to get it merge.

helm/kanister-operator/values.yaml
```diff
@@ -89,3 +89,4 @@ tolerations: []
 #
 # node labels for pod assignment. Evaluated as template
 nodeSelector: {}
+
```
viveksinghggits, line 92: are we not adding any new fields in values.yaml for `.Values.controller.containerSecurityContext`? Is it already part of the values.yaml?

helm/kanister-operator/templates/_helpers.tpl
```diff
@@ -111,3 +111,12 @@ Define a custom kanister-tools image
...
 {{- end -}}
+
+{{/*
+Define a security Context at Container level
+*/}}
+{{- define "controller.containerSecurityContext" -}}
+{{- if .Values.controller.containerSecurityContext -}}
+{{ toYaml .Values.controller.containerSecurityContext | indent 10  }}
+{{- end -}}
+{{- end -}}
```
viveksinghggits, line 122:
```suggestion
{{/*
Define a security Context at Container level
*/}}
{{- define "controller.containerSecurityContext" -}}
    {{- if .Values.controller.containerSecurityContext -}}
        {{ toYaml .Values.controller.containerSecurityContext | indent 10  }}
    {{- end -}}
{{- end -}}
```

helm/kanister-operator/values.yaml
anishbista60: @viveksinghggits no we are not adding . Please check this [discussion] (https://github.com/kanisterio/kanister/pull/3075#discussion_r1776908764)

and regarding for nodeSelector . It was already there before this PR. i added the new line based on this [recommendation] (https://github.com/kanisterio/kanister/pull/3075#pullrequestreview-2330976614)

viveksinghggits: @anishbista60 I think there was a miscommunication. I was not talking about removing entire helm field. I was talking about removing the things that you had commented out https://github.com/kanisterio/kanister/pull/3075#discussion_r1770596824 and removing the default values that you had put. We would still the new helm fields that the PR is introducing.

hairyhum:
@anishbista60 @viveksinghggits the PR was marked stale. Is there any progress? Should someone take over?

viveksinghggits:
> @anishbista60 @viveksinghggits the PR was marked stale. Is there any progress? Should someone take over?

I think @anishbista60 didn't get change to work on this comment https://github.com/kanisterio/kanister/pull/3075#discussion_r1810525801 otherwise it could have been a good addition.

PR #3095, opened 2024-09-06
feat: Introduce new MultiContainerRun function to run pods with two containers

hairyhum:
## Change Overview

In order to separate data dump generation by database tools and export by `kando`, a pod performing database dump can have two separate containers and set up the pipe over file instead of anonymous pipe.

The goal of this function is to enable database-specific blueprints to use official database images separately from kanister-tools image instead of building a custom image for each databse blueprint.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E


hairyhum:
There is an example in tests and docs.
Another example of blueprint:
```
kind: Blueprint
metadata:
  name: dump-bp
  namespace: kanister
actions:
  dump:
    phases:
    - func: Dump
      name: dump
      args:
        namespace: kanister
        sharedStorageMedium: Memory
        sharedStorageSize: 1Gi
        sharedStorageDir: "/tmp/"
        dumpImage: ubuntu:latest
        dumpCommand:
        - bash
        - -o
        - errexit
        - -o
        - pipefail
        - -c
        - |
          mkfifo /tmp/pipe-file
          for i in {1..10}
          do
            echo $i
            sleep 0.5
          done > /tmp/pipe-file
        exportImage: ghcr.io/kanisterio/kanister-tools:v9.99.9-dev
        exportCommand:
        - bash
        - -o
        - errexit
        - -o
        - pipefail
        - -c
        - |
          while [ ! -e /tmp/pipe-file  ]; do sleep 1; done
          kando output numbers "$(cat /tmp/pipe-file)"

```

docs/functions.rst
```diff
@@ -108,23 +108,32 @@ Example:
         - |
           echo "Example"
 
-KubeTask
+Dump
```
viveksinghggits, line 111: are we removing KubeTask function in favour of this function? docs changes look like that but release note say new function is added.

viveksinghggits:
Hi @hairyhum ,
Are we doing this to solve any performance issue that were reported? Can you please share some background just for my understanding.

hairyhum:
> Hi @hairyhum , Are we doing this to solve any performance issue that were reported? Can you please share some background just for my understanding.

The main goal of this change is to separate image management for databse dump blueprints. Not the performance.

hairyhum:
Renamed function into `KubeTaskParallel` and the arguments for containers to `background` and `output` containers.

e-sumin (commented):
Overall LGTM, but for me function name looks a bit misleading.
But, frankly, I can't propose better name. Maybe other reviewers will have an ideas ?

pkg/function/kube_task_parallel.go
```diff
@@ -0,0 +1,325 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: jobPrefix,
+			Namespace:    ktpf.namespace,
+			Labels:       ktpf.labels,
+			Annotations:  ktpf.annotations,
+		},
+		Spec: podSpec,
+	}
+
+	log.Info().Print("POD: ", field.M{"pod": pod})
```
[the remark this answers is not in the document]
hairyhum: Removed

e-sumin (approved):
LGTM

docs/functions.rst
```diff
@@ -150,6 +150,82 @@ Example:
         - |
           echo "Example"
 
+KubeTaskParallel
```
viveksinghggits, line 153: I think maybe we should the name of the function in next community call? It's good right now, but by reading the name it looks like in this function we are doing things parallelly that kubetask does sequentially, which doesn't seem to be the case right?
I think it doesn't necessarily have to have KubeTask in it's name right? Maybe `MultiContainerUpload` or something in that line?

docs/functions.rst
```diff
@@ -150,6 +150,82 @@ Example:
...
+----------------
+
+KubeTaskParallel spins up a new pod with two containers connected
+via shared `emptyDir`_ volume.
+It's similar to KubeTask, but allows using multiple images to move backup data.
+"background" container is one responsible for generating data, while "output" container
+should export it to destination.
+The main difference between them is that phase outputs can only generated from the
```
viveksinghggits, line 161:
```suggestion
The main difference between these containers is that phase outputs can only be generated from the
```

nit

docs/functions.rst
```diff
@@ -150,6 +150,82 @@ Example:
...
+"output" container outputs.
```
viveksinghggits, line 162:
```suggestion
"output" container.
```
?

docs/functions.rst
```diff
@@ -150,6 +150,82 @@ Example:
...
+The function also supports an optional init container to set up the volume contents.
+
+.. csv-table::
+   :header: "Argument", "Required", "Type", "Description"
+   :align: left
+   :widths: 5,5,5,15
+
+   `namespace`, No, `string`, namespace in which to execute (the pod will be created in controller's namespace if not specified)
+   `backgroundImage`, Yes, `string`, image to be used in "background" container
+   `backgroundCommand`, Yes, `[]string`,  command list to execute in "background" container
```
viveksinghggits, line 172: should we call this container snapshot or backup container? or maybe data container? just a suggestion, if you think `background` is ok, I am good.

docs/functions.rst
```diff
@@ -150,6 +150,82 @@ Example:
...
+   `outputImage`, Yes, `string`, image to be used in "output" container
+   `outputCommand`, Yes, `[]string`,  command list to execute in "output" container
```
viveksinghggits, line 174: should this be called upload container?

releasenotes/notes/kube-task-parallel-function-d488516c0f3b22c6.yaml
```diff
@@ -0,0 +1,2 @@
+---
+features: Introduce new KubeTaskParallel function to run pods with two containers connected by shared volume
```
viveksinghggits, line 2: nit:
```suggestion
features: Introduced new Kanister function ``KubeTaskParallel`` to run pods with two containers connected by shared volume
```
?

viveksinghggits (commented):
looks good to me overall. just left some nit comments/formatting suggestion.

docs/functions.rst
hairyhum: I was trying to avoid the "purpose" style names like "dump", "upload" or "export". While this function is used for that purpose it's deliberately made generic.

docs/functions.rst
hairyhum: Same note as a note before, trying to name things for what they do rather than what they can be used for.

docs/functions.rst
viveksinghggits: ok, let's see what others think @pavannd1 @PrasadG193 @mlavi .

PrasadG193: Somehow `KubeTaskParallel` doesn't sound good to me. I agree that it would be better if we avoid `KubeTask*` prefix for the new function.

hairyhum: We can use `MultiContainerPod` or `MultiContainerRun`, but we already use `KubeTaslk` instead of `Pod` in the other function.

viveksinghggits: I like `MultiContainerRun`. Do we need to change the arguments name as well? that are currently named background and ouptut I think.

pavannd1: Naming is hard. Okay with `MultiContainerRun` or maybe `MultiContainerJob`

hairyhum: I don't have better names than `background` and `output` for containers.

pavannd1: I was thinking about the use of the named pipe - one of the containers pushes data while the other one receives it.
Here are some options I could think of:
- `input` and `output`
- `producer` and `consumer`
- `reader` and `writer`

hairyhum: > I was thinking about the use of the named pipe - one of the containers pushes data while the other one receives it.

That is not a part of the function itself, but a part of the command. The only practical difference in the implementation of those containers is that only one of them sends its ouput into the function output. Otherwise you can make the output container wirte data to a file and you can make the background container read from the file.

hairyhum:
Renamed function to `MultiContainerRun'

PR #3177, opened 2024-10-04
Tail log output for kopia server commands

aaron-kasten:
## Change Overview

This PR adds log tailing for each kopia command.  The log tailing occurs concurrently with the command, allowing the output of a log file to stdout and/or stderr for the process.

Unit tests pending.

## Pull request type

- [X] :construction: Work in Progress
- [X] :sunflower: Feature

## Issues 

k10-25828

## Test Plan

Will add unit tests.  Ideally will also add integration testing.

- [X] :muscle: Manual (current)
- [X] :zap: Unit test (future)


pkg/controllers/repositoryserver/server.go
```diff
@@ -65,6 +65,9 @@ func (h *RepoServerHandler) startRepoProxyServer(ctx context.Context) (err error
 			Background:       true,
 		},
 	)
+
+	cmd = append([]string{"tail", "-v", "-F", "/tmp/kopia-log/cli-logs/latest.log", "&"}, cmd...)
```
julio-lopez, line 69: @aaron-kasten

IIUC, this is tailing the logs as part of the same KubeExec call.
What about redirecting the `kopia server` output to a file that can be read or tailed from the container's **entry point** ?

Note: `latest.log` is a symlink, does that affect `tails -F` behavior?

Questions:
- Are kubeExec commands executed via a shell? Pardon my ignorance, this is for my edification.
- What are the effects of tailing the log output as part of the kubeExec? what happens when the connection is interrupted?

pkg/kopia/cli/internal/opts/common_opts.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	defaultLogLevel = "error"
+	defaultLogLevel = "debug"
```
julio-lopez, line 24: Is this a temporary change only for debugging the PR?

Does `defaultLogLevel` apply to:
  1. the logs sent to the console;
  2. the logs sent to files;
  3. both 1 & 2?

pkg/kopia/command/common.go
```diff
@@ -61,7 +61,8 @@ type CommandArgs struct {
 
 func bashCommand(args logsafe.Cmd) []string {
```
julio-lopez, line 62: This answer my question from [above](#r1795739134)

pkg/kopia/command/common.go
```diff
@@ -43,7 +47,7 @@ func GetEnvOrDefault(name, def string) string {
 
 // LogLevel will return either value from env or "error" as default value
 func LogLevel() string {
-	return GetEnvOrDefault(LogLevelVarName, LogLevelError)
+	return GetEnvOrDefault(LogLevelVarName, "debug")
```
e-sumin, line 50: Seems that comment and code are not consistent.

aaron-kasten:
Went with chaning ENDPOINT in K10 as a solution

PR #3195, opened 2024-10-23
build: bump golangci-lint version in build image

hairyhum:
## Change Overview

Old version of golangci-lint has a memory leak causing it to crash with oom in recent go versions.
Updated golangci-lint version in build image.
Fixed lint errors.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/function/create_volume_snapshot.go
+++ b/pkg/function/create_volume_snapshot.go
@@ -134,7 +134,7 @@ func createVolumeSnapshot(ctx context.Context, tp param.TemplateParams, cli kube
 	}
 	wg.Wait()
 
-	err := fmt.Errorf(strings.Join(errstrings, "\n"))
+	err := fmt.Errorf("%s", strings.Join(errstrings, "\n"))
 	if len(err.Error()) > 0 {
 		return nil, errkit.Wrap(err, "Failed to snapshot one of the volumes")
 	}
--- a/pkg/function/create_volume_snapshot.go
+++ b/pkg/function/create_volume_snapshot.go
@@ -134,7 +134,7 @@ func createVolumeSnapshot(ctx context.Context, tp param.TemplateParams, cli kube
 	}
 	wg.Wait()
 
-	err := fmt.Errorf("%s", strings.Join(errstrings, "\n"))
+	err := errkit.New(strings.Join(errstrings, "\n"))
 	if len(err.Error()) > 0 {
 		return nil, errkit.Wrap(err, "Failed to snapshot one of the volumes")
 	}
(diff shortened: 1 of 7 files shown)


pkg/function/delete_rds_snapshot.go
```diff
@@ -55,22 +55,22 @@ func (*deleteRDSSnapshotFunc) Name() string {
 	return DeleteRDSSnapshotFuncName
 }
 
-func deleteRDSSnapshot(ctx context.Context, snapshotID string, profile *param.Profile, dbEngine RDSDBEngine) (map[string]interface{}, error) {
```
KastenMike, line 58: Should this be part of this PR?

hairyhum: This fixes an `unparam` linter warning.

julio-lopez: @KastenMike I am assuming this was flagged by a linter as part of the upgrade. @hairyhum can confirm.

If that's the case, there are 2 options:
- Put all the changes in a single PR, like this one; or
- 2 PRs in this order: the first one addressing the linter warnings, and the second one upgrading the linter version.

pkg/function/create_volume_snapshot.go
```diff
@@ -134,7 +134,7 @@ func createVolumeSnapshot(ctx context.Context, tp param.TemplateParams, cli kube
 	}
 	wg.Wait()
 
-	err := fmt.Errorf(strings.Join(errstrings, "\n"))
+	err := fmt.Errorf("%s", strings.Join(errstrings, "\n"))
```
julio-lopez, line 137: It seems that this should use `errors.New`

```suggestion
	err := errors.New(strings.Join(errstrings, "\n"))
```

Question unrelated to this PR why are newlines (`\n`) being used inside the error message?

julio-lopez (approved):
LG. See minor comment

pkg/function/create_volume_snapshot.go
hairyhum: Or maybe errkit, but I was trying to keep the changes minimal.

pkg/function/delete_rds_snapshot.go
hairyhum: Upgrading linter version is a github action, which I've already run, so it's just a matter of getting it to the `master` codebase

pkg/function/create_volume_snapshot.go
hairyhum: Changed to errkit because we already have it in this file.

Issue #3202, opened 2024-10-25
Fix code scanning alert - Incorrect conversion between integer types


viveksinghggits:
Tracking issue for:
- [ ] https://github.com/kanisterio/kanister/security/code-scanning/200

PR #3205, opened 2024-10-25
Fix incorrect conversion between int types

viveksinghggits:
## Change Overview

Fixes https://github.com/kanisterio/kanister/issues/3202

In the older version of the logic where we were reading replicas of workload in `ScaleWorkload` function; if val (string type) has an int value that is beyond the range of int32, even though call `strconv.Atoi(val)` would succeed, the next of `replicas = int32(v)` would result into unexpected result because the val is beyond what can be accommodated into int32 type.
This change fixes that.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes https://github.com/kanisterio/kanister/issues/3202

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

```
/kanister/pkg/function (fix-codescanning-incorrect-intconv*) » go test -check.f "ScaleWorkloadSuite"
OK: 1 passed
PASS
ok      github.com/kanisterio/kanister/pkg/function     0.894s
```


pkg/function/scale_workload.go
```diff
@@ -167,8 +167,8 @@ func (s *scaleWorkloadFunc) setArgs(tp param.TemplateParams, args map[string]int
 	case int64:
 		replicas = int32(val)
 	case string:
-		var v int
-		if v, err = strconv.Atoi(val); err != nil {
+		v, err := strconv.ParseInt(val, 10, 32)
```
viveksinghggits, line 170: In the older version if `val` (`string` type) has an int value that is beyond the range of `int32`, even though call `strconv.Atoi(val)` would succeed, the next of `replicas = int32(v)` would result into unexpected result because the val is beyond what can be accommodated into int32 type.
This change fixes that.

PrasadG193 (approved):

pkg/function/scale_workload_test.go
```diff
@@ -0,0 +1,46 @@
+package function
```
pavannd1, line 1: This file needs the license header

viveksinghggits: I will make sure that I add this as part of one of the next PRs.

PR #3210, opened 2024-11-01
Install gcloud auth plugin in kanister build image

viveksinghggits:
## Change Overview

When running the integration test using `make integration-test` we are asked to install minio using `make install-minio` but if our k8s cluster is gke and in the kuebconfig gcloud auth plugin is used, `make install-minio` fails with below error

```
Unable to connect to the server: getting credentials: exec: executable gke-gcloud-auth-plugin not found

It looks like you are trying to use a client-go credential plugin that is not installed.

To learn more about this feature, consult the documentation available at:
      https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins

Install gke-gcloud-auth-plugin for use with kubectl by following https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#install_plugin
```
And this seems to be happening because our build image doesn't have gcloud auth plugin installed. This PR installs that.

## Pull request type

- [x] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/docker/build/Dockerfile
+++ b/docker/build/Dockerfile
@@ -13,6 +13,7 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /
 RUN apt update && apt install -y docker-ce docker-ce-cli containerd.io \
     && apt-get clean
 
+RUN apt-get install apt-transport-https ca-certificates gnupg
 RUN apt-get install -y google-cloud-sdk-gke-gcloud-auth-plugin
 
 COPY --from=bitnami/kubectl:1.26 /opt/bitnami/kubectl/bin/kubectl /usr/local/bin/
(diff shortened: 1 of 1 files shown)


docker/build/Dockerfile
```diff
@@ -13,6 +13,13 @@ RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /
 RUN apt update && apt install -y docker-ce docker-ce-cli containerd.io \
     && apt-get clean
 
+RUN apt-get install apt-transport-https ca-certificates
```
julio-lopez, line 16: Any reason for not adding this in the first `apt-get install` above (line 7)?

Also, remove the `apt-get clean` above (line 14) and move it to after the last `apt-get install` below.

viveksinghggits: Hi @julio-lopez ,
Thank you for the review, I can work on above suggestions.
But I initially thought that we would just need to install gcloud and gcloud auth plugin, in the kanister build image, and things would work but that doesn't seem to be the case.

We will also have to set up gcloud authentication, which I think needs to be discussed with the team and then worked on, so I am planning to park this for now and work on it after discussing this in community call.

PR #3269, opened 2024-12-01
Update ubi-minimal base image to ubi-minimal:9.5-1731593028

infraq:
Update ubi-minimal base image to ubi-minimal:9.5-1731593028


viveksinghggits:
It looks like we will need to check why CI is failing it looks like the commits are not properly signed.

PR #3289, opened 2024-12-13
feat: Add options to specify AWS credential source for RDS functions 

hairyhum:
## Change Overview

Currently AWS credentials are always taken from the profile, which limits
the ability to export into non-s3 storage and s3 storage in other regions.

Added options to specify a separate secret or use IAM role for service account.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- Run RDS example with new arguments added in the blueprint
  - [ ] Secret credentials with S3 in another region
  - [ ] Serviceaccount credentials

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


viveksinghggits:
Hi @hairyhum,
I was looking into the changes and was wondering if it would be a good idea to just extend profile in such a way that we can accept the new fields/IAM from profile instead of introducing the new secret way.
I think introducing the secret to accept the creds complicates the things and maybe blueprints as well.
Or, do you think it's not a good idea to just extend the profile for this new thing.

hairyhum:
> I was looking into the changes and was wondering if it would be a good idea to just extend profile in such a way that we can accept the new fields/IAM from profile instead of introducing the new secret way.

There is a principal issue with using profile as a source of the credentials. Because the credentials here are the *source* credentials, while profile is by design the *destination* credential.
So we either need to extend profile to have both source and destination, which doesn't make sense outside of RDS case, or we need to use another API to supply source credentials.

viveksinghggits:
Hi @hairyhum ,
Thank you for the explanation, it looks like I totally misunderstood the PR. I will look into the PR again.

docs_new/functions.md
```diff
@@ -938,6 +938,15 @@ Arguments:
   | ---------- | :------: | ------ | ----------- |
   | instanceID | Yes      | string | ID of RDS instance you want to create snapshot of |
   | dbEngine   | No       | string | Required in case of RDS Aurora instance. Supported DB Engines: `aurora` `aurora-mysql` and `aurora-postgresql` |
+  | credentialsSource | No | string | Source for aws credentials. Supported sources: `profile`, `secret`, `serviceaccount`. Default value is `profile` |
+  | credentialsSecret | No | string | Secret to get credentials from. Only used with `credentialsSource: secret`. Secret with this name should be referenced in the Actionset |
+  | region | No | string | AWS region to use. Derived from profile or serviceaccount if not set |
+
+::: tip NOTE
+
+`credentialsSource: serviceaccount` uses kanister operator service account. IAM role for service
```
viveksinghggits, line 947: would it not be a good idea to have another field (`credentialsSA`) like `credentialsSecret`, that can can be used to configured the service account?

docs_new/functions.md
```diff
@@ -1006,6 +1019,19 @@ associated with instance with `instanceID` and will pass the same. - If
 set, `default` DB Subnet group will be used.
 :::
 
+::: tip NOTE
+
+If `credentialsSource` is configured to `profile` (default), the profile used has to be
+an S3 profile in the same region as the database snapshot.
+It's recommended to export to another region. To do that `credentialSource: secret` or `credentialSource: serviceaccount` can be used.
```
viveksinghggits, line 1026:
```suggestion
If `credentialsSource` is configured to `profile` (default behaviour), the profile used has to be
an S3 profile configured with the same region as the database snapshot.
If it's required to export to another region, `credentialSource: secret` or `credentialSource: serviceaccount` can be used.
```

releasenotes/notes/rds-credentials-1fa9817a21a2d80a.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+  - Add AWS credential source configurations to RDS functions
```
viveksinghggits, line 3: `Enable RDS functions to accept AWS credentials using a Secret or ServiceAccount.`

WDYT?

viveksinghggits (commented):
LGTM overall. left some small comments.

docs_new/functions.md
hairyhum: Maybe. The idea here was that an AWS IAM admin would allow kanister service account access to RDS by creating a role for it.
Associating a different service account would be better because it's not trivial to assign multiple roles to a single SA, but implementation-wise it's harder because we currently run RDS operation from within kanister controller (in the function code) meaning that we can't change SA on the flight there.

We can also try using AssumeRole and configure which role it will assume. This is going to be more flexible with less code changes.

I'd suggest we keep it for the follow-up PRs.

releasenotes/notes/rds-credentials-1fa9817a21a2d80a.yaml
hairyhum: Yes, it's better.

docs_new/functions.md
hairyhum: Updated

docs_new/functions.md
viveksinghggits: ok 👍 .

viveksinghggits (approved):

PR #3321, opened 2025-01-13
add Wait command

aaron-kasten:
## Change Overview

Add WaitProcess to kanx client/server. WaitProcess blocks for the completion of the process identified by PID. Implementing this protocol simplifies waiting for process completion both as an internal API and from the command line.

## Pull request type

- [X] :sunflower: Feature

## Issues 

There is no GitHub issue associated with this PR.

## Test Plan

- [X] :zap: Unit test


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kanisterio/kanister
 
-go 1.22.7
+go 1.23.2
```
tdmanv, line 3: please separate this change. will this break dependent projects that use this as a library?

aaron-kasten: no.  ive skipped this pr now for the prs in the chain.  the function was being used to wait after sending signals in the Signal API tests, but ive backported those to use Stdout and GetProcess.

PR #3325, opened 2025-01-14
feat: C: add Signal command

aaron-kasten:
## Change Overview

Add SignalProcess to kanx client/server. SIgnalProcess sends a signal to the identified process. Implementing this protocol allows for sending signals to remote processes via the API or command line.  All signals are identified by number to keep the API as simple and generic as possible.  Symbolic signals will be considered for cross platform signaling.

This PR adds a `fault` field that stores errors in process communication.

## Pull request type

- [X] :sunflower: Feature

## Issues 

There is no GitHub issue associated with this PR.

## Test Plan

- [X] :zap: Unit test


pkg/kando/process_client.go
```diff
@@ -42,3 +46,11 @@ func processAsJSONFlagValue(cmd *cobra.Command) bool {
 	}
 	return b
 }
+
+func processSignalIdentifierFlagValue(cmd *cobra.Command) int {
+	q, err := cmd.Flags().GetInt(processSignalIdentifierFlagName)
```
tdmanv, line 51: If this parameter is required, why not make it a command line argument? If there's a parameter arguement, we should return an error with usage rather than panic which can be hard to hard to understand.

pkg/kando/process_client_signal.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/encoding/protojson"
+
+	"github.com/kanisterio/kanister/pkg/kanx"
+)
+
+func newProcessClientSignalCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "signal",
```
tdmanv, line 31: please include the required parameters here.

pkg/kando/process_client.go
aaron-kasten: Fair point.  Ill add it as an argument.

tdmanv (approved):

PR #3335, opened 2025-01-17
feat: I: proxy signals to subordinate processes ...

aaron-kasten:
## Change Overview

This PR adds the ability to proxy signals from the `kando process client output` client to a subordinate process on the server.

```
$ kando process server &
$ kando process client create --as-json -- bash -c 'function hi() { echo "hi"; }; trap hi SIGINT; while true; do sleep 1; done'
{"pid":"1921","state":"PROCESS_STATE_RUNNING"}
$ ./kando process client output --signal-proxy 1921 &
[2] 3344

$ kill -SIGINT 3344
hi

$ kill -SIGKILL 3344
[2]+  Killed: 9               ./kando process client output --signal-proxy 1921

$ ./kando process client get 1921
Process:  pid:1921 state:PROCESS_STATE_RUNNING
```

## Pull request type

- [X] :sunflower: Feature

## Issues

There is no GitHub issue for this PR

- fixes #issue-number

## Test Plan

- [X] :zap: Unit test


pkg/kando/process_client_output.go
```diff
@@ -30,22 +41,74 @@ func newProcessClientOutputCommand() *cobra.Command {
...
+		for {
+			select {
+			case sig := <-signalTermChan:
+				ossig, ok := sig.(syscall.Signal)
+				if !ok {
+					log.Info().WithContext(ctx).Print(fmt.Sprintf("signal %v is invalid, ignored for process %d", sig, pid))
+					continue
+				}
+				log.Info().WithContext(ctx).Print(fmt.Sprintf("signal %v received for process %d", sig, pid))
+				_, err := kanx.SignalProcess(ctx, addr, pid, int64(ossig))
```
Shrekster, line 71: one issue here... can kanx reuse a PID number ?

I think linux can, so I'd assume that kanx can too.

Is it possible that you might end up killing the wrong process due to pid reuse ? I'll result in random disconnects in the field for a system that is heavily loaded and is reusing PID numbers.

What do you think ?

Shrekster (approved):
LGTM, no objection, but just know that this is a best effort clean feature. Doesn't plug the hole completely and we may still have orphaned processes in kanx until we completely restart the server.

I completely agree with your "Orphaned Kanx Subprocess" problem that you highlighted over slack, and this must be fixed in a future iteration.

pkg/kando/process_client_output.go
Shrekster: PID reuse wil be solved later, we discussed this over slack

PR #3336, opened 2025-01-17
feat: J: features for remote command line execution from K10

aaron-kasten:
## Change Overview

Features for remote command line execution from K10

## Pull request type

- [X] :sunflower: Feature

## Issues 

There is no GitHub issue for this PR

## Test Plan

- [X] :zap: Unit test

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,7 @@ The [golangci-lint](https://golangci-lint.run/) tool is used to enforce many
 styling and safety rules.
 
 When creating or updating code, ensure the copyright year and author header are up to date
-by statically copying the [boilerplace header](/build/boilerplate.go.txt) at the top or
+by statically copying the [boilerplate header](/build/boilerplate.go.txt) at the top or
 by including it for dynamic source code files.
 
 ### Creating A Local Build
--- a/pkg/kando/process_client.go
+++ b/pkg/kando/process_client.go
@@ -15,11 +15,22 @@
 package kando
 
 import (
+	"context"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/kanisterio/errkit"
 	"github.com/spf13/cobra"
+
+	"github.com/kanisterio/kanister/pkg/kanx"
+	"github.com/kanisterio/kanister/pkg/log"
 )
 
 const (
-	processAsJSONFlagName = "as-json"
+	processSignalProxyFlagName = "signal-proxy"
+	processAsJSONFlagName      = "as-json"
 )
 
 func newProcessClientCommand() *cobra.Command {
@@ -28,6 +39,7 @@ func newProcessClientCommand() *cobra.Command {
 		Short: "Send commands to the process server",
 	}
 	cmd.AddCommand(newProcessClientCreateCommand())
+	cmd.AddCommand(newProcessClientExecuteCommand())
 	cmd.AddCommand(newProcessClientGetCommand())
 	cmd.AddCommand(newProcessClientListCommand())
 	cmd.AddCommand(newProcessClientSignalCommand())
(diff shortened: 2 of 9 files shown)


pkg/kando/process_client.go
```diff
@@ -43,3 +54,40 @@ func processAsJSONFlagValue(cmd *cobra.Command) bool {
 	}
 	return b
 }
+
+func proxySetup(ctx context.Context, addr string, pid int64) {
+	log.Info().WithContext(ctx).Print(fmt.Sprintf("signal proxy is running for process %d", pid))
+	signalTermChan := make(chan os.Signal, 1)
+	signal.Notify(signalTermChan, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+	BREAK:
```
tdmanv, line 63: please put this block in a separate function. the breaks can then be a returns. also interesting to note that syscall.Signal is just an int, not and int64, maybe the type for kanx.SignalProcess should just be an int as well.

pkg/kando/process_client.go
```diff
@@ -43,3 +54,40 @@ func processAsJSONFlagValue(cmd *cobra.Command) bool {
...
+			case sig := <-signalTermChan:
+				ossig, ok := sig.(syscall.Signal)
+				if !ok {
+					log.Info().WithContext(ctx).Print(fmt.Sprintf("signal %v is invalid, ignored for process %d", sig, pid))
+					continue
+				}
+				log.Info().WithContext(ctx).Print(fmt.Sprintf("signal %v received for process %d", sig, pid))
+				_, err := kanx.SignalProcess(ctx, addr, pid, int64(ossig))
+				if err != nil {
+					signal.Reset(ossig)
```
tdmanv, line 75: could you please comment on why you reset here?

aaron-kasten: In this context, I'm mostly **agnostic** on this technique as K8s reliably uses signal escalation to terminate processes.

aaron-kasten: I can argue that it will prevent long running signal handlers from running twice (multiple SIGINTs) due to frustration, timeouts etc... this though can be addresses with a `*Once` or re-establishing the Signal handler once the signal handler has completed successfully.

pkg/kando/process_client.go
aaron-kasten: You were correct about the security error identified by static analysis.  The error was **not** because of conversion to an architecture-specific type.

Yea, the `int` is required by the system call - which I guess makes sense.  My default is to maintain the type of the incoming call for as long as possible - in this case, near the system call.

protobuf doesn't have types whose storage changes according to architecture.  `int64` keeps the incoming API consistent (as in identical to other integer fields).  `int32` I think would be safest because I'm not aware of a go platform that supports free `int` -> `int16` conversion, only `int` to `int32` or `int64`.  I'm also not aware of any signal on any platform that is greater than or equal to `int32`.  `int32` would prevent rounding as `int32`s fit into `int64`s.

aaron-kasten: @tdmanv I recommend that we keep the signal type as either `int64` or `int32`.  The system calls use the architecture specific `int`.  If we convert the signal parameter over to be `int` then the pid should be converted to the same, otherwise there's the irrational casting of types in relation to the dependent system calls.

CONTRIBUTING.md
```diff
@@ -72,6 +72,10 @@ Specifically, this means:
 The [golangci-lint](https://golangci-lint.run/) tool is used to enforce many
 styling and safety rules.
 
+When creating or updating code, ensure the copyright year and author header are up to date
+by statically copying the [boilerplace header](/build/boilerplate.go.txt) at the top or
```
Shrekster, line 76:
```suggestion
by statically copying the [boilerplate header](/build/boilerplate.go.txt) at the top or
```

pkg/kando/process_client.go
```diff
@@ -43,3 +55,45 @@ func processAsJSONFlagValue(cmd *cobra.Command) bool {
 	}
 	return b
 }
+
+func proxySetup(ctx context.Context, addr string, pid int64) {
```
Shrekster, line 59: Can we use the `proxySetup()` from your previous PR that already merged here: https://github.com/kanisterio/kanister/pull/3335/files#diff-2ed1938fd31979cf263c8dfc679327d8641260d70da102c57053e8785693c48bR56

This function seems refactored but is basically identical to the the one preexisting in `pkg/kando/process_client_output.go` inside master branch.

Maybe also rebase your branch from master ?

Shrekster: Oh nvm, looks like you moved this here from `pkg/kando/process_client_output.go` in this PR.

pkg/kando/process_client_execute.go
```diff
@@ -0,0 +1,86 @@
...
+	"io"
+
+	"github.com/kanisterio/errkit"
+	"github.com/spf13/cobra"
+	"google.golang.org/protobuf/encoding/protojson"
+
+	"github.com/kanisterio/kanister/pkg/kanx"
+)
+
+func newProcessClientExecuteCommand() *cobra.Command {
```
Shrekster, line 29: Doesn't kanx already have the execute command line functionality that does `kanx.CreateProcess()` ?
Is there something missing with the existing ?

In other words, xannot we use `kando process client create` or `kando process client output` ? What does your new `kando process client execute` do that `kando process client create` does not already do ?

Shrekster: Discussion ongoing off this thread in Slack ....

Shrekster (approved):

PR #3376, opened 2025-02-21
fix: fixup gRPC buffer overflow bug with unit test.

aaron-kasten:
## Change Overview

gRPC has a over-the-wire buffer limit of 4MB.  gRPC buffer overflow errors can occur if this figure is use as a before the wire buffer measure as gRPC adds overhead (JSON overhead) when sending messages from gRPC API clients.

This fix just divides the gRPC buffer limit in half (2MB) which is much smaller than the theoretical max (theoretical_max = 4MB - protocol overhead) but intuitively large enough to hold enough data.

This also includes a fix for using `bytes.Buffer` as a circular buffer, which its not very good at.  Instead a `[]byte` buffer is used.  If performance is a concern then the buffer copy should be split across 2 threads in a bucket brigade.

## Pull request type

- [X] :bug: Bugfix

## Issues 

There is no github issue for this fix.

## Test Plan

- [X] :zap: Unit test

### DIFF
--- a/pkg/kanx/kanx_test.go
+++ b/pkg/kanx/kanx_test.go
@@ -35,7 +35,7 @@ func tmpDir(c *C) string {
 	// limit the size of the address while retaining some of the properties
 	// of the original directory name
 	hs := adler32.Checksum([]byte(c.TestName()))
-	d, err := os.MkdirTemp("", fmt.Sprintf("%08x%.20s", hs, c.TestName()))
+	d, err := os.MkdirTemp("", fmt.Sprintf("%.20s%08x", c.TestName(), hs))
 	c.Log("Directory: ", d)
 	c.Assert(err, IsNil)
 	return d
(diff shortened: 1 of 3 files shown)


pkg/kanx/kanx_test.go
```diff
@@ -35,7 +35,7 @@ func tmpDir(c *C) string {
 	// limit the size of the address while retaining some of the properties
 	// of the original directory name
 	hs := adler32.Checksum([]byte(c.TestName()))
-	d, err := os.MkdirTemp("", fmt.Sprintf("%08x%.20s", hs, c.TestName()))
```
aaron-kasten, line 38: Change to make these temp files easier to find.  Having the filename hash first can be really confusing when using `ls` and an alphabetical sort.

pkg/kanx/server.go
```diff
@@ -219,7 +218,7 @@ func (s *processServiceServer) streamOutput(ss sender, p *process, fh *os.File)
 			<-t.C
 			continue
 		}
-		o := &Output{Output: buf.String()}
```
aaron-kasten, line 222: Performs an extra allocation for the string object which may add measurable overhead considering how large the buffer is (2MB).  May want to convert this code to a 2 thread bucket brigade at some point.

pkg/kanx/kanx_test.go
```diff
@@ -449,3 +449,37 @@ func (s *KanXSuite) TestCreateProcess_Exit2(c *C) {
 	c.Assert(p0.GetExitErr(), Equals, "signal: killed")
 	c.Assert(p0.GetExitCode(), Equals, int64(-1))
 }
+
+func (s *KanXSuite) TestCreateProcess_BufferOverflow_1(c *C) {
```
aaron-kasten, line 453: Skipped test that demonstrates how the buffer overflow is tested.

pkg/kanx/server.go
```diff
@@ -203,13 +202,13 @@ type sender interface {
 }
 
 func (s *processServiceServer) streamOutput(ss sender, p *process, fh *os.File) error {
-	buf := bytes.NewBuffer(make([]byte, 0, streamBufferBytes)) // 4MiB is the max size of a GRPC request
+	buf := make([]byte, streamBufferBytes) // 4MiB is the max size of a GRPC request
```
hairyhum, line 205: It's no longer 4Mib, right?

aaron-kasten: correct.  comment is confusing.

aaron-kasten:
```suggestion
	buf := make([]byte, streamBufferBytes) // 2MiB buffer for gRPC requests
```

hairyhum (approved):

PR #3496, opened 2025-06-02
build: Remove boringcrypto from default builds

### LINKED ISSUE #3495: Support build without `boringcrypto`
Go 1.24 FIPS support is incompatible with `boringcrypto` which breaks running kanister built images with FIPS flags.

We could remove use of `boringcrypto` from the build process now for building with go1.24+

hairyhum:
## Change Overview

Previously kanister used boringcrypto in order to build with FIPS-compliant mode. Since version 1.24, golang ships FIPS-compliant modules by defatult and there is no need to use external crypto library.

It is still possible to build with boringcrypto by setting the GOEXPERIMENT variable on build tasks, but should be unnecessary.

Also build scripts which produce the release are configured with `GOFIPS140=latest` to enforce FIPS compliant crypto modules by default.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #3495

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pavannd1 (approved):

ryanmt (approved):
LGTM. If there isn't urgency I think having Eliott's vote is more valuable but certainly this seems entirely consistent w/ our other FIPS/1.24 work.

.goreleaser.yml
```diff
@@ -18,11 +18,11 @@ builds:
   env: &env
   - GO111MODULE=on
   - CGO_ENABLED=1
-  - GOEXPERIMENT=boringcrypto
   - CC=gcc
   - CXX=g++
   - GO_EXTLINK_ENABLED=0
-  goos:
+  - GOFIPS140=latest
```
kidgilson, line 24: @hairyhum just FYI that this means this build will use the FIPS 140-3 implementation by default which has repercussions in terms of TLS and other algorithms that are used. Were you intending on default FIPS to true here?

docker/tools/Dockerfile
```diff
@@ -5,7 +5,7 @@ ARG kopia_build_commit=master
 ARG kopia_repo_org=kopia
 ARG restic_vsn=v0.16.5
 ARG gosu_vsn=1.17
-ENV CGO_ENABLED=1 GOEXPERIMENT=boringcrypto GO_EXTLINK_ENABLED=0
+ENV CGO_ENABLED=1 GOFIPS140=latest GO_EXTLINK_ENABLED=0
```
kidgilson, line 8: Likewise here this means FIPS mode is on by default which I don't think was the intended case right?

.goreleaser.yml
kidgilson: https://go.dev/doc/security/fips140#the-gofips140-environment-variable
>latest is like off, but enables FIPS 140-3 mode by default.

PR #3571, opened 2025-07-16
Ephemeral/add label annotations and errors

m-wurtz:
## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test
- [ ] :building_construction: Build

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

PR #3576, opened 2025-07-21
Add support for NetworkPolicy, Service, PVC, and Pod to `pkg/ephemeral` appliers

m-wurtz:
## Change Overview
This pull request improves on the ephemeral package in a few ways.
1) Extends the return type of `Applier` to return an error.  All call-sites are within error returning functions which makes this a simple extension.
2) Add support to `Applier` for the following objects:
  - NetworkPolicy
  - PersistentVolumeClaim
  - Pod
  - Service
  
Adds on to @kidgilson's code originally over in https://github.com/kanisterio/kanister/pull/3571

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E


pkg/controllers/repositoryserver/utils.go
```diff
@@ -112,22 +112,29 @@ func getPodOverride(ctx context.Context, reconciler *RepositoryServerReconciler,
...
 
-func podOverrideSpecForCACertificate(podSpec corev1.PodSpec, podOverride map[string]interface{}) {
-	if volName, proceed := volumeMountSpecForName(podSpec, podOverride, customCACertName); proceed {
+func podOverrideSpecForCACertificate(podSpec corev1.PodSpec, podOverride map[string]interface{}) error {
+	volName, proceed, err := volumeMountSpecForName(podSpec, podOverride, customCACertName)
+	if err != nil {
+		return err
+	}
+	if proceed {
 		volumeSpecForName(podSpec, podOverride, volName)
```
fmilichovsky, line 127: Do we wanna handle the error from this?

pkg/controllers/repositoryserver/utils.go
```diff
@@ -112,22 +112,29 @@ func getPodOverride(ctx context.Context, reconciler *RepositoryServerReconciler,
...
 	}
+	return nil
 }
 
 // volumeMountSpecForName adds a container spec to the override map
 // if the pod spec's volumeMount's SubPath matches with a given certificate name.
 // The container spec will include the matching volumeMount.
-func volumeMountSpecForName(podSpec corev1.PodSpec, podOverride map[string]interface{}, certName string) (string, bool) {
+func volumeMountSpecForName(podSpec corev1.PodSpec, podOverride map[string]interface{}, certName string) (string, bool, error) {
```
fmilichovsky: Have you deduced what the bool return value represents exactly while modifying this?

It's not totally obvious to me and multiple primitive returns like this are always kinda hard to figure out without jumping into the implementation. So I'm wondering if we could boy-scout improve this and name the returns for understandability while we're here.

pkg/validate/validate.go
```diff
@@ -372,3 +375,26 @@ func osSecretFromProfile(ctx context.Context, pType objectstore.ProviderType, p
...
+
+func ValidateLabels(labels map[string]string) error {
+	for k, v := range labels {
+		if errs := validation.IsQualifiedName(k); len(errs) > 0 {
+			return errkit.New(fmt.Sprintf("label key '%s' failed validation. %s", k, errs))
+		}
+
+		if errs := validation.IsValidLabelValue(v); len(errs) > 0 {
+			return errkit.New(fmt.Sprintf("label value '%s' failed validation. %s", v, errs))
+		}
```
fmilichovsky, line 387: optional: I could see some value in processing the full set and `errors.Join`ing all the problems at the end - so if you happen to have multiple validation failures, you don't resolve them one by one only to discover the next bad one on a separate run...

pkg/validate/validate.go
```diff
@@ -372,3 +375,26 @@ func osSecretFromProfile(ctx context.Context, pType objectstore.ProviderType, p
...
+	}
+	return nil
+}
+
+func ValidateAnnotations(annotations map[string]string) error {
+	for k := range annotations {
+		if errs := validation.IsQualifiedName(k); len(errs) > 0 {
+			return errkit.New(fmt.Sprintf("annotation key '%s' failed validation. %s", k, errs))
```
fmilichovsky, line 395: optional: Same here if we decide to `errors.Join` every validation issue.

fmilichovsky (approved):
Looking good overall, a couple optional suggestions.

pkg/controllers/repositoryserver/utils.go
m-wurtz: volumeSpecForName does not have an error to be returned

pkg/controllers/repositoryserver/utils.go
m-wurtz: I agree that it is confusing, I think naming the variables is a good idea.

pkg/validate/validate.go
m-wurtz: I looked around the repo and it consistently follows the pattern of returning a single wrapped `errkit`. I think your idea is good but since `errors.Join` is not used elsewhere I think it would be inconsistent here and feel out of place.

pkg/controllers/repositoryserver/utils.go
fmilichovsky: Ah, I think I was hallucinating the function signature/looking somewhere else.

pkg/validate/validate.go
fmilichovsky: Yep, sounds reasonable, better stay consistent.

pkg/ephemeral/annotations_test.go
```diff
@@ -0,0 +1,176 @@
...
+		"acs.amadeus.com/securityZone": "app",
+		"app.kubernetes.io/component":  "kasten",
+	}
+
+	var registeredAppliers ephemeral.ApplierList[kube.PodOptions]
+	set := ephemeral.AnnotationsFromEnvVar(annotationsEnvVar)
+	registeredAppliers.Register(set.PodOptions)
+
+	// OS environment variable set
+	err := os.Setenv(annotationsEnvVar, annotationsEnvVarValues)
```
ryanmt: While this would not be consistent w/ the rest of the the codebase, adopting `t.Setenv` instead of this `os.Setenv/defer` pattern would greatly improve the readability of these tests.

IMO that sort of sweeping change is probably too much to adopt in a single commit/PR across an entire codebase which means it has to start somewhere and this single file being consistent on that pattern would be a good place to start. 

Also as currently implemented this does not clean up on failure (because the defer is conditional because `Assert` is going to exit). That's not really a problem since it is just modifying the envvars available to the test runner which is exiting at that point.


Yeah... you don't get to assert `nil` on that error but that error is always nil on most platforms.  (ref: https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/syscall/env_unix.go, which applies to ~ all things !windows AFAICT)

m-wurtz: I agree with you that t.Setenv would be a better pattern for this suite of tests. If I am going to break the standard already set in the repo I would probably want a maintainer of kanister to weigh in on what they would prefer.

hairyhum: What is `t` in that context? I don't see any. `check` library does not seem to provide a `SetEnv` function.

m-wurtz: They are referring to using `testing` here instead of `check`. https://pkg.go.dev/testing#T.Setenv

m-wurtz:
Converting back to a draft as this is no longer blocking Alison's AFI work. Will add back in the networkpolicies, services, and PVC labelling.

pkg/ephemeral/annotations_test.go
ryanmt: 🤦 Yeah, sorry. I've been adopting `quicktest` and didn't think check was missing something like this since most testing tooling seems to be adopting the pattern.  `testing.T` definitely implements it but since check doesn't and the consistency would suffer it wouldn't be a great win here.

ryanmt:
The functionality has been manually confirmed in k10. Marking as ready for review again.

ryanmt:
The changes required to make this work downstream in K10 are staged here: https://github.com/kastenhq/k10/pull/32354

ryanmt:
Smoke tests and k10 tests are passing against this change here: https://github.com/kastenhq/k10/pull/32354

PR #3549, opened 2025-07-09
BugFix: improved environment variables for testing

varuni-dev:
## Change Overview
Previously, certain tests (e.g., DataSuite in data_test.go) were skipped during make test execution because they required specific environment variables to be set. This led to incomplete test coverage.

This change refactors the handling of environment variables to ensure they are set as needed during test execution. As a result, all tests, including those previously skipped, now run as part of the standard `make test` workflow.

## Pull request type

Please check the type of change your PR introduces:
- [ ] :construction: Work in Progress
- [ ] :rainbow: Refactoring (no functional changes, no api changes)
- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :world_map: Documentation
- [ ] :robot: Test
- [ ] :building_construction: Build

## Issues 

- fixes #3531 

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

**1. Run tests:**
Execute the test suite using the standard build command:
```
make test
```

**2. Verify Output:**
- Confirm that all test suites are executed, including the one that skipped previously.
- Confirm all tests pass with no failures or errors.

PR #3579, opened 2025-07-23
Refactor tests to work on GCloud and AWS env

varuni-dev:
## Change Overview

This PR refactors existing tests to ensure compatibility with both GCloud (GKE) and AWS environments
Changes-
 - Introduced conditionals or abstractions where necessary to support both cloud providers
 - Minor cleanups and structure alignment

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues 

- fixes #issue-number 

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

**Kanister**
1. Run tests:
Execute the test suite using the standard build command:

``
make test
``

2. Verify Output:

Confirm that all test suites are executed, including the one that skipped previously.
Confirm all tests pass with no failures or errors.

**K10-Kanister**
1. Run tests:
Execute the test suite using the standard build command:

````
./build.sh kube_init <gke-cluster-name> -e <gke environment>

./build.sh codefresh_go_test_kanister <gke-cluster-name> -e <gke environment>
````

2. Verify Output:

Confirm that all test suites are executed.
Confirm all tests pass with no failures or errors.

### DIFF
--- a/pkg/blockstorage/awsebs/awsebs.go
+++ b/pkg/blockstorage/awsebs/awsebs.go
@@ -27,7 +27,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/jpillora/backoff"
 	"github.com/kanisterio/errkit"
 
@@ -46,15 +45,14 @@ var _ zone.Mapper = (*EbsStorage)(nil)
 
 // EbsStorage implements blockstorage.Provider
 type EbsStorage struct {
-	Ec2Cli     *EC2
-	Role       string
-	config     *aws.Config
-	newEC2Func func(string, *aws.Config) (*EC2, error)
+	Ec2Cli *EC2
+	Role   string
+	config *aws.Config
 }
 
 // EC2 is kasten's wrapper around ec2.EC2 structs
 type EC2 struct {
-	ec2iface.EC2API
+	*ec2.EC2
 	DryRun bool
 }
 
@@ -79,10 +77,9 @@ func NewProvider(ctx context.Context, config map[string]string) (blockstorage.Pr
 	}
 
 	return &EbsStorage{
-		Ec2Cli:     ec2Cli,
-		Role:       config[awsconfig.ConfigRole],
-		config:     awsConfig,
-		newEC2Func: newEC2Client,
+		Ec2Cli: ec2Cli,
+		Role:   config[awsconfig.ConfigRole],
+		config: awsConfig,
 	}, nil
 }
 
@@ -96,7 +93,7 @@ func newEC2Client(awsRegion string, config *aws.Config) (*EC2, error) {
 		return nil, errkit.Wrap(err, "Failed to create session for EBS")
 	}
 	conf := config.WithMaxRetries(maxRetries).WithRegion(awsRegion).WithCredentials(config.Credentials)
-	return &EC2{EC2API: ec2.New(s, conf)}, nil
+	return &EC2{EC2: ec2.New(s, conf)}, nil
 }
 
 // VolumeCreate is part of blockstorage.Provider
@@ -247,7 +244,7 @@ func (s *EbsStorage) snapshotParse(ctx context.Context, snap *ec2.Snapshot) *blo
 		Type:         s.Type(),
 		Encrypted:    aws.BoolValue(snap.Encrypted),
 		SizeInBytes:  aws.Int64Value(snap.VolumeSize) * blockstorage.BytesInGi,
-		Region:       aws.StringValue(s.Ec2Cli.EC2API.(*ec2.EC2).Config.Region),
+		Region:       aws.StringValue(s.Ec2Cli.Config.Region),
 		Volume:       vol,
 		CreationTime: blockstorage.TimeStamp(aws.TimeValue(snap.StartTime)),
 	}
@@ -285,7 +282,7 @@ func (s *EbsStorage) SnapshotCopy(ctx context.Context, from, to blockstorage.Sna
 		return nil, errkit.New(fmt.Sprintf("Snapshot %v destination ID must be empty", to))
 	}
 	// Copy operation must be initiated from the destination region.
-	ec2Cli, err := newEC2Client(to.Region, &s.Ec2Cli.EC2API.(*ec2.EC2).Config)
+	ec2Cli, err := newEC2Client(to.Region, &s.Ec2Cli.Config)
 	if err != nil {
 		return nil, errkit.Wrap(err, "Could not get EC2 client")
 	}
@@ -296,7 +293,7 @@ func (s *EbsStorage) SnapshotCopy(ctx context.Context, from, to blockstorage.Sna
 		si := ec2.CopySnapshotInput{
 			SourceSnapshotId:  aws.String(from.ID),
 			SourceRegion:      aws.String(from.Region),
-			DestinationRegion: s.Ec2Cli.EC2API.(*ec2.EC2).Config.Region,
+			DestinationRegion: s.Ec2Cli.Config.Region,
 		}
 		rq, _ := ec2Cli.CopySnapshotRequest(&si)
 		su, err2 := rq.Presign(120 * time.Minute)
@@ -321,7 +318,7 @@ func (s *EbsStorage) SnapshotCopy(ctx context.Context, from, to blockstorage.Sna
 		Description:       aws.String("Copy of " + from.ID),
 		SourceSnapshotId:  aws.String(from.ID),
 		SourceRegion:      aws.String(from.Region),
-		DestinationRegion: s.Ec2Cli.EC2API.(*ec2.EC2).Config.Region,
+		DestinationRegion: s.Ec2Cli.Config.Region,
 		Encrypted:         encrypted,
 		PresignedUrl:      presignedURL,
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: build/test.sh [hunk 1]
```diff
@@ -44,6 +44,7 @@ check_dependencies() {
         export LOCATION_CLUSTER_ENDPOINT="http://minio.minio.svc.cluster.local:9000"
         export TEST_REPOSITORY_ENCRYPTION_KEY="testKopiaRepoPassword"
         unset AWS_SESSION_TOKEN
+        export USE_MINIO=1
```
PrasadG193:
```suggestion
        export USE_MINIO="true"
```

pkg/blockstorage/awsebs/awsebs.go
```diff
@@ -238,14 +239,15 @@ func (s *EbsStorage) snapshotParse(ctx context.Context, snap *ec2.Snapshot) *blo
...
+
 	// TODO: fix getting region from zone
 	return &blockstorage.Snapshot{
 		ID:           aws.StringValue(snap.SnapshotId),
 		Tags:         tags,
 		Type:         s.Type(),
 		Encrypted:    aws.BoolValue(snap.Encrypted),
 		SizeInBytes:  aws.Int64Value(snap.VolumeSize) * blockstorage.BytesInGi,
-		Region:       aws.StringValue(s.config.Region),
+		Region:       aws.StringValue(s.Ec2Cli.EC2API.(*ec2.EC2).Config.Region),
```
PrasadG193:
```suggestion
		Region:       aws.StringValue(s.Ec2Cli.Config.Region),
```
Wouldn't this work? Referring from the original changes from https://github.com/kanisterio/kanister/pull/3549/files

pkg/blockstorage/blockstorage_test.go
```diff
@@ -59,7 +63,21 @@ func (s *BlockStorageProviderSuite) SetUpSuite(c *check.C) {
...
+			"testtag":     "testtagvalue",
+			"SizeInBytes": int64(1024),
+			testNameKey:   testNameValue,
+		}
+	} else {
+		s.provider, err = getter.New().Get(s.storageType, config)
+		s.testData = map[string]any{
+			"testtag":     c.TestName(),
+			"SizeInBytes": int64(1 * blockstorage.BytesInGi),
+			testNameKey:   c.TestName(),
```
PrasadG193, line 78: @varuni-dev can we not  use same values for tags and size that we used for mocked provider? We don't have to maintain 2 separate test data then

pkg/blockstorage/blockstorage_test.go
```diff
@@ -184,7 +202,10 @@ func (s *BlockStorageProviderSuite) TestSnapshotCopy(c *check.C) {
 	log.Print("Snapshot copied", field.M{"FromSnapshotID": srcSnapshot.ID, "ToSnapshotID": snap.ID})
 
 	config := s.getConfig(c, dstSnapshot.Region)
-	provider, err := mockblockstorage.NewGetter().Get(s.storageType, config)
+	provider, err := getter.New().Get(s.storageType, config)
+	if useMinio, ok := os.LookupEnv("USE_MINIO"); ok && useMinio == "1" {
+		provider, err = mockblockstorage.NewGetter().Get(s.storageType, config)
+	}
 	c.Assert(err, check.IsNil)
```
PrasadG193:
```suggestion
        var provider blockstorage.Provider
	if useMinio, ok := os.LookupEnv("USE_MINIO"); ok && useMinio == "1" {
		provider, err = mockblockstorage.NewGetter().Get(s.storageType, config)
		c.Assert(err, check.IsNil)
	} else {
	        provider, err := getter.New().Get(s.storageType, config)
	        c.Assert(err, check.IsNil)
	}
```

pkg/blockstorage/awsebs/awsebs.go
varuni-dev: no as it is an interface api (i.e ec2iface.EC2API) which directly does not expose config. thus had to type assert it.

pkg/blockstorage/blockstorage_test.go
varuni-dev: The mock structs are used by multiple test cases, so I didn’t modify the existing mock values and used separate set for this case. We can refactor it later if needed

build/test.sh
varuni-dev: updated the suggested value

pkg/blockstorage/blockstorage_test.go
varuni-dev: updated suggestion

pkg/blockstorage/awsebs/awsebs.go
PrasadG193: Its not a best practice to type assert interfaces directly. It might cause panic.
```
    if e, ok := s.Ec2Cli.EC2API.(*ec2.EC2); ok {
        region = e.Config.Region
    } else {
        // handle if assertion not possible
    }
```
We can first check if its assertion is possible. But overall I think we should look for better way. Can we not revert it to the original way of handling like in https://github.com/kanisterio/kanister/pull/3549/files?

PrasadG193: Can we please revert back the interface change? We need to find better way to do this.

varuni-dev: have revert back the interface changes.

pkg/testutil/fixture.go
```diff
@@ -86,11 +86,13 @@ func ObjectStoreProfileOrSkip(c *check.C, osType objectstore.ProviderType, locat
...
-		if endpoint, ok := os.LookupEnv("LOCATION_ENDPOINT"); ok && location.Endpoint == "" {
-			location.Endpoint = endpoint
+		if useMinio, ok := os.LookupEnv("USE_MINIO"); ok && useMinio == "true" {
+			if region, ok := os.LookupEnv(awsconfig.Region); ok && location.Region == "" {
+				location.Region = region
+			}
+			if endpoint, ok := os.LookupEnv("LOCATION_ENDPOINT"); ok && location.Endpoint == "" {
+				location.Endpoint = endpoint
+			}
 		}
```
PrasadG193, line 96: We also want to set it for S3, right?

varuni-dev: no it takes default values in case of s3.

PrasadG193: @varuni-dev what if we use different S3-compatible store? We won't be able to override the values to use that. If we want to use different Region and Endpoint for the store

pkg/blockstorage/blockstorage_test.go
```diff
@@ -48,6 +51,7 @@ type BlockStorageProviderSuite struct {
 	volumes       []*blockstorage.Volume
 	snapshots     []*blockstorage.Snapshot
 	args          map[string]string
+	testData      map[string]any
```
PrasadG193, line 54:
```suggestion
	testData      map[string]string
```
can we use string as value? We don't have to typecast then while validating each time

varuni-dev: This includes multiple test values such as tags and size which use different data types.

pkg/blockstorage/blockstorage_test.go
PrasadG193: Can you please file an issue for that so that we can track it?

varuni-dev: https://github.com/kanisterio/kanister/issues/3580

pkg/blockstorage/blockstorage_test.go
```diff
@@ -59,7 +63,21 @@ func (s *BlockStorageProviderSuite) SetUpSuite(c *check.C) {
...
+		s.provider, err = mockblockstorage.NewGetter().Get(s.storageType, config)
+		s.testData = map[string]any{
+			"testtag":     "testtagvalue",
+			"SizeInBytes": int64(1024),
+			testNameKey:   testNameValue,
+		}
+	} else {
+		s.provider, err = getter.New().Get(s.storageType, config)
+		s.testData = map[string]any{
+			"testtag":     c.TestName(),
```
ankitjain-veeam, line 76: Might be worth moving these keys as constants, can be done in a followup.

pkg/function/data_test.go
```diff
@@ -49,7 +49,7 @@ type DataSuite struct {
 }
 
 const (
-	testBucketName = "tests.kanister.io"
+	testBucketName = "kio-store-tests"
```
ankitjain-veeam, line 52: Do we need to create this bucket in MinIO or we are just using mock data?

pkg/testutil/fixture.go
```diff
@@ -86,11 +86,13 @@ func ObjectStoreProfileOrSkip(c *check.C, osType objectstore.ProviderType, locat
...
 	case objectstore.ProviderTypeS3:
-		if region, ok := os.LookupEnv(awsconfig.Region); ok && location.Region == "" {
-			location.Region = region
-		}
-		if endpoint, ok := os.LookupEnv("LOCATION_ENDPOINT"); ok && location.Endpoint == "" {
-			location.Endpoint = endpoint
+		if useMinio, ok := os.LookupEnv("USE_MINIO"); ok && useMinio == "true" {
+			if region, ok := os.LookupEnv(awsconfig.Region); ok && location.Region == "" {
+				location.Region = region
+			}
```
ankitjain-veeam, line 92: Isn't region common irrespective of if it's minio or S3?

ankitjain-veeam (approved):
Added minor questions, suggestions. Can be addressed in a follow-up PR if required.

pkg/testutil/fixture.go
varuni-dev: region is common but its been set from aws creds but as minio does not delivers region in creds had to explicitly set it here.

pkg/function/data_test.go
varuni-dev: No. As `kio-store-tests` is common bucket available for testing in diff env. We are mocking data only for minio

pkg/blockstorage/blockstorage_test.go
varuni-dev: have created another issue regarding the same
https://github.com/kanisterio/kanister/issues/3580

PR #3717, opened 2025-11-06
build(release): Fix Kanister Version Bump Pipeline

timothymamo:
## Change Overview

As per title, fix to allow for multiple version bump (currently set to previous 2 versions).

## Pull request type

## Issues 

- fixes #[K10-31450](https://kasten.atlassian.net/browse/K10-31450)

## Test Plan

- [x] :muscle: Manual

### DIFF
--- a/build/bump_version.sh
+++ b/build/bump_version.sh
@@ -43,7 +43,7 @@ main() {
         grep -F "${prev_array[$i]}" -Ir  "${pkgs[@]}" --exclude-dir={docs,mod,bin,html,frontend,tmp} --exclude=\*.sum --exclude=\*.mod | cut -d ':' -f 1 | uniq | xargs -r sed -ri "s/${prev_array[$i]}/${next//./\\.}/g" || continue
     done
 
-    # Modify the first instance of kanister_tools_version in docs/constants.ts as only the latest version of k10 uses the latest version of kanister
+    # Modify the first instance of kanister_tools_version in docs/constants.ts
     if [ ${#pkgs[@]} -eq 1 ]; then
       file_count=$(find "${pkgs[@]}/docs" -maxdepth 1 -name "constants.ts" | wc -l)
       if [ $file_count -eq 1 ]; then


build/bump_version.sh
```diff
@@ -20,21 +20,38 @@ set -o xtrace
...
 
 usage() {
-    echo ./build/bump_version.sh previous_version release_version
+    echo ./build/bump_version.sh array_length previous_versions release_version
     exit 1
 }
 
 main() {
+    local array_size=${1:?"$(usage)"}; shift
     local prev=${1:?"$(usage)"}; shift
```
PrasadG193: Both are reading 1st CLI arg, what is the difference? Should prev read second CLI arg?

build/bump_version.sh
```diff
@@ -20,21 +20,38 @@ set -o xtrace
...
-    # -F matches for exact words, not regular expression (-E), that is what required here
-    grep -F "${prev}" -Ir  "${pkgs[@]}" --exclude-dir={mod,bin,html,frontend} --exclude=\*.sum --exclude=\*.mod | cut -d ':' -f 1 | uniq | xargs sed -ri "s/${prev}/${next//./\\.}/g"
+
+    for ((i=${array_size}-1; i>=0; i--)); do
+        echo "Processing version: ${prev[$i]}"
+        # -F matches for exact words, not regular expression (-E), that is what required here
+        grep -F "${prev[$i]}" -Ir  "${pkgs[@]}" --exclude-dir={docs,mod,bin,html,frontend,tmp} --exclude=\*.sum --exclude=\*.mod | cut -d ':' -f 1 | uniq | xargs -r sed -ri "s/${prev[$i]}/${next//./\\.}/g" || continue
+    done
+
+    # Modify the first instabnce of kanister_tools_version in docs/constants.ts
```
PrasadG193:
```suggestion
    # Modify the first instance of kanister_tools_version in docs/constants.ts
```

PrasadG193: Why do we want to replace only first instance? What is the reason? Maybe we should add the reason to the comment as well to avoid confusion.

build/bump_version.sh
timothymamo: The [shift](https://www.man7.org/linux/man-pages/man1/shift.1p.html) command shifts the inputs, hence we keep using the `$1` paramter.

build/bump_version.sh
```diff
@@ -20,21 +20,38 @@ set -o xtrace
 
 
 usage() {
-    echo ./build/bump_version.sh previous_version release_version
+    echo ./build/bump_version.sh array_length previous_versions release_version
```
PrasadG193: Can we accept comma separated versions instead accepting size and values as a two different inputs?

`./build/bump_version.sh 0.114.0,0.115.0 0.116.0`

build/bump_version.sh
```diff
@@ -20,21 +20,36 @@ set -o xtrace
...
             pkgs=( "$@" )
     fi
-    # -F matches for exact words, not regular expression (-E), that is what required here
-    grep -F "${prev}" -Ir  "${pkgs[@]}" --exclude-dir={mod,bin,html,frontend} --exclude=\*.sum --exclude=\*.mod | cut -d ':' -f 1 | uniq | xargs sed -ri "s/${prev}/${next//./\\.}/g"
+
+    for ((i=${array_size}-1; i>=0; i--)); do
+        echo "Processing version: ${prev_array[$i]}"
+        # -F matches for exact words, not regular expression (-E), that is what required here
+        grep -F "${prev_array[$i]}" -Ir  "${pkgs[@]}" --exclude-dir={docs,mod,bin,html,frontend,tmp} --exclude=\*.sum --exclude=\*.mod | cut -d ':' -f 1 | uniq | xargs -r sed -ri "s/${prev_array[$i]}/${next//./\\.}/g" || continue
+    done
```
PrasadG193, line 44: Can we simply do
```suggestion
    for release in "${prev_array[@]}"; do
        # -F matches for exact words, not regular expression (-E), that is what required here
        grep -F "${release}" -Ir  "${pkgs[@]}" --exclude-dir={docs,mod,bin,html,frontend,tmp} --exclude=\*.sum --exclude=\*.mod | cut -d ':' -f 1 | uniq | xargs -r sed -ri "s/${release}/${next//./\\.}/g" || continue
    done
```

timothymamo: This has been tested to work as is. I'd rather not change it.

PR #3791, opened 2025-12-11
refactor: update to get custom tag for images

varuni-dev:
## Change Overview

This pull request adds support for custom image tag in blueprints with a fallback to the default dev tag when the image doesn't exist.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes [K10-29294](https://kasten.atlassian.net/browse/K10-29294)

## Test Plan

- [x] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/app/bp_test.go
+++ b/pkg/app/bp_test.go
@@ -133,7 +133,7 @@ func (bs *BlueprintSuite) TestUpdateImageTags(c *check.C) {
 			},
 		},
 	} {
-		updateImageTags(bp, "")
+		updateImageTags(bp, "v9.99.9-dev")
 		validateImageTags(c, bp)
 	}
 }


pkg/app/bp.go
```diff
@@ -44,6 +45,7 @@ type AppBlueprint struct {
 	Path           string
 	UseDevImages   bool
 	readFromBPRepo bool
+	ImageTag       string
```
PrasadG193: I was wondering if we need `UseDevImages` anymore. We can pass dev image tag as a ImageTag arg so I believe we can get rid of `UseDevImages`.

pkg/app/bp.go
```diff
@@ -115,7 +118,18 @@ func updateImageTags(bp *crv1alpha1.Blueprint) {
 
 			if strings.HasPrefix(imageStr, imagePrefix) {
 				// ghcr.io/kanisterio/tools:v0.xx.x => ghcr.io/kanisterio/tools:v9.99.9-dev
-				phase.Args["image"] = fmt.Sprintf("%s:v9.99.9-dev", strings.Split(imageStr, ":")[0])
+				baseImage := strings.Split(imageStr, ":")[0]
+				if imageTag == "" {
+					imageTag = "v9.99.9-dev"
```
PrasadG193: Let's use const for dev image tag `v9.99.9-dev`

pkg/app/bp.go
```diff
@@ -115,7 +118,18 @@ func updateImageTags(bp *crv1alpha1.Blueprint) {
...
+				}
+
+				updatedImage := fmt.Sprintf("%s:%s", baseImage, imageTag)
+				if imageTag != "v9.99.9-dev" && !imageExists(updatedImage) {
```
PrasadG193: We shouldn't check if image exist. Let it fail. This can mislead test - e.g tests passed for the image tag - even if its not valid. So reports incorrect result

pkg/app/bp.go
varuni-dev: Kanister integration tests still use the dev tag, so we need this to decide whether to use a commit-based tag or the dev tag.

pkg/app/bp.go
varuni-dev: sure

pkg/app/bp.go
PrasadG193: What I am suggesting, we can get rid of `UseDevImages` since we have the ability pass `ImageTag`? For Kanister integration test, we can pass `ImageTag=v9.99.9-dev`

varuni-dev: make sense, will update it accordingly

ankitjain-veeam (approved):
LGTM

PR #3803, opened 2025-12-23
Execute integration test in batch of 3 apps per interation

### LINKED ISSUE #3779: [BUG] CI - Integration tests failing due to minio storate limit issue
**Describe the bug**

We use minio as a storage in integration test. We started facing issue in integration test about hitting storage limit in minio.
```
Error: failed to write contents to bucket 'tests.kanister.io': PutObject, putting object: XMinioStorageFull: Storage backend has reached its minimum free drive threshold. Please delete a few objects to proceed.
```

This issue is observed when integration tests are run on Github actions only. 

Ref: https://github.com/kanisterio/kanister/actions/runs/19924410699/job/57252030718

**Expected behavior**
The integration test for all the apps should be run without any issue.
We can also try to reduce the storage consumed during tests.

abdul-saqib:
## Change Overview

Enable mssql app in integration test.

Execute integration test in batch of 3 apps per interation

Resolve minio storage threshold reached due to hardware constraint on GitHub Action runner

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues

- fixes #3779

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [x] :green_heart: E2E

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -153,3 +153,4 @@ jobs:
       extra_tags: |
         ${{ needs.image_tags.outputs.tag_short }}
         ${{ needs.image_tags.outputs.tag_long }}
+


build/integration-test.sh
```diff
@@ -124,9 +122,25 @@ else
...
-go test -v ${TEST_OPTIONS} -check.f "${TEST_APPS}" -installsuffix "static" . -check.v
+INDEX=0
+while true; do
+    CONCURRENT_TEST_APPS="$(
+        echo "$TEST_APPS" \
+        | tr '|' '\n' \
+        | tail -n +$((INDEX * DOP + 1)) \
+        | head -n "$DOP" \
+        | paste -sd'|' -
+    )"
```
PrasadG193, line 153: I was wondering if we can simplify it with maintaining two maps/variables for TIER_0_APPS (for lightweight app) AND TIER_1_APPS (for heavyweight apps). TIER_0_APPS can be run concurrently and TIER_1_APPS can be run sequentially

abdul-saqib: Yes, we can do that but it would eventually increase the total runtime. The current implementation runs all the apps concurrently while still conserving space to avoid storage issues. While storage issues are still possible with the current setup they are less likely to occur.

PrasadG193: What exactly we are doing with this command, can you please explain? Also better to add comment as its not very straight-forward to understand.

abdul-saqib: It uses standard Unix utilities to select batches of apps from the pipe separated **TEST_APPS** list (^PostgreSQL$|^MySQL$|^MongoDB$|^MSSQL$).

On each loop iteration, it selects the next **3** apps because **DOP=3** based on the current INDEX and runs tests for only that batch in parallel. 

Since this logic runs inside a loop all apps are processed sequentially in controlled parallel batches.

Example:
```
If TEST_APPS = A|B|C|D|E|F|G and DOP = 3:

INDEX = 0 → runs A|B|C

INDEX = 1 → runs D|E|F

INDEX = 2 → runs G
```

I will add the comment on the top of it in script file.

.github/workflows/main.yaml
```diff
@@ -129,17 +129,7 @@ jobs:
...
-      tag_short: ${{ steps.image_tags.outputs.tag_short }}
-      tag_long: ${{ steps.image_tags.outputs.tag_long }}
-    steps:
-    - id: image_tags
-      env:
-        COMMIT_SHA: ${{ github.sha }}
-      run: |
-        echo "tag_short=short-commit-${COMMIT_SHA::12}" >> $GITHUB_OUTPUT
-        echo "tag_long=commit-${COMMIT_SHA}" >> $GITHUB_OUTPUT
+    uses: ./.github/workflows/image-tags.yaml
```
PrasadG193: Are these changes required?

abdul-saqib: No, these are not related, let me check if I commit those here mistakenly!!

PrasadG193 (approved):
LGTM

PR #4009, opened 2026-04-15
feat(function): add volumes arg to KubeTask

anand-veeam:
## Change Overview

Adds an optional `volumes` argument to the KubeTask function that accepts 
a map of PVC name → mount path. Each PVC is validated at execution time 
and mounted into the job pod automatically.

This enables Kanister blueprints to mount backup-csi-driver PVCs directly 
via blueprint args, without requiring users to inject volume mounts through 
podOverride.

**Backward compatible** — `volumes` is optional with a nil default. All 
existing blueprints that do not specify `volumes` behave identically to before.

Also updates the only other internal caller (`execDumpCommand` in 
`export_rds_snapshot_location.go`) to pass nil for the new parameter.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [ ] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/function/export_rds_snapshot_location.go
+++ b/pkg/function/export_rds_snapshot_location.go
@@ -381,7 +381,7 @@ func execDumpCommand(
 		}
 	}()
 
-	return kubeTask(ctx, cli, namespace, postgresToolsImage, command, injectPostgresSecrets(secretName), annotations, labels)
+	return kubeTask(ctx, cli, namespace, postgresToolsImage, command, nil, injectPostgresSecrets(secretName), annotations, labels)
 }
 
 func prepareCommand(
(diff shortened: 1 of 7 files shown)


pkg/function/export_rds_snapshot_location.go
```diff
@@ -381,7 +381,7 @@ func execDumpCommand(
 		}
 	}()
 
-	return kubeTask(ctx, cli, namespace, postgresToolsImage, command, injectPostgresSecrets(secretName), annotations, labels)
+	return kubeTask(ctx, cli, namespace, postgresToolsImage, command, nil, injectPostgresSecrets(secretName), annotations, labels)
```
PrasadG193, line 384: @anand-veeam  did you check `PrepareData` function? Its intended to be used for such use cases. https://docs.kanister.io/functions.html#preparedata
We don't need to modify KubeTask.

PR #4051, opened 2026-05-07
ci(Mergify): Update mergify config to unblock merge

PrasadG193:
This change has been made by @PrasadG193 from the Mergify Queue Rule Configurator.


r4rajat (approved):
LGTM

PR #4120, opened 2026-07-02
fix(secrets): accept Azure secrets using federated identity

vm-ashwinkolhatkar:
## Change Overview

Azure Workload Identity (federated identity) location profiles have no storage key and are marked in the credential secret with an `azure_federated_identity` field. `ValidateAzureCredentials` rejected these secrets as having an unknown field, and `ExtractAzureCredentials` unconditionally required a storage key, so any backup/export/restore referencing such a profile failed before any phase ran.

This adds `AzureFederatedIdentity` as a recognized field: it's counted in the allow-list during validation, and `ExtractAzureCredentials` only requires a storage key when the field isn't set to `"true"`. Existing behavior for non-federated secrets is unchanged.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [x] :bug: Bugfix
- [ ] :sunflower: Feature
- [ ] :robot: Test

## Issues

- Addresses K10-37697 (Kasten K10 Jira, no corresponding kanister GitHub issue)

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

Added table-driven test cases in `pkg/secrets/azure_test.go`:
- Secret with `azure_storage_account_id` + `azure_federated_identity: "true"` and no key succeeds.
- Secret with `azure_federated_identity` present but not `"true"` still requires the storage key.
- Existing unrelated-extra-field rejection case is unchanged.


pkg/secrets/azure.go
```diff
@@ -87,7 +94,8 @@ func ExtractAzureCredentials(secret *corev1.Secret) (*objectstore.SecretAzure, e
 	if envName, ok := secret.Data[AzureStorageEnvironment]; ok {
 		azSecret.EnvironmentName = string(envName)
 	}
-	if azSecret.StorageAccount == "" || azSecret.StorageKey == "" {
+	federatedIdentity := string(secret.Data[AzureFederatedIdentity]) == "true"
```
PrasadG193:
```suggestion
	federatedIdentity := strings.EqualFold(string(secret.Data[AzureFederatedIdentity]), "true")
```
To avoid cases where "true" == "True"

PrasadG193 (approved):
LGTM, left minor suggestion about case sensitive checks

PR #4148, opened 2026-07-28
K10-38020: Azure Federated Identity support for kando location commands

ranveer-avhad-veeam:
## Change Overview

`kando location push/pull/delete` to an Azure **Workload (Federated) Identity** location profile
failed, because the object-store path assumes a shared account key. This adds SAS-based auth for
keyless Azure profiles, and a generic hook so a controller can mint and inject the credential.

- **objectstore**: `SecretAzure` gains `FederatedIdentity`/`SASToken`; `MintAzureUserDelegationSAS`
  mints a short-lived, **container-scoped, least-privilege** user-delegation SAS
  (`AzureSASPermissions`, cached per account/container/permission); `azureConfig` forwards the SAS
  into the stow driver.
- **secrets**: `ExtractAzureCredentials` propagates the federated-identity marker.
- **location**: for a federated profile, kando reads the SAS from `AZURE_SAS_TOKEN_FILE` (a path,
  never the token value).
- **ephemeral**: new generic `PodCredentialInjector` hook (`RegisterPodCredentialInjector` /
  `InjectPodCredentials`), invoked from `kubeTask()` at pod-build time with cleanup. **It is a no-op
  unless a consumer registers an injector**, so behaviour is unchanged for existing users; the RDS
  caller of `kubeTask` is untouched.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- Jira: K10-38020 (no GitHub issue)

## Test Plan

- [x] :muscle: Manual — verified on an AKS Workload-Identity cluster: push/pull/delete succeed with
  per-operation SAS (`racw` / `r` / `rdl`); requires the stow container-probe fix.
- [x] :zap: Unit test — `pkg/ephemeral/credinject_test.go`, `pkg/objectstore/azure_sas_test.go`,
  `pkg/secrets/azure_test.go`

Files: pkg/ephemeral/credinject.go(+test), pkg/function/kube_task.go, pkg/location/location.go, pkg/objectstore/{azure_sas.go(+test),models.go,objectstore.go}, pkg/secrets/azure.go(+test), go.mod/go.sum. Depends on the stow PR (go.mod pins the fixed stow).


pkg/objectstore/azure_sas.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
+	"github.com/kanisterio/errkit"
+)
+
+const (
+	// azureSASValidity is the lifetime of a minted user-delegation SAS token.
+	azureSASValidity = 24 * time.Hour
```
alisonb-veeam: Could we do something like 2 hours here? What are the longest export times we need to consider for this?

We are calling the injector from `kanister-svc` so right now the tokens will be useable against azure blob storage for 24hrs.

Someone with `get secrets` in the namespace has a short window to read the Secret (the pod's lifetime), but whatever they read keeps working for up to 24h afterwards. 

Deleting the Secret doesn't revoke anything, and a user-delegation SAS can't be revoked individually and the cache hands that same token to other pods, in other namespaces, for up to 23h.

ranveer-avhad-veeam: Agreed on both the TTL and the cache blast radius:

TTL: I'll shorten azureSASValidity from 24h. The only constraint is it must outlast the longest single kando operation + the refresh margin, or a large in-flight location push could have its SAS expire mid-upload. Do we have a worst-case single-operation duration to support? If a couple of hours covers it I'll set ~2h (well under the 7-day user-delegation-key ceiling).
Cache: you're right, the process cache (keyed account/container/perms) hands the same token to other pods/namespaces for its lifetime, so "per-pod" is really "per-key for the TTL." With no per-token revocation for user-delegation SAS, the TTL is effectively the only control. Shortening it is the cheapest mitigation; if we want to go further I can drop the cross-pod cache and mint per-invocation (tighter blast radius, more GetUserDelegationKey calls).
Proposal: shorten the TTL (to cover the max op duration, hence the question) and keep the cache, documenting that revocation isn't possible so TTL is the bound. LMK the duration ceiling and whether you'd prefer per-invocation minting.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.26.1
 
 replace (
 	github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.1
-	github.com/graymeta/stow => github.com/kastenhq/stow v0.2.6-kasten.1.0.20250930185039-ccb63ada4a21
+	github.com/graymeta/stow => github.com/ranveer-avhad-veeam/stow v0.0.0-20260728060125-ec5cf8ffe01b
```
alka-infracloud: Nit: Update the reference once stow Pr is merged.

pkg/function/kube_task.go
```diff
@@ -59,6 +59,22 @@ func (*kubeTaskFunc) Name() string {
...
+type kubeTaskOption func(*kubeTaskConfig)
+
+type kubeTaskConfig struct {
+	// tp, when set, enables running the registered pod credential injectors
+	// (e.g. minting a short-lived cloud credential into an ephemeral Secret and
+	// projecting it into the pod) against this pod's options.
+	tp *param.TemplateParams
+}
+
+// withPodCredentialInjection runs the registered ephemeral pod credential
```
alka-infracloud, line 72: only KubeTask is wired. withPodCredentialInjection(tp) is added solely in kubeTaskFunc.Exec. KubeExec runs kando inside a pre-existing application pod where injection is impossible, and the functions that build pods directly aren't covered. The design doc promises "kando location push/pull/delete"; this delivers "kando run via KubeTask". Please state the supported surface explicitly.

pkg/objectstore/azure_sas.go
ranveer-avhad-veeam: Reworked this around the TB-scale export requirement: a single kando location push can be multi-terabyte, and the SAS must outlast the whole upload or it expires mid-stream. Worst-case at a conservative ~50 MB/s: 5 TiB ≈ 29h; with margin I set azureSASValidity = 48h (covers ~8 TiB @50 MB/s, ~4 TiB even @25 MB/s; well under the 7-day user-delegation-key ceiling). On the blast radius you flagged, rather than pair a long TTL with the shared process cache (which would hand one long-lived token to pods across namespaces), I dropped the cross-pod cache and now mint per invocation, so each token is bound to a single worker pod's operation. That's one extra GetUserDelegationCredential per pod-build; if it ever matters we can cache just the user-delegation key (safe for local signing) and still sign a distinct per-pod SAS. Since a user-delegation SAS can't be revoked individually, per-invocation minting keeps the token's audience as tight as possible.

go.mod
ranveer-avhad-veeam: Sure

pkg/function/kube_task.go
ranveer-avhad-veeam: Done

alisonb-veeam:
LGTM - but I'll let someone form kanister :) Thanks for the good work on this!

alka-infracloud:
LGTM

alka-infracloud (approved):

mattbator (commented):
RN approval

mattbator (approved):

PR #4226, opened 2026-09-11
feat:support for conditional phase execution

ap-kulkarni:
## Change Overview

Add support to execute a phase based on condition. This PR introduces support for a new field `if` that can be added to a phase. The `if` field supports an expression in go template format. The expression must be conditional. i.e. it should evaluated to either `true` or `false`. If the expression evaluates to `true` then the phase will be executed. If it evaluates to `false`, the phase will be skipped from execution and status will reflect accordingly. To preserve backward compatibility, if a phase doesn't have `if` field, it will be executed.
Example of phase with `if` condition:
```
kind: Blueprint
metadata:
  name: time-log-bp
  namespace: kanister
actions:
  backup:
    phases:
    - func: KubeExec
      name: backupToS3
      if: '{{ eq .Deployment.Namespace "time-logger-ns" }}'
      args:
        namespace: "{{ .Deployment.Namespace }}"
        pod: "{{ index .Deployment.Pods 0 }}"
        container: test-container
        command:
          - sh
          - -c
          - echo /var/log/time.log
```

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [x] :sunflower: Feature
- [ ] :robot: Test

## Issues 

- fixes #issue-number

## Test Plan

- [ ] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

### DIFF
--- a/pkg/apis/cr/v1alpha1/types.go
+++ b/pkg/apis/cr/v1alpha1/types.go
@@ -190,6 +190,8 @@ const (
 	StateFailed State = "failed"
 	// StateComplete means this action or phase finished successfully.
 	StateComplete State = "complete"
+	// StateSkipped means this action or phase was skipped due to condition not met
+	StateSkipped State = "skipped"
 )
 
 // Error represents an error that occurred when executing an actionset.
@@ -208,6 +210,8 @@ type Phase struct {
 	Output map[string]interface{} `json:"output,omitempty"`
 	// Progress represents the phase execution progress.
 	Progress PhaseProgress `json:"progress,omitempty"`
+	// Reason shows why a phase skipped. It doesn't have significance in other phase states yet.
+	Reason string `json:"reason,omitempty"`
 }
 
 // PhaseProgress represents the execution state of the phase.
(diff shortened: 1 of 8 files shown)


pkg/apis/cr/v1alpha1/types.go
```diff
@@ -190,6 +190,8 @@ const (
 	StateFailed State = "failed"
 	// StateComplete means this action or phase finished successfully.
 	StateComplete State = "complete"
+	// StateSkipped means this action or phase was skipped due to condition not met
+	StateSkipped State = "skipped"
```
alka-infracloud, line 194: Please add ticket link in the description. Why we need to support if block here?

pkg/phase.go
```diff
@@ -49,6 +60,13 @@ func (p *Phase) Name() string {
```
alka-infracloud, line 60: Shouldn't we add skipped case here as well https://github.com/kanisterio/kanister/blob/master/pkg/progress/action.go#L143. 
Please identify if there is any other place as well where we need these kind of checks.

pkg/apis/cr/v1alpha1/types.go
alka-infracloud: if this is something we need for k10 better to create corresponding ticket in kanister and add here.

PR #4229, opened 2026-09-15
Fix PodSuite.TestPod label count assertion on Kubernetes 1.35

### LINKED ISSUE #4228: PodSuite.TestPod fails on Kubernetes 1.35: exact label count breaks with PodTopologyLabels admission
`pkg/kube/pod_test.go` asserts `len(pod.Labels) == len(po.Labels)+1`. Kubernetes 1.35 enables the `PodTopologyLabelsAdmission` feature gate by default; the admission plugin copies `topology.kubernetes.io/zone` and `topology.kubernetes.io/region` from the node onto pods created with `spec.nodeName`. `TestPod` creates such pods, so it sees two extra labels and fails:

```
pod_test.go:241:
    c.Check(len(pod.ObjectMeta.Labels), check.Equals, len(po.Labels)+1)
... obtained int = 3
... expected int = 1
```

Reproduced on kind v1.35.8 with zone/region labels on the node; passes on kind v1.34.11. The same pod created with `spec.nodeName` gets `topology.kubernetes.io/region` and `topology.kubernetes.io/zone` on 1.35.8 and nothing extra on 1.34.11.

The lines that follow already assert the `createdBy` label and every caller-supplied label, so the count check can go.

hr2822:
## Change Overview

Assert the exact pod label count in `PodSuite.TestPod` on the spec Kanister builds (`GetPodObjectFromPodOptions`) instead of on the created pod. Kubernetes 1.35 admission adds topology labels to pods created with `spec.nodeName`, which made the count on the returned pod fail; Kanister's own label and caller labels are still asserted on the created pod.

## Pull request type

- [ ] :hamster: Trivial/Minor
- [ ] :bug: Bugfix
- [ ] :sunflower: Feature
- [x] :robot: Test

## Issues 

- fixes #4228

## Test Plan

- [x] :muscle: Manual
- [x] :zap: Unit test
- [ ] :green_heart: E2E

`TEST_FILTER=PodSuite make go-test` on kind v1.35.8 and v1.34.11 with `topology.kubernetes.io/zone` and `region` set on the node. Unpatched fails on 1.35.8 only; patched passes 16/16 on both.


pkg/kube/pod_test.go
```diff
@@ -238,7 +238,7 @@ func (s *PodSuite) TestPod(c *check.C) {
 			c.Assert(pod.ObjectMeta.Name, check.Equals, po.Name)
 		}
 
-		c.Check(len(pod.ObjectMeta.Labels), check.Equals, len(po.Labels)+1)
+		// Admission can add labels of its own (e.g. topology labels on pods with a nodeName), so only Kanister's labels are asserted.
```
PrasadG193: This removes test coverage of - validate Kanister doesn't add any other annotation than expected. We can still have the validation with 

```suggestion
		// Admission can add labels, so assert the exact set pre-submission.
		specPod, err := GetPodObjectFromPodOptions(context.Background(), s.cli, po)
		c.Assert(err, check.IsNil)
		c.Check(len(specPod.ObjectMeta.Labels), check.Equals, len(po.Labels)+1)
 		c.Check(pod.ObjectMeta.Labels[consts.LabelKeyCreatedBy], check.Equals, consts.LabelValueKanister)
```

PrasadG193 (commented):
Thanks for fixing this, left minor suggestion to reduce impact on test coverage

pkg/kube/pod_test.go
hr2822: yep fixed

PrasadG193 (approved):

hr2822:
Lint fixed in 1fffc3f (moved the CreatePod error assert up next to the call). The integration-test failure on the previous run was the bitnami helm index coming back empty on the runner (`no cached repo found ... bitnami-index.yaml: empty index.yaml file`) while installing MongoDB, unrelated to this change; the same job passed on the first run of this PR.

